Issue #13, opened 2021-08-28
Asset replacing breaks sometimes when using 'gif' as format


veni-vidi-code:
### Link

https://github.com/Rapptz/discord.py/issues/7345

ooliver1:
how did this not get closed from that pr

PR #56, opened 2021-08-29
Prefer `static_format` over `format` with static assets

Astrea49:
## Summary

In short: something like `Asset.replace(format='gif',static_format='png')` would break if it encountered a static asset, as static avatars cannot be stored in a GIF format and d.py preferred to use the `format` format vs. the `static_format` format, even with static assets. According to Danny, this was an intentional feature, but I personally do not see how keeping the old behavior would be beneficial to anyone at all.
Link to issue here: #13
Link to issue on `discord.py`: [#7345](https://github.com/Rapptz/discord.py/issues/7345)

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/docs/migrating_2.rst
+++ b/docs/migrating_2.rst
@@ -86,10 +86,14 @@ Assets have been changed.
 
     emoji_bytes = await emoji.read() # previously emoji.url.read
     # Same applies to Sticker and PartialEmoji.
-        
-Minimum Python version bumped
+
+Python Version Change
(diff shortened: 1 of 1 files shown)


sairam4123 (approved):

ooliver1 (approved):
LGTM

Astrea49:
Oops, this was a mistake. I'll open an identical PR to this soon - I just need to declutter my fork of nextcord.

PR #61, opened 2021-08-30
Split examples into a different section

tnix100:
## Summary
Makes examples a separate section because some people could get confused that the example is for the Optional Packages (https://discord.com/channels/881118111967883295/881118112492191796/881721194708082698)

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -71,7 +71,8 @@ Please note that on Linux installing voice you must install the following packag
 * libffi-dev (or ``libffi-devel`` on some systems)
 * python-dev (e.g. ``python3.6-dev`` for Python 3.6)
 
-
+Examples
+----------
 Quick Example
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -71,7 +71,8 @@ Please note that on Linux installing voice you must install the following packag
 * libffi-dev (or ``libffi-devel`` on some systems)
 * python-dev (e.g. ``python3.6-dev`` for Python 3.6)
 
-
+Examples
```
TAG-Epic, line 74: I personally think this should be solved by changing the name of Quick Example as we are only ever gonna show one in the readme at a time. Maybe change that instead of making another section.

TAG-Epic (changes requested):

PR #52, opened 2021-08-29
Fix pr reviews for #50 

TAG-Epic:
## Checklist

- [x] If code changes were made then they have been tested.
    - [ ] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.
- [ ] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

PR #62, opened 2021-08-30
Deleting the discord directory due to importerrors

rf20008:
## Summary

ImportErrors happen when trying to import both nextcord and discord.py, so removing the discord directory should fix that. Also most of the functionality is in the nextcord directory

## Checklist

- [ ] This PR fixes an issue.


TAG-Epic:
The import errors was fixed in #52. The reason the discord folder exists is because old code or external libraries should use nextcord.

rf20008:
that makes sense
:)

redcreeper14385:
Does #70 fix this at all?

redcreeper14385:
I'd rather not remove the folder for compatibility reasons.

redcreeper14385:
Oh I'm blind - didn't realise it was closed.

PR #86, opened 2021-08-31
Add `Guild.online_members`

redcreeper14385:
## Summary

This pull request adds `online_members` to the `Guild` class.
This exists to save time for bot developers who need to quickly check if a user is online, or to find out how many users are online.

## Checklist

- [ ] This PR fixes an issue.


TAG-Epic:
Let's not add properties for everything.

redcreeper14385:
I just thought they would be useful - it doesn't hurt to have them because it saves time for the developer.

abrahammurciano:
`online_members` imo is definitely useful

ooliver1:
Online members will be pretty useful for a few things, the extra properties make things more useful and cleaner

TAG-Epic:
Considering all of these can be solved in one line I think it's really unnecessary.

DenverCoder1:
Danny has usually been against implementing things that are not part of the Discord API.

If too many extra features are added it could add clutter and make it difficult to find the essential properties.

I do agree that it can be nice to have extra utils. It saves developers time so they don't have to write it themselves. Not sure if it needs to be part of the API Wrapper though.

ooliver1:
Online members is kinda part of the mobile client but I don’t know if the permission thing really works

TheMultii:
I don't really know if this needs to be part of the API wrapper. It might be a good idea to create some other library to add properties like `online_members` 🤔

redcreeper14385:
> I don't really know if this needs to be part of the API wrapper. It might be a good idea to create some other library to add properties like `online_members` 🤔 

I wouldn't say a completely separate library, although perhaps an `ext` module so that users can choose. 

It could be called `ext.variables` or something similar.

ooliver1:
surely it would have to be part of the Guild class though, its really just add it or dont add it

ooliver1:
unless if you pass in a Guild object, thats just more effort than making the list comp yourself

redcreeper14385:
Could you explain why you suddenly closed this?

redcreeper14385:
Anything?

abrahammurciano:
> Let's not add properties for everything.

Probably this is why

redcreeper14385:
> > Let's not add properties for everything.
> 
> Probably this is why

Yes but we then agreed to let the community give their opinion first.

AstroidDev:
Maybe add `nextcord.ext.utils` and then you could add `nextcord.ext.utils.getOnlineMembers(guild: discord.Guild)` along with other useful utils.

ooliver1:
That's about as long as just

``[m for m in guild.members if member.status.online]``

TAG-Epic:
This is unnecessary imo as it can be solved by a simple oneliner.

Issue #85, opened 2021-08-31
Suggestion to add nextcord's advantages to readme


DenverCoder1:
### Summary

The info in #project-info should be more prominently displayed to keep people interested.

### What is the feature request for?

The documentation

### The Problem

The readme currently is lacking a lot of info about what makes nextcord unique.

### The Ideal Solution

It is important to make sure people know the exact reasons they should use nextcord over alternatives.

It should be mentioned that nextcord is improving the code for ease-of-use and maintainability and the info such as what is found in #project-info could be added to the readme.

![image](https://user-images.githubusercontent.com/20955511/131562989-f7ef7172-1a44-4bd2-9aa3-6047ab2beeec.png)

DenverCoder1:
Closing due to closure of #​92

PR #66, opened 2021-08-30
add-welcome-screen

ooliver1:
## Summary

Adds getting welcome screen and editing functionality.
mostly copy of https://github.com/Rapptz/discord.py/pull/6477

## Checklist

- [x] If code changes were made then they have been tested.
    - [x] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.
- [x] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

https://cdn.discordapp.com/attachments/881874888590241832/881875044907765790/unknown.png https://cdn.discordapp.com/attachments/881874888590241832/881875292359127051/unknown.png https://cdn.discordapp.com/attachments/881874888590241832/881876231291805766/unknown.png https://cdn.discordapp.com/attachments/881874888590241832/881877075575865384/unknown.png https://cdn.discordapp.com/attachments/881874888590241832/881878620228972564/unknown.png https://cdn.discordapp.com/attachments/881874888590241832/881878683688771634/unknown.png https://cdn.discordapp.com/attachments/881874888590241832/881878756246028338/unknown.png

# **emoji does not work right now**

PR #92, opened 2021-09-01
Nextcord's advantages

TheMultii:
## Summary

Adding Nextcord's advantages to README (#85).

## Checklist

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,16 +32,16 @@ Key Features
 - Proper rate limit handling
 - Optimised in both speed and memory
 
-Reasons to use Nextcord over alternatives
--------------
+Why choose Nextcord?
+--------------------
 
-* No breaking changes unless necessery
-* A lot of bug fixes
-* Interactions (which are currently under development)
-* Compatibility with libraries such as:
+- No breaking changes unless necessary
+- Lots of bug fixes
+- Interactions (currently under development)
+- Compatibility with libraries such as:
 
-  * `jishaku <https://github.com/Gorialis/jishaku>`_
-  * `dislash <https://github.com/EQUENOS/dislash.py>`_
+  - `jishaku <https://github.com/Gorialis/jishaku>`_
+  - `dislash <https://github.com/EQUENOS/dislash.py>`_
 
 Installing
 ----------
--- a/README.rst
+++ b/README.rst
@@ -36,6 +36,7 @@ Why choose Nextcord?
 --------------------
 
 - No breaking changes unless necessary
+- Welcome Screen support
 - Lots of bug fixes
 - Interactions (currently under development)
 - Compatibility with libraries such as:
--- a/README.rst
+++ b/README.rst
@@ -32,6 +32,17 @@ Key Features
 - Proper rate limit handling
 - Optimised in both speed and memory
 
+Reasons to use Nextcord over alternatives
+-------------
+
+* No breaking changes unless necessery
+* A lot of bug fixes
+* Interactions (which are currently under development)
+* Compatibility with libraries such as:
+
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -32,6 +32,17 @@ Key Features
...
+Reasons to use Nextcord over alternatives
+-------------
+
+* No breaking changes unless necessery
+* A lot of bug fixes
+* Interactions (which are currently under development)
+* Compatibility with libraries such as:
+
+  * `jishaku <https://github.com/Gorialis/jishaku>`_
+  * `dislash <https://github.com/EQUENOS/dislash.py>`_
```
DenverCoder1, line 44: :+1: Looks good

A few suggestions:

- maybe make heading shorter
- remove extra words
- keep bullet-style consistent with other lists (`-` instead of `*`)
- Possibly "compatible libraries" could be a section on its own?

```suggestion
Why choose Nextcord?
--------------------

- No breaking changes unless necessary
- Lots of bug fixes
- Interactions (currently under development)
- Compatibility with libraries such as:

  - `jishaku <https://github.com/Gorialis/jishaku>`_
  - `dislash <https://github.com/EQUENOS/dislash.py>`_
```

DenverCoder1 (approved):

README.rst
```diff
@@ -32,6 +32,17 @@ Key Features
 - Proper rate limit handling
 - Optimised in both speed and memory
 
+Why choose Nextcord?
+--------------------
+
+- No breaking changes unless necessary
```
TAG-Epic, line 38: Maybe add welcome screen?

TAG-Epic (changes requested):

README.rst
TheMultii: Something like "Here's a list of reasons why you should start using Nextcord instead of other alternatives"? Because I don't quite understand what else a "welcome screen" would be.

TAG-Epic: Welcome screen support. You know those channels when you join a server?

ooliver1: see pr #66 for what they're on about

README.rst
```diff
@@ -32,6 +32,18 @@ Key Features
...
+- Welcome Screen support
+- Lots of bug fixes
+- Interactions (currently under development)
+- Compatibility with libraries such as:
```
TAG-Epic, line 42: I personally dont think this should be included.

TAG-Epic (changes requested):

The-Bow-Hunter (commented):
Maybe add the upgrade to the new API version

ooliver1:
V9 doesn't have anything new iirc

veni-vidi-code:
v8 and below do not support threads

ooliver1:
I meann, surely v8 does otherwise why can dpy use all features of threads

Astrea49:
...because d.py 2.0 is on v9.

ooliver1:
@Astrea49  https://github.com/nextcord/nextcord/blob/master/nextcord/http.py#L112

Astrea49:
> @Astrea49 https://github.com/nextcord/nextcord/blob/master/nextcord/http.py#L112

https://github.com/nextcord/nextcord/commit/72c66a17060d662bcead9ad6db91374e38fafbc1

ooliver1:
thats encoding not the base api lmao

Astrea49:
That's the gateway API. Which is the thing d.py uses to get threads.
The REST API is behind (or, well, was behind - that was fixed in a recent PR here), yes, but v9 is the *only* API version as of right now that has threads.

TAG-Epic:
Even if this is a good idea, it's probably gonna be biased. Would love to see a fork comparison or something from a unaffiliated party.

PR #101, opened 2021-09-01
Default command_prefix in commands.Bot()

DenverCoder1:
## Problem

If you don't want any prefix or even pings to trigger commands (for example if you only have slash commands or have no prefixed commands), you currently need a hackish workaround such as:

```py
bot = commands.Bot(command_prefix=" ")  # initial whitespace is ignored in messages
```

## Solution

This change makes it the default to not listen for any prefix. This way developers do not need to specify a prefix if they don't have any commands that use it (for example, if the bot only uses **slash commands** or listeners such as `on_message`)

Previously there was no default for `command_prefix`, so this should not cause any breaking changes.

This allows leaving out the `command_prefix` argument or passing an empty iterable to make the bot have no prefix:

```py
bot = commands.Bot()  # creates a Bot with no prefix
```

## Checklist

- [ ] This PR fixes an issue.


sairam4123 (approved):
Looks good to me.

ooliver1 (approved):

vcokltfre:
I don't think as a library we should be encouraging people to do this. If people decide of their own accord that they wish to use a mention prefix, as will undoubtedly be shown in various tutorials, that's up to them, but since Discord wants everyone to transition to slash commands I feel like providing this as a 'default escape route' is not helpful.

DenverCoder1:
I think there should at least be some default since with slash commands, people will want *no* prefix.

I've seen hacks in slash command bots such as
```py
bot = commands.Bot(command_prefix=" ")
```
in order to remove the prefix (white space is ignored so this will make it have no prefix)

DenverCoder1:
@vcokltfre It's not just about "providing a default escape route" but rather to allow bots to simply say 

```py
bot = commands.Bot()
```

when they do not want to have a prefix (for example if they only use slash commands).

It doesn't make sense to make the user choose a prefix when none of their commands use it.

The-Bow-Hunter:
> @vcokltfre It's not just about "providing a default escape route" but rather to allow bots to simply say 
> 
> ```py
> bot = commands.Bot()
> ```
> 
> when they do not want to have a prefix (for example if they only use slash commands).
> 
> It doesn't make sense to make the user choose a prefix when none of their commands use it.

Maybe make this using missing?

TAG-Epic (changes requested):
Imo command prefix should default to no prefix

nextcord/ext/commands/bot.py
```diff
@@ -122,9 +122,9 @@ def __repr__(self):
 _default = _DefaultRepr()
 
 class BotBase(GroupMixin):
-    def __init__(self, command_prefix, help_command=_default, description=None, **options):
+    def __init__(self, command_prefix=MISSING, help_command=_default, description=None, **options):
         super().__init__(**options)
-        self.command_prefix = command_prefix
+        self.command_prefix = command_prefix if command_prefix is not MISSING else tuple()
```
MaskDuck, line 127: seems great, but why is it a empty tuple instead of an empty string?

DenverCoder1: Empty string would trigger on every message (it is a prefix of all messages). Instead we want an empty non-string iterable, since that will not trigger anything (there are no prefixes in it to check).

MaskDuck: oh ok!

MaskDuck (approved):
really gud

MouseMoosz:
It would be nice if it wasn't required, as of now we can also use None but whenever there is ANY message sent, it raises a exception and logs it to the stderr.

Skelmis (approved):

PR #102, opened 2021-09-01
Feat: Healthcheck

TAG-Epic:
## Summary
This checks that everything in the library is as it should

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/__init__.py
+++ b/nextcord/__init__.py
@@ -60,7 +60,7 @@ from .stage_instance import *
 from .interactions import *
 from .components import *
 from .threads import *
-
+from .health_check import *
 
 class VersionInfo(NamedTuple):
     major: int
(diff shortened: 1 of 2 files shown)


ooliver1 (approved):
assuming you've tested, LGTM

rf20008 (approved):
LGTM assuming you've tested the changes :)

vcokltfre (changes requested):
health_check.py is definitely not PEP8 compliant, I'd suggest trying to keep new modules PEP8 compliant to make contributing easier and more consistent.

rf20008:
@TAG-Epic you need to catch pkg_resources.DistributionNotFound instead of all exceptions because warnings are technically built-in exceptions. Also you might want to warn with something other than a UserWarning which is the default warning if you don't specify a different warning in the category parameter in warnings.warn()

otherwise LGTM

rf20008 (approved):
LGTM. One thing you should consider is a different type of warning :)

rf20008 (approved):
LGTM

PR #138, opened 2021-09-02
feat: add timestamp object

vcokltfre:
## Summary

This PR adds a `Timestamp` object to allow the suer of Discord's native timestamps.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/timestamp.py
+++ b/nextcord/timestamp.py
@@ -4,6 +4,8 @@ from datetime import datetime as dt
 
 @dataclass
 class Timestamp:
+    """A class to represent Discord's native timestamp format."""
+
     time: int
 
     @property
--- a/nextcord/timestamp.py
+++ b/nextcord/timestamp.py
@@ -1,0 +1,70 @@
+from dataclasses import dataclass
+from datetime import datetime as dt
+
+
+@dataclass
+class Timestamp:
+    time: int
+
+    @property
+    def datetime(self) -> dt:
+        """The datetime object of the timestamp."""
+
+        return dt.fromtimestamp(self.time)
+
+    @property
+    def short_time(self) -> str:
+        """The timestamp in short time format."""
+
+        return f"<t:{self.time}:t>"
+
+    @property
+    def long_time(self) -> str:
+        """The timestamp in long time format."""
+
+        return f"<t:{self.time}:T>"
+
+    @property
+    def short_date(self) -> str:
+        """The timestamp in short date format."""
+
+        return f"<t:{self.time}:d>"
+
+    @property
+    def long_date(self) -> str:
+        """The timestamp in long date format."""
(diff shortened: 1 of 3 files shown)


ooliver1 (approved):

nextcord/timestamp.py
```diff
@@ -0,0 +1,72 @@
+from dataclasses import dataclass
+from datetime import datetime as dt
+
+
+@dataclass
+class Timestamp:
```
TAG-Epic, line 6: There should include a type imo

TAG-Epic (changes requested):

nextcord/timestamp.py
vcokltfre: What?

TAG-Epic: The time format type?

nextcord/timestamp.py
```diff
@@ -0,0 +1,70 @@
...
+    time: int
```
abrahammurciano: maybe change the from_datetime to be a constructor?

```suggestion
class Timestamp:
    @overload
    def __init__(self, unix_time: int):
        ...
    
    def __init__(self, datetime: dt):
        ...
        
   def __init__(self, arg: Union[int, dt]):
       if isinstance(arg, int):
           self.time = arg
       elif isinstance(arg, dt):
           self.time = round(datetime.timestamp())
       else:
           raise TypeError("First argument must be an integer or a datetime object")
```

vcokltfre: This seems like a needlessly complex way to construct it when a classmethod works; it makes more sense to me as a user if to construct from a timestamp I need to call a classmethod, rather than assuming it will do magic in the constructor

nextcord/timestamp.py
vcokltfre: Did you look at the class' properties?

nextcord/timestamp.py
abrahammurciano: no need to assume, thats what the overloads are for. your IDE will tell u. generally when I want to construct an object the first place I check is the constructor

nextcord/timestamp.py
```diff
@@ -0,0 +1,72 @@
+from dataclasses import dataclass
+from datetime import datetime as dt
+
+
+@dataclass
+class Timestamp:
+    """A class to represent Discord's native timestamp format."""
+
+    time: int
```
jgayfer, line 9: I think we should pass in a `datetime` object over a timestamp.

Very rarely am I dealing with a unix timestamp directly; I've almost always had a `datetime` that I'd like to display directly.

TAG-Epic:
I'm really not a fan of how this is implemented. It is missing features and is messy imo

PR #152, opened 2021-09-06
Making aliasgen.py PEP8 conform and adding support for other filesystems

veni-vidi-code:
## Summary

- flaking the file
- Adding support for e.g. Windows Filesystems using \ instead of /

## Checklist

- [ ] This PR fixes an issue.

**This is not towards the main branch!**

### DIFF
--- a/aliasgen.py
+++ b/aliasgen.py
@@ -4,7 +4,7 @@ from time import sleep
 
 with open(".aliasignorerc") as f:
     ignores = [line.strip() for line in f.readlines()]
-with open("alias_license.py") as f:
+with open("alias_license.txt") as f:
     license_text = f.read()
 
 print(f"Ignoring: {ignores}")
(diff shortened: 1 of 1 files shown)


aliasgen.py
```diff
@@ -46,18 +47,18 @@ def scan_dir(folder):
...
-
             with (alias_folder / unprefixed).open("w+") as f:
                 f.write(alias_file)
             sleep(0.01)
             # Test that it works
             import_name = "discord" + import_name[len("nextcord"):]
             try:
                 import_module(import_name)
-            except:
+            except ModuleNotFoundError:
```
TAG-Epic, line 57: We are checking for broken shit like typeerrors here too!

TAG-Epic (changes requested):

aliasgen.py
veni-vidi-code: maybe checking for Exception instead of using a bare except? or do you want to catch keyboard interrupts aswell?

TAG-Epic: I don't see where a keyboard interrupt would happen here

veni-vidi-code: Scenario: There is a problem siomewhere resulting in the code to hang up there for whatever reason. In order of stopping you would normally use a keyboard interrupt. instead of stopping you would have to repeat this for every single submodule now to stop the entire code. See PEP8 guide for more infos

>When catching exceptions, mention specific exceptions whenever possible instead of using a bare except: clause.
>[...]
> A bare except: clause will catch SystemExit and KeyboardInterrupt exceptions, making it harder to interrupt a program with Control-C, and can disguise other problems. If you want to catch all exceptions that signal program errors, use except Exception: (bare except is equivalent to except BaseException:).
> A good rule of thumb is to limit use of bare ‘except’ clauses to two cases:
> 
> If the exception handler will be printing out or logging the traceback; at least the user will be aware that an error has occurred.
> If the code needs to do some cleanup work, but then lets the exception propagate upwards with raise. try...finally can be a better way to handle this case.

TAG-Epic: Hmmm fair. Maybe change to anything but keyboard interrupt

veni-vidi-code: I changed it to Exception

TAG-Epic (approved):

Issue #155, opened 2021-09-08
Requesting removal of my Welcome Screen implementation.


NCPlayz:
Hi!

I am the writer of the Welcome Screen implementation that can be found on the official repository as a Pull Request (https://github.com/Rapptz/discord.py/pull/6477), or as a branch on my personal fork here: [`v1/guild/welcome-screen`](https://github.com/NCPlayz/discord.py/tree/v1/guild%2Fwelcome-screen).
As copyright of unmerged code belongs to the pull request author, and since I did not give permission for the use of this code in this repository, I am requesting that all code that belongs to this Pull Request, and is present in this repository, be removed.

Thank you for understanding.

noirscape:
to be clear this has no legal basis whatsoever.

you released your original contributions under the [MIT license](https://github.com/NCPlayz/discord.py/blob/v1/guild/welcome-screen/LICENSE). Licensing has typically been ruled to be covered by [estoppel](https://en.wikipedia.org/wiki/Estoppel), or in other words, you are not allowed to renege your licensing decisions on stuff you already released.

That isn't to say you are prevented from relicensing your code, but you released your contribution under the MIT, which means that those contributions would still be covered under the MIT. (This is the same mechanism btw that protects people's ability to fork off GPL projects that go closed source.)

TAG-Epic:
> to be clear this has no legal basis whatsoever.
> 
> you released your original contributions under the [MIT license](https://github.com/NCPlayz/discord.py/blob/v1/guild/welcome-screen/LICENSE). Licensing has typically been ruled to be covered by [estoppel](https://en.wikipedia.org/wiki/Estoppel), or in other words, you are not allowed to renege your licensing decisions on stuff you already released.
> 
> That isn't to say you are prevented from relicensing your code, but you released your contribution under the MIT, which means that those contributions would still be covered under the MIT. (This is the same mechanism btw that protects people's ability to fork off GPL projects that go closed source.)

You could argue it, although I respect their wish to get the code removed and had it removed. It's the right thing to do.

noirscape:
> You could argue it, although I respect their wish to get the code removed and had it removed. It's the right thing to do.

That's fair, I'm merely saying that permission wise there is no copyright problems with this whatsoever.

PR #157, opened 2021-09-08
code ownership for pull template

veni-vidi-code:
## Summary

In order of avoiding stuff like #155 in the future

## Checklist

- [ ] This PR fixes an issue.


TAG-Epic (changes requested):
Imo we should just require it be under MIT and be under Danny or my copyright or be public domain.

We then shouldn't need the original creator to be required (although I'm for this being optional)

PR #159, opened 2021-09-09
Example updates.

Shlol762:
Just connecting the example to nextcord.

### DIFF
--- a/examples/secure_token_storage.py
+++ b/examples/secure_token_storage.py
@@ -2,7 +2,7 @@
 # You should load the token as an environment variable, then access it like this:
 
 import os
-from nextcord.ext.commands import Bot
+from nextcord.ext import commands
 
 # There are other ways to load environment variables but this is one
 from dotenv import load_dotenv # https://pypi.org/project/python-dotenv/
--- a/examples/secure_token_storage.py
+++ b/examples/secure_token_storage.py
@@ -10,7 +10,7 @@ from dotenv import load_dotenv # https://pypi.org/project/python-dotenv/
 # load_dotenv reads from a file called .env in the same directory as the python files which should roughly look like TOKEN="1234567890"
 load_dotenv()
 token = os.getenv('BOT_TOKEN')
-bot = Bot()
+bot = commands.Bot()
 
 # run the bot using the token in .env
 bot.run(token)


examples/secure_token_storage.py
```diff
@@ -2,13 +2,15 @@
 # You should load the token as an environment variable, then access it like this:
 
 import os
+from nextcord.ext.commands import Bot
```
TAG-Epic, line 5:
```suggestion
from nextcord.ext import commands
```

TAG-Epic (changes requested):

TAG-Epic (approved):

examples/secure_token_storage.py
```diff
@@ -2,13 +2,15 @@
...
 import os
+from nextcord.ext import commands
 
 # There are other ways to load environment variables but this is one
 from dotenv import load_dotenv # https://pypi.org/project/python-dotenv/
 
 # load_dotenv reads from a file called .env in the same directory as the python files which should roughly look like TOKEN="1234567890"
 load_dotenv()
 token = os.getenv('BOT_TOKEN')
+bot = commands.Bot()
```
DenverCoder1, line 13: `commands.Bot()` does not have a default value for `command_prefix` so I think we should not be leaving that out in the example, unless something like #101 gets merged.

ooliver1: That won't be merged as vco had a pretty good point, maybe making the prefix something simple like ! Or ?

examples/secure_token_storage.py
```diff
@@ -2,13 +2,15 @@
 # You should load the token as an environment variable, then access it like this:
 
 import os
+from nextcord.ext import commands
 
 # There are other ways to load environment variables but this is one
 from dotenv import load_dotenv # https://pypi.org/project/python-dotenv/
 
 # load_dotenv reads from a file called .env in the same directory as the python files which should roughly look like TOKEN="1234567890"
```
ooliver1, line 10: Oh yeah one thing, that comment is incorrect "should look like TOKEN=" but the env taken is BOT_TOKEN

examples/secure_token_storage.py
TAG-Epic: Yeah I agree we need a example prefix here. Maybe use the same as in the other examples?

H4ckerxx44: I would suggest "?" as it is the prefix of all other examples

TAG-Epic: > I would suggest "?" as it is the prefix of all other examples

We have been using "$" everywhere else.

H4ckerxx44: Oh whoops, I meant $, lol.

PR #200, opened 2021-10-05
Fixed Bot.is_owner()'s type hint (added Member)

WieeRd:
## Summary

`Bot.is_owner()`'s doc says it can also take `nextcord.Member`, but the type hint in real code only had `nextcord.User`
## Checklist

- [ ] This PR fixes an issue.


veni-vidi-code:
FIrst please alos update the docs (the parameters must be changed in the comment below your change)

Second i would like to discuss if it is correct that member does not inherit from User

ooliver1:
it only really caes about the id, a member is basically a user

PR #202, opened 2021-10-06
Put Halloween banner into readme

veni-vidi-code:
## Summary

It puts the new halloween banner into the readme

## Checklist

- [ ] This PR fixes an issue.


DenverCoder1:
I think this banner is a bit too narrow, takes up a lot of space in the readme.

Maybe a new wide banner could be made in the same size.

TAG-Epic:
I honestly don't think we should include the Halloween banner in the readme, imo it would be better for the discord only.

PR #234, opened 2021-10-29
Add interaction.send()

ooliver1:
this was discussed at https://canary.discord.com/channels/881118111967883295/881124672056467456/903322918413541377

this has been tested by me too

## Summary

Adds ``interaction.send``, a fallback function for ``interaction.response.send_message`` and ``interaction.channel.send``, useful for everyone: converting and not having to make sure an interaction hasnt been responded to yet like in an interaction_check for a ``View``

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -30,7 +30,12 @@ import asyncio
 
 from . import utils
 from .enums import try_enum, InteractionType, InteractionResponseType
-from .errors import InteractionResponded, HTTPException, ClientException
+from .errors import (
+    InteractionResponded,
+    HTTPException,
+    ClientException,
+    InvalidData
+)
 from .channel import PartialMessageable, ChannelType
 
 from .user import User
@@ -358,6 +363,43 @@ class Interaction:
             self.token,
             session=self._session,
         )
+
+    async def send(self, *args, **kwargs) -> Optional[Message]:
+        """|coro|
+
+        This is a fallback function for help when sending messages in
+        response to an interaction. If the response
+        :property:`InteractionResponse.is_done()` then a messahe os sent
+        to the :property:`Interaction.channel` instead.
+
+        .. warning::
+
+            Ephemeral messages should not be sent with this as if
+            the :property:`Interaction.channel` is fallen back to,
+            ephemeral messages cannot be sent with this.
+
+        Returns
+        -------
+        Optional[:class:`Message`]
+            Message if the interaction has been responded to and the
+            interaction's channel was sent to. Else ``None``
+
+        Raises
+        ------
+        InvalidData
+            Somehow :property:`Interaction.channel` was ``None``,
+            this may occur in threads.
+        """
+
+        if not self.response.is_done():
+            await self.response.send_message(*args, **kwargs)
+        else:
+            if self.channel is not None:
+                await self.channel.send(*args, **kwargs)
+            else:
+                raise InvalidData(
+                    "Interaction.channel is None, this may occur in threads"
+                )
 
 
 class InteractionResponse:
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -392,14 +392,12 @@ class Interaction:
         """
 
         if not self.response.is_done():
-            await self.response.send_message(*args, **kwargs)
-        else:
-            if self.channel is not None:
-                await self.channel.send(*args, **kwargs)
-            else:
-                raise InvalidData(
-                    "Interaction.channel is None, this may occur in threads"
-                )
+            return await self.response.send_message(*args, **kwargs)
+        if self.channel is not None:
+            return await self.channel.send(*args, **kwargs)
+        raise InvalidData(
+            "Interaction.channel is None, this may occur in threads"
+        )
 
 
 class InteractionResponse:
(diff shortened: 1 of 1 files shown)


nextcord/interactions.py
```diff
@@ -359,6 +364,43 @@ async def delete_original_message(self) -> None:
             session=self._session,
         )
 
+    async def send(self, *args, **kwargs) -> Optional[Message]:
+        """|coro|
+
+        This is a fallback function for help when sending messages in
+        response to an interaction. If the response
+        :property:`InteractionResponse.is_done()` then a messahe os sent
+        to the :property:`Interaction.channel` instead.
```
dzshn, line 373: I see a few typos. Also a bit picky, but shouldn't it be this?
```rst
        This is a shorthand function for helping in sending messages in
        response to an interaction. If the response
        :property:`InteractionResponse.is_done()` then the message is sent
        via the :property:`Interaction.channel` instead.
```

nextcord/interactions.py
```diff
@@ -359,6 +364,43 @@ async def delete_original_message(self) -> None:
...
+        -------
+        Optional[:class:`Message`]
+            Message if the interaction has been responded to and the
+            interaction's channel was sent to. Else ``None``
+
+        Raises
+        ------
+        InvalidData
+            Somehow :property:`Interaction.channel` was ``None``,
+            this may occur in threads.
```
dzshn, line 391: Correct me if I'm wrong but I haven't seen `` :property:`...` `` be used, is it `` :attr:`...` ``?

dzshn (approved):

nextcord/interactions.py
ooliver1: oh god i did spell a few things wrong

nextcord/interactions.py
ooliver1: https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html?highlight=py%3Aproperty#directive-py-property

ooliver1: sphinx docs seem to be incorrect as ``docstring of nextcord.interactions.Interaction.send:18: WARNING: Unknown interpreted text role "property".``

ooliver1:
https://cdn.tooty.xyz/EyH4 as a preview

VincentRPS (approved):
lgtm

VincentRPS (commented):
isn't this just a shorter interaction.response.send_message?

TAG-Epic:
Voiced my issue with this on discord earlier. Forgot to close this.

ooliver1:
This never really got much discussion

ooliver1:
If it gets more discussion I'll consider similar

VincentRPS:
eh i don't think it's needed if people are too lazy to just to send_message then they shouldn't be using nextcord

ooliver1:
Fallback function, it's not too lazy

like it's just useful to not have to worry about if the interaction has already been responded to

Skelmis:
I personally don't think it's lazy. I think it's useful and maintains consistency across the library.
I see nothing wrong with making software *easier* to use


An example of this:
> eh i don't think it's needed if people are too lazy to just do response.send_message then they shouldn't be using nextcord

Like how the library creates DM channels if required, unless you create them yourself..?

ooliver1:
i do agree with this, we have other helper functions like utils which dont *need* to exist but help

RPSMain:
> i do agree with this, we have other helper functions like utils which dont _need_ to exist but help

Couldn’t you just do 
interaction.send = interaction.response.send_message?

ooliver1:
> > i do agree with this, we have other helper functions like utils which dont _need_ to exist but help
> 
> Couldn’t you just do interaction.send = interaction.response.send_message?

thats not the point of this, if the interaction has been responded to, it sends to the channel

RPSMain:
> > > i do agree with this, we have other helper functions like utils which dont _need_ to exist but help
> > 
> > 
> > Couldn’t you just do interaction.send = interaction.response.send_message?
> 
> thats not the point of this, if the interaction has been responded to, it sends to the channel

Oh I see

RPSMain (approved):

Skelmis:
You likely could. But the given idea will do more then just be a pointer, it'll ensure send works as first interaction and then subsequent messages.


Also, if the library doesn't implement this I find it quite likely someone else will for usage sake..
I know personally I would, albiet not a good enough reason but still. 

Also maintaining a stable interface among the library is a good idea, so by excluding send from here and changing that behavior your breaking the interface when used in conjunction with the rest of the library

ooliver1:
bear in mind if this gets a few approves ill add edit 
response.edit_message
if responded
interaction.message.edit

nextcord/interactions.py
```diff
@@ -359,6 +364,43 @@ async def delete_original_message(self) -> None:
...
+        if not self.response.is_done():
+            await self.response.send_message(*args, **kwargs)
+        else:
+            if self.channel is not None:
+                await self.channel.send(*args, **kwargs)
+            else:
+                raise InvalidData(
+                    "Interaction.channel is None, this may occur in threads"
+                )
+
```
Skelmis, line 404: This is a personal thing, so feel free to ignore
```python
if not self.response.is_done():
    return await self.response.send_message(*args, **kwargs)

if self.channel is not None:
    return await self.channel.send(*args, **kwargs)

raise InvalidData(
    "Interaction.channel is None, this may occur in threads"
)
```

Skelmis (approved):

vcokltfre (approved):

PR #235, opened 2021-10-29
Feat: Add 'thread' field to messages.

Artemis21:
## Summary

Parse and use the `thread` field on messages, as documented [here](https://discord.com/developers/docs/resources/channel#message-object).

## Checklist

- If code changes were made then they have been tested.
- I have updated the documentation to reflect the changes.
- This PR adds something new (e.g. new method or parameters).

### DIFF
--- a/nextcord/message.py
+++ b/nextcord/message.py
@@ -683,7 +683,7 @@ class Message(Hashable):
             self.guild = state._get_guild(utils._get_as_snowflake(data, 'guild_id'))
 
         thread_data = data.get('thread')
-        if self.guild and thread_data:
+        if thread_data:
             self.thread: Optional[Thread] = Thread(guild=self.guild, state=state, data=thread_data)
         else:
             self.thread: Optional[Thread] = None
--- a/nextcord/message.py
+++ b/nextcord/message.py
@@ -599,7 +599,7 @@ class Message(Hashable):
 
         .. versionadded:: 2.0
     thread: Optional[:class:`Thread`]
-        The thread created from this message, if any.
+        The thread created from a message, if any.
 
         .. versionadded:: 2.0
     guild: Optional[:class:`Guild`]
--- a/nextcord/message.py
+++ b/nextcord/message.py
@@ -598,6 +598,10 @@ class Message(Hashable):
         A list of components in the message.
 
         .. versionadded:: 2.0
+    thread: Optional[:class:`Thread`]
+        The thread created from this message, if any.
+
+        .. versionadded:: 2.0
     guild: Optional[:class:`Guild`]
         The guild that the message belongs to, if applicable.
     """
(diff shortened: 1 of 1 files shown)


ooliver1 (approved):
I assume this has been tested

Artemis21:
Yes, I made the same change to a Discord.py fork a while back and it has been used successfully in production.

nextcord/message.py
```diff
@@ -677,6 +682,12 @@ def __init__(
         except AttributeError:
             self.guild = state._get_guild(utils._get_as_snowflake(data, 'guild_id'))
 
+        thread_data = data.get('thread')
+        if self.guild and thread_data:
```
TAG-Epic, line 686: Threads are a guild only feature? Why do this check?

TAG-Epic (changes requested):
Needs testing too, don't have time atm

nextcord/message.py
Artemis21: To convince the type checker (and myself) that it won't break. I can remove it if you want, but I think it's best for safety - Discord may theoretically add threads to DMs in the future, in which case it's better that this code doesn't load them than that it breaks.

Artemis21:
I have tested it, and the branch this pull request comes from is currently being used in production (and actively using the feature) without issues.

nextcord/message.py
TAG-Epic: Honestly it should just break... Doubt they will be adding thread support for DMs

nextcord/message.py
```diff
@@ -597,6 +597,10 @@ class Message(Hashable):
     components: List[:class:`Component`]
         A list of components in the message.
 
+        .. versionadded:: 2.0
+    thread: Optional[:class:`Thread`]
+        The thread created from this message, if any.
```
RPSMain, line 602:
```suggestion
        The thread created from a message, if any.
```

RPSMain (approved):

Skelmis:
@Artemis21 Any updates?

Artemis21:
@Skelmis addressed reviews...

vcokltfre (approved):

TAG-Epic (approved):
LGTM, not tested

koala9712:
Returns `None` in a regular channel as well as a thread
with code:
```python
@commands.command()
async def test(self, ctx):
    await ctx.send(str(ctx.message.thread))
```

image
image

Artemis21:
@koala9712 

> Returns `None` in a regular channel as well as a thread with code:

To get the channel or thread a message was sent in, use `Message.channel`. This attribute is for determining which if any thread was created *from* this message, as documented.

koala9712:
>This attribute is for determining which if any thread was created _from_ this message, as documented.

So... this should theoretically have...?
unknown
unknown

Issue #123, opened 2021-09-02
Add a way for group commands to inherit handlers (checks, before/after invoke)


sairam4123:
### Link

https://github.com/Rapptz/discord.py/projects/3#card-58810560

PR #238, opened 2021-10-30
Feature: Group's hook & check inheritance

Skelmis:
## Summary

Add's the ability for commands in groups to inherit checks, pre_invoke and after_invoke calls

#123

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/ext/commands/core.py
+++ b/nextcord/ext/commands/core.py
@@ -283,9 +283,8 @@ class Command(_BaseCommand, Generic[CogT, P, T]):
 
         .. note::
             You can define a ``pre_invoke`` or ``after_invoke`` on this
-            command in conjunction with ``inherit_hooks=True`` however
-            the one defined for this command will not be replaced by
-            the parent hook.
+            command in conjunction with ``inherit_hooks=True``, however
+            it will override the parent hook if applicable
 
         .. versionadded:: 2.0.0
     """
--- a/nextcord/ext/commands/core.py
+++ b/nextcord/ext/commands/core.py
@@ -277,6 +277,17 @@ class Command(_BaseCommand, Generic[CogT, P, T]):
 
 
         .. versionadded:: 2.0
+    inherit_hooks: :class:`bool`
+        If ``True`` and this command has a parent :class:`Group` then this command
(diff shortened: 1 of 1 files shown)


nextcord/ext/commands/core.py
```diff
@@ -386,6 +397,31 @@ def __init__(self, func: Union[
         else:
             self.after_invoke(after_invoke)
 
+        # Attempt to bind to parent hooks if applicable
+        if not kwargs.get("inherit_hooks", False) or not self.parent:
```
TAG-Epic, line 401: Wouldnt it make more sense for it to error if it doesn't have a parent?

TAG-Epic (changes requested):

nextcord/ext/commands/core.py
Skelmis: No, unless I misunderstand how the internals work.

``self.parent`` is None in the case of things like ``@bot.command()`` (fairly sure) so it ignores those rather than throwing an error was my line of thinking

Skelmis: @TAG-Epic

VincentRPS (approved):
lgtm

ooliver1 (approved):

nextcord/ext/commands/core.py
```diff
@@ -277,6 +277,17 @@ class Command(_BaseCommand, Generic[CogT, P, T]):
...
 
         .. versionadded:: 2.0
+    inherit_hooks: :class:`bool`, default=False
+        If ``True`` and this command has a parent :class:`Group` then this command
+        will inherit all checks, pre_invoke and after_invoke's defined on the the :class:`Group`
+
+        .. note::
+            You can define a ``pre_invoke`` or ``after_invoke`` on this
+            command in conjunction with ``inherit_hooks=True`` however
+            the one defined for this command will not be replaced by
```
TAG-Epic, line 287: maybe rewrite it to say that the one you set per command will override the parent ones?

TAG-Epic (changes requested):

TAG-Epic (approved):
LGTM

PR #244, opened 2021-11-02
Updated License

VincentRPS:
## Summary

Chore: Update Licenses To All Files

## Checklist

- [x] This PR Updates license info

### DIFF
--- a/nextcord/__init__.py
+++ b/nextcord/__init__.py
@@ -4,7 +4,7 @@ Discord API Wrapper
 
 A basic wrapper for the Discord API.
 
-:copyright: (c) 2015-present Rapptz
+:copyright: (c) 2015-2021 Rapptz & 2021-present tag-epic
 :license: MIT, see LICENSE for more details.
 
 """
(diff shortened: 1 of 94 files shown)


nextcord/__init__.py
```diff
@@ -4,7 +4,7 @@
 
 A basic wrapper for the Discord API.
 
-:copyright: (c) 2015-present Rapptz
+:copyright: (c) 2015-2021 Rapptz & 2021-present tag-epic
```
veni-vidi-code, line 7: and from 2021 on?

veni-vidi-code (changes requested):
I would swap the order (first epic then Rapptz

nextcord/__init__.py
VincentRPS: Well only this one cause i am not spending another 20 minutes copy-pasting license info again

TAG-Epic:
Let's only add my copyright to files I've modified.

nextcord/__init__.py
veni-vidi-code: this one here is different from the others and should definitly be changed

VincentRPS: yeah

> this one here is different from the others and should definitly be changed

PR #255, opened 2021-11-14
Adding 3.10 in supported python versions

VincentRPS:
## Summary

Adds Python 3.10 As A Supported Version Of Python

## Checklist


TAG-Epic:
3.10 should be tested before it's supported

VincentRPS:
> 3.10 should be tested before it's supported

I have tested it it’s my main python version and I’ve had no problems with it and nextcord

veni-vidi-code:
have you also tested voice, every method and commands ext?

VincentRPS:
> have you also tested voice, every method and commands ext?

I’ve tested it in most commands ext method but actually not voice yet

TAG-Epic (changes requested):
This doesn't change it in the unaliased version

PR #268, opened 2021-11-23
Update secure_token_storage.py

ArinjoyProgrammer:
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/examples/secure_token_storage.py
+++ b/examples/secure_token_storage.py
@@ -13,4 +13,4 @@ load_dotenv()
 bot = commands.Bot(command_prefix="$")
 
 # run the bot using the token in .env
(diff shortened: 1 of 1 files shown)


VincentRPS (changes requested):
I don't think this is gonna work since BOT_TOKEN isn't defined

Skelmis (changes requested):
Your changes make this code fail to run by introducing a NameError. 

Not sure what the aim of this PR is also, given none of the checklist is checked and there's no description..

ooliver1 (changes requested):
This is not how os.getenv works, it takes a string

PR #271, opened 2021-11-24
Add Build Test

RPSMain:
## Summary

Adding A Test Which Builds Nextcord

## Checklist

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,22 @@
+name: Build
+
+on: [push, pull_request]
+
+jobs:
+  docs:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        python-version: [ 3.8 ]
+    steps:
+    - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


TAG-Epic (changes requested):
We don't use build to build nextcord? Also is this needed?

ooliver1:
Nc isn't a built library

RPSMain:
> We don't use build to build nextcord? Also is this needed?

It is great for pointing out errors before pypi uploads and stuff, and if you want I could change it to sdist even if there is no difference

Skelmis:
What "errors"

You can build distributable with invalid code fine..?

TAG-Epic:
Yeah... Even if this would be merged I would prefer using no deps aaand testing unaliased too.

ooliver1:
python is an interpreted language mostly (cpthon), so unlike compiled languages, it's harder to point out errors

as it's dynamically typed, it's hard to pick out every error

PR #279, opened 2021-12-03
Remove **kwargs where applicable

Skelmis:
## Summary

Discord.py had a prolific use of **kwargs where it is unnecessary. This makes it both hard for new developers and experienced developers alike to get arguments right in what should be as simple as your editor pointing out the arguments.

This PR removes **kwargs where applicable while maintaining the same usage. 

Note, this will break people being dumb and passing key word arguments that things don't take. Although I could keep that functionality although quite frankly I think its not a good idea.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -56,15 +56,15 @@ from .sticker import GuildSticker, StickerItem
 from . import utils
 
 __all__ = (
-    'Snowflake',
-    'User',
-    'PrivateChannel',
-    'GuildChannel',
-    'Messageable',
-    'Connectable',
+    "Snowflake",
+    "User",
+    "PrivateChannel",
+    "GuildChannel",
+    "Messageable",
+    "Connectable",
 )
 
-T = TypeVar('T', bound=VoiceProtocol)
+T = TypeVar("T", bound=VoiceProtocol)
 
 if TYPE_CHECKING:
     from datetime import datetime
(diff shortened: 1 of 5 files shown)


Skelmis:
If this is something you are against (and I don't see why you'd be against making it easier to use), please let me know asap so I don't waste my time improving the library.

RPSMain:
is it just me or was the code linted? 🤔

ooliver1:
@Skelmis could you redo it without auto formatting

ooliver1:
However personally I would say it's a good idea, I've had many times I'd have to open a browser just to see valid kwargs

rf20008:
This is going to be a breaking change (so we should keep **kwargs but not use it). 
However, LGTM to me so far (assuming this is tested)

Skelmis:
Stalling this over the holidays then will see where everything stands down the line.

TAG-Epic:
> This is going to be a breaking change (so we should keep **kwargs but not use it). 

Honestly not a fan of it

ooliver1:
^, like why is the user still using old kwargs, it should tell them its not needed anymore/doesnt exist

Skelmis:
I no longer have the motivation to do this. 

Might re-open in the future but as it currently stands I don't think I'll be getting back to that point any time soon.

PR #298, opened 2021-12-22
Merge: #298

Lulalaby:


VincentRPS (approved):

vcokltfre (approved):

ToasterUwU (approved):
With the get function fix, it doesn't have the potential to break and adds a valueable feature. So here is my approval.

TAG-Epic (changes requested):
This should probably be a enum and then fallback to a string??

Lulalaby:
No.
Better don't do this.

Lulalaby:
Reason: https://docs.dcs.aitsys.dev/articles/application_commands/translations/reference.html#valid-locales

vcokltfre:
I'm not sure how that link somehow proved that we shouldn't use an enum?

Lulalaby:
We talked in the guild about it.
I won't do anything more.
If you want to switch it to an enum, please do it yourself.

DenverCoder1 (approved):
I think it's fine to leave it as a string

It's more easily maintainable this way and it matches the [docs' specifications](https://discord.com/developers/docs/interactions/receiving-and-responding#interaction-object-interaction-structure)

Imo, maintaining a list of locales would be unnecessary and overcomplicating it

TAG-Epic (approved):
Hmmm revised my thoughts on this. Why not!

Lulalaby:
Finally :p

PR #313, opened 2021-12-29
Fix remove_application_command popping keys instead of values

Scaarus:
## Summary

When removing a command via `remove_application_command`, it attempts to remove command ids from dictionary keys rather than values. This fix simply switches to removing by value instead.

## Checklist

- [x] This PR fixes an issue.


vcokltfre:
Fixed by #320

PR #316, opened 2021-12-29
Documenting that guild.chunk returns a list of members

Scrumpyy:
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/guild.py
+++ b/nextcord/guild.py
@@ -2845,7 +2845,7 @@ class Guild(Hashable):
 
         await self._state.http.edit_widget(self.id, payload=payload)
 
-    async def chunk(self, *, cache: bool = True) -> None:
+    async def chunk(self, *, cache: bool = True) -> List[Member]:
         """|coro|
 
         Requests all members that belong to this guild. In order to use this,
--- a/nextcord/guild.py
+++ b/nextcord/guild.py
@@ -2864,6 +2864,11 @@ class Guild(Hashable):
         -------
         ClientException
             The members intent is not enabled.
+
+        Returns
+         --------
+         members :class:`list`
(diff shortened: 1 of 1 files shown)


vcokltfre (changes requested):
While we're here, the return type of the function should be changed to a list of Members too `-> None` -> `-> List[Member]` (make sure List is imported from typing)

TAG-Epic (changes requested):
This can also be None?

nextcord/guild.py
```diff
@@ -2845,7 +2845,7 @@ async def edit_widget(self, *, enabled: bool = MISSING, channel: Optional[Snowfl
 
         await self._state.http.edit_widget(self.id, payload=payload)
 
-    async def chunk(self, *, cache: bool = True) -> None:
+    async def chunk(self, *, cache: bool = True) -> List[Member]:
```
ooliver1, line 2848: This also needs to have the return type with optional

ooliver1 (changes requested):

nextcord/guild.py
```diff
@@ -2845,7 +2845,7 @@ async def edit_widget(self, *, enabled: bool = MISSING, channel: Optional[Snowfl
 
         await self._state.http.edit_widget(self.id, payload=payload)
 
-    async def chunk(self, *, cache: bool = True) -> None:
+    async def chunk(self, *, cache: bool = True) -> Optional(List[Member]):
```
ooliver1, line 2848: Isn't it Optional[] not Optional()

ooliver1 (changes requested):

ooliver1 (approved):

vcokltfre (approved):

Issue #310, opened 2021-12-26
Setting up 3 slash commands caused cloudflare ban


FightMan01:
### Summary

I tried to set up 3 slash commands with a bot over 11k guilds

### Reproduction Steps

Set up multiple slash comands.

### Expected Results

Slash commands will be created.

### Actual Results

I've got a global rate limit then a cloudflare ban.
I have a log and also I log all the rate limits into my console. My console was spammed with this:
<img width="1792" alt="image" src="https://user-images.githubusercontent.com/29693718/147405231-ff9aaf7e-17a0-4c3a-a5b9-83b897a458dc.png">
I have 21k lines of this :/

### Intents

all

### System Information

- Python v3.9.2-final
- nextcord v2.0.0-alpha
    - nextcord pkg_resources: v2.0.0a4
- aiohttp v3.8.1
- system info: Linux 5.10.0-9-amd64 #1 SMP Debian 5.10.70-1 (2021-09-30)

### Checklist

- [X] I have searched the open issues for duplicates.
- [X] I have shown the entire traceback, if possible.
- [X] I have removed my token from display, if visible.

ZRunner:
Hey, would you share the code you used to registrer your commands please? Just to make sure you did absolutely everything right (also it may help debugging the code, who knows)

alentoghostflame:
As part of synchronization between the bot and Discord, one web call is made per guild on startup. Across 11k guilds, this behavior is obviously not helpful. To prevent the web calls from being made, please override the `on_guild_available` event, as that's where the calls are being made.

Be warned, this will prevent the bot from syncing and deleting unknown or adding new commands across guilds. If you want to have it sync with a specific guild, use `get_guild` or something similar to get the `Guild` object, then `await guild.deploy_application_commands()` and `await guild.register_new_application_commands()`.

A toggle to make the bot not check for existing commands when no commands are added to them is coming soon.

FightMan01:
I had a handmade slash handler when nextcord didn't have one. 
I just used the `https://discord.com/api/v9/applications/{bot_user_id_here}/commands` endpoint to register a slash command globally. This also synced new commands and the changes in the existing commands when I needed. For me needs (and I think the need for most of us) this is enough.
I don't see the point why nextcord goes through each guild and get the existing slash commands.

To @ZRunner :
```py
    @slash_command(description="A bot prefixének visszaállítása.", name="prefix")
    async def _prefix_slash(self, interaction: Interaction):
[...]

PR #320, opened 2021-12-30
Slash fixes

alentoghostflame:
SlashOption's choice param now takes an iterable.
rollout_all_guilds kwarg for Client now determines if all Guilds are rollout'd to instead of just the ones specified by the guild_ids param in commands. This is a fix towards the cloudflare ban.
handle_slash_argument is now async, and will auto-fetch user/member if they aren't found in the cache.
Added basic cog reloading support, fix application commands not actually getting removed.
Guild now has its own rollout command.

## Summary

Adds a fix towards #310, but the only real fix for that is fixing http.
Contains the same exact fix as #313, sorry.
When a member/user can't be gotten from the cache (hi intents!), slash commands will now attempt to fetch it.
Adds a variety of fixes and a few helpers for slash stuff.

## Checklist

- [X] This PR fixes an issue.
- [X] This PR is a breaking change (e.g. methods or parameters removed/renamed) [for some inner slash stuff]

### DIFF
--- a/nextcord/guild.py
+++ b/nextcord/guild.py
@@ -2994,6 +2994,44 @@ class Guild(Hashable):
             update_known=update_known
         )
 
+    async def rollout_application_commands(
+            self,
+            associate_known: bool = True,
+            delete_unknown: bool = True,
+            update_known: bool = True,
+            register_new: bool = True
+    ) -> bool:
+        """|coro|
+        Rolls out application commands to the guild, associating, deleting, updating, and/or newly
+        registering as needed.
+
+        Parameters
+        ----------
+        associate_known: :class:`bool`
+            Whether commands on Discord that match a locally added command should be associated with each other.
+            Defaults to ``True``
+        delete_unknown
+        update_known
+        register_new
+
+        Returns
+        -------
+        :class:`bool`
+            ``False`` if no commands that specify a guild have been added to the bot. ``True`` otherwise.
+        """
+        if self._state.get_guild_application_commands(self.id, rollout=True):
+            guild_payload = await self._state.http.get_guild_commands(self._state.application_id, self.id)
+            await self.deploy_application_commands(
+                data=guild_payload,
+                associate_known=associate_known,
+                delete_unknown=delete_unknown,
+                update_known=update_known
+            )
+            if register_new:
+                await self.register_new_application_commands(data=guild_payload)
+            return True
+        return False
+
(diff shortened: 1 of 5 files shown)


VincentRPS:
@alentoghostflame client.add_cog exists but there is no way to actually make a cog without commands.Bot

VincentRPS:
@alentoghostflame client.add_cog exists but there is no way to actually make a cog without commands.Bot

alentoghostflame:
> @alentoghostflame client.add_cog exists but there is no way to actually make a cog without commands.Bot

Two things about this:
1: ClientCog exists.
2: Discussions about bringing cogs into Client have been made, so it will likely be relevant for the future.

VincentRPS:
On number 2 why should they be moved? Cogs aren’t a discord feature in anyway so we shouldn’t treat it like one

Skelmis:
> On number 2 why should they be moved? Cogs aren’t a discord feature in anyway so we shouldn’t treat it like one, since every feature in nextcord/ is treated as a discord feature and every thing nextcord/ext an extension

I believe it comes under 'implementation details', you can't tell me that if you were making a bot using just ``discord.Client`` you *wouldn't* want the ability to separate your code into multiple files to make it overall better?

Tbh I'm surprised Cogs weren't a thing in client from the start

alentoghostflame:
Cogs are the only way currently to easily break a project up into multiple files when using Slash or Listeners/Events.

VincentRPS:
If you want to use cogs just use commands.Bot, There’s no downside to it

Skelmis:
As a fork we have been given a chance to *improve* useability, theres nothing wrong with porting the ability to code a bot in multiple files. Your argument is just we don't do it now, why bother

nextcord/guild.py
```diff
@@ -2994,6 +2994,44 @@ async def deploy_application_commands(
...
+        """|coro|
+        Rolls out application commands to the guild, associating, deleting, updating, and/or newly
+        registering as needed.
+
+        Parameters
+        ----------
+        associate_known: :class:`bool`
+            Whether commands on Discord that match a locally added command should be associated with each other.
+            Defaults to ``True``
+        delete_unknown
```
ooliver1, line 3013: will these get any docs

vcokltfre (approved):

ooliver1 (approved):

PR #324, opened 2022-01-02
Add raw_member_remove event

Gugu7264:
## Summary

Adds a raw_member_remove event, useful in case the member isn't cached (or chunking is disabled) which makes the client not fire an event at all when a member leaves a guild.

## Checklist

- [ ] This PR fixes an issue.


Skelmis:
Why does this pr include work on allowing the editing of files in messages? That's unrelated isn't it.

Gugu7264:
oww crap, that was from my other PR, making a new PR soon

PR #325, opened 2022-01-02
Add raw_member_remove event

Gugu7264:
## Summary

Adds a raw_member_remove event, useful in case the member isn't cached (or chunking is disabled) which makes the client not fire an event at all when a member leaves a guild.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/raw_models.py
+++ b/nextcord/raw_models.py
@@ -54,7 +54,7 @@ __all__ = (
     'RawReactionClearEmojiEvent',
     'RawIntegrationDeleteEvent',
     'RawTypingEvent',
-    "RawMemberRemoveEvent"
+    "RawMemberRemoveEvent",
 )
 
 
--- a/nextcord/raw_models.py
+++ b/nextcord/raw_models.py
@@ -36,11 +36,13 @@ if TYPE_CHECKING:
         ReactionClearEvent,
         ReactionClearEmojiEvent,
         IntegrationDeleteEvent,
-        TypingEvent
+        TypingEvent,
+        MemberRemoveEvent
     )
     from .message import Message
     from .partial_emoji import PartialEmoji
     from .member import Member
+    from .user import User
 
 
 __all__ = (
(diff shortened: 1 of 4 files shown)


ooliver1 (changes requested):
did you add anything to the docs?

nextcord/raw_models.py
```diff
@@ -52,6 +54,7 @@
     'RawReactionClearEmojiEvent',
     'RawIntegrationDeleteEvent',
     'RawTypingEvent',
+    "RawMemberRemoveEvent"
```
vcokltfre, line 57:
```suggestion
    "RawMemberRemoveEvent",
```

nextcord/raw_models.py
```diff
@@ -311,4 +314,21 @@ def __init__(self, data: TypingEvent) -> None:
...
+
+class RawMemberRemoveEvent(_RawReprMixin):
+    """Represents the payload for a :func:`on_raw_member_remove` event.
+    .. versionadded:: 2.0
+    Attributes
+    -----------
+    guild_id: :class:`int`
+        The guild ID where the member left from.
+    user: :class:`User`
+        The user who left the guild."""
```
vcokltfre, line 328:
```suggestion
        The user who left the guild.
    """
```

vcokltfre (changes requested):
Looks good for the most part, just a couple of nitpicks I'd like fixed, the trailing commas are good practice for the future, since other things added won't require 2 lines of diff, and the docstring change is to remain consistent with other docstrings

nextcord/raw_models.py
ooliver1: ^ for this, https://www.python.org/dev/peps/pep-0257/#handling-docstring-indentation is the reason

TAG-Epic (approved):
Not tested

ooliver1 (approved):

ooliver1:
3 approves so merging (commits after vco were his requested changes)

PR #342, opened 2022-01-08
Add handler for message thread field update

Artemis21:
## Summary

Parse and use the `thread` field on message update, not just when creating a new message.

For example, the following code:

```python
import nextcord

client = nextcord.Client(intents=nextcord.Intents.all())

@client.event
async def on_message_edit(before, after):
    await after.channel.send(str(after.thread))

client.run("token")
```

Would send `None` in any circumstance before this change, but will send the name of the thread created from the edited message (if any) after this change. This has been tested.

## Checklist

- [x] This PR fixes an issue.


koala9712:
Tested and works!

PR #344, opened 2022-01-09
Slash fixes

alentoghostflame:
## Summary

* Reverted command check inside Guild.rollout_applicaiton_commands, added it to Client.on_guild_available. Fixed rollout_all_guilds kwarg.
* Fixed payload mismatch issue by forcing option names to be strings.
* Fixed optional params being broken.
* Added functions to get global commands and all added ApplicationCommand objects.
* Self argument for the callback of ApplicationCommand now gettable via .self_argument
* Added Guild.get_application_commands() to get commands added to the guild.
* Added ability for autocomplete callbacks to return a dict or Iterable instead of calling Interaction.response.send_autocomplete()
* Messages commands will now return the cached message object if it exists.
* Fully commented out and moved register_bulk_application_commands() to state.py

## Checklist

    - Note: Git had a conflict in guild.py, I'm *pretty sure* I resolved it correctly but ehhhhh...
- [ ] This PR fixes an issue.
- [X] This PR is a breaking change (e.g. methods or parameters removed/renamed)
    - For Slash functions.

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -304,11 +304,11 @@ class CommandOption(SlashOption):
             return state._get_message(int(argument))
         return argument
 
-    async def handle_message_argument(self, *args: List[Any]):
+    async def handle_message_argument(self, state: ConnectionState, argument: Any, interaction: Interaction):
         """For possible future use, will handle arguments specific to Message Commands (Context Menu type.)"""
         raise NotImplementedError  # TODO: Even worth doing? We pass in what we know already.
 
-    async def handle_user_argument(self, *args: List[Any]):
+    async def handle_user_argument(self, state: ConnectionState, argument: Any, interaction: Interaction):
         """For possible future use, will handle arguments specific to User Commands (Context Menu type.)"""
         raise NotImplementedError  # TODO: Even worth doing? We pass in what we know already.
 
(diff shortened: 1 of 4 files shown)


nextcord/application_command.py
```diff
@@ -304,11 +304,11 @@ async def handle_slash_argument(self, state: ConnectionState, argument: Any, int
...
 
-    async def handle_message_argument(self, *args: List[Any]):
+    async def handle_message_argument(self, state: ConnectionState, argument: Any, interaction: Interaction):
         """For possible future use, will handle arguments specific to Message Commands (Context Menu type.)"""
         raise NotImplementedError  # TODO: Even worth doing? We pass in what we know already.
 
-    async def handle_user_argument(self, *args: List[Any]):
+    async def handle_user_argument(self, state: ConnectionState, argument: Any, interaction: Interaction):
         """For possible future use, will handle arguments specific to User Commands (Context Menu type.)"""
         raise NotImplementedError  # TODO: Even worth doing? We pass in what we know already.
```
vcokltfre, line 313: Can these be removed, since they are not even planned to be implemented in the API there is no reason to preemptively add them to Nextcord

vcokltfre (changes requested):

nextcord/application_command.py
alentoghostflame: I'm planning on moving some of the message/user specific argument stuff to those, but the fixes present in this PR are more important than spending time doing that right now.

vcokltfre (approved):

nextcord/state.py
```diff
@@ -775,6 +787,14 @@ async def delete_application_command(self, command: ApplicationCommand, guild_id
         self._application_command_ids.pop(command.command_ids[guild_id], None)
         self._application_command_signatures.pop(command.get_signature(guild_id))
 
+    # async def register_bulk_application_commands(self) -> None:
```
TAG-Epic, line 790: Dead code?

TAG-Epic (changes requested):

nextcord/state.py
vcokltfre: > It might be good to keep this around as a reminder for future work.

TAG-Epic: My bad.

TAG-Epic (approved):

PR #355, opened 2022-01-19
Allow `__future__` `annotations` import in slash

### LINKED ISSUE #328: Client.slash_command fails when annotations future is enabled
### Summary

Client.slash_command fails when annotations future is enabled

### Reproduction Steps

Produce a basic bot using nextcord.Client, then try to create a slash command with `from __future__ import annotations` enabled.

I have not tested if this is also the case with nextcord.ext.Bot.

### Minimal Reproducible Code

```python

import nextcord

bot = nextcord.Client()


@bot.slash_command("echo")
async def echo(it, arg: str):
    await it.response.send_message(arg)


if __name__ == '__main__':
    bot.run(TOKEN)
```

### Expected Results

A bot with the command "/echo"

This is the case for if the annotations import is commented out.

### Actual Results

A `NotImplementedError` stating that "str" is not a valid type for an argument. 

Debugger shows that the value for `typing` in `application_command.CommandOption.get_type` is `{str} 'str'`

```
  File "E:\scratch\bugtest.py", line 9, in <module>
    async def test(it, arg: str):
  File "E:\venv\lib\site-packages\nextcord\client.py", line 1972, in decorator
  File "E:\venv\lib\site-packages\nextcord\application_command.py", line 1232, in decorator
    app_cmd = ApplicationCommand(
  File "E:\venv\lib\site-packages\nextcord\application_command.py", line 757, in __init__
  File "E:\venv\lib\site-packages\nextcord\application_command.py", line 387, in __init__
    self._from_callback(callback)
  File "E:\venv\lib\site-packages\nextcord\application_command.py", line 465, in _from_callback
    arg = CommandOption(value)
  File "E:\venv\lib\site-packages\nextcord\application_command.py", line 224, in __init__
  File "E:\venv\lib\site-packages\nextcord\application_command.py", line 263, in get_type
    raise NotImplementedError(f'Type "{typing}" isn\'t a supported typing for Application Commands.')
NotImplementedError: Type "str" isn't a supported typing for Application Commands.
```

### Intents

None

### System Information

- Python v3.10.0-final
- nextcord v2.0.0-alpha
    - nextcord pkg_resources: v2.0.0a5
- aiohttp v3.7.4.post0
- system info: Windows 10 10.0.19044

### Checklist

### Additional Context

The annotations future implicitly makes all type hints on the file it uses into forward references (essentially casting to string) in order to allow for type hints that reference things that have not been imported yet (useful for resolving circular imports) or are only partially constructed (e.g. annotating the class name during the definition of the class.)

According to [PEP-563][PEP-563], code using annotations should be using `typing.get_type_hints()` to correctly resolve type hints at runtime (as passing strings into type hints is already valid without the future import anyway.)

[PEP-563]: https://www.python.org/dev/peps/pep-0563/#resolving-type-hints-at-runtime

ooliver1:
## Summary

Fixes #328 code on behalf of @alentoghostflame 

## Checklist

- [x] This PR fixes an issue.


ooliver1:
ignroe those other commits, couldnt get them to go away but they made no code changes

TAG-Epic:
See https://github.com/Pycord-Development/pycord/pull/863 a few more issues is listed here

ooliver1:
That looks more like overcomplicating it than more issues resolved, what issues would overcomplicating it solve?

TAG-Epic:
> That looks more like overcomplicating it than more issues resolved, what issues would overcomplicating it solve?

Nonono look at the issues they had, maybe some apply here too

EmreTech:
Hello! As a person who is currently fixing the same issue in Pycord, I would like to say that using `eval` to get the type might not be the best solution. I know that Nextcord implements things differently than Pycord, but I think it would be better to use `typing.get_type_hints` and possibly redesign how Options are set so the type for the Option is set as the type hint.

ooliver1:
Thank you for that, closing this pr so alento can open a new one

PR #348, opened 2022-01-10
slash checks/hooks

okay1204:
## Summary

This PR adds application command checks, similar to the existing prefix command checks.

Changes:
 - adds global, cog, and local checks to the application command, along with all the premade ones such as checks.guild_only()
 - implements before and after invoke hooks
 - new client event, on_application_error, similar to on_command_error
 - adds on_error error handlers to application commands
 - adds attached of type InteractionAttached (more on that in the next line), application_command, client, and bot attributes to the Interaction class
 - adds InteractionAttached class, added to each instance of Interaction. Used to store any data within the interaction itself.
 - moves _get_overridden_method from commands.Cog to application_commands.ClientCog
 - moves owner_id, owner_ids, and is_owner method from commands.Bot to nextcord.Client

(The documentation and type hinting may need some review, not too sure if I did those correctly)

## Checklist

- [x] If code changes were made then they have been tested.
    - [ ] I have updated the documentation to reflect the changes.
- [ ] This PR fixes an issue.
- [x] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

PR #358, opened 2022-01-20
Add client to interaction class

DenverCoder1:
## Summary

Adds `client` as an attribute of `Interaction`

This functionality is part of the changes in PR #348 but, with permission from @TAG-Epic, I am adding this in its own PR so it can be merged sooner so that https://github.com/nextcord/nextcord-ext-menus/pull/27 can be released.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -177,7 +177,7 @@ class Interaction:
                 pass
 
     @property
-    def client(self):
+    def client(self) -> Client:
         """:class:`Client`: The client that handled the interaction."""
         return self._client
 
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -122,7 +122,6 @@ class Interaction:
         'guild_locale',
         'token',
         'version',
-        '_client',
         '_permissions',
         '_state',
         '_session',
@@ -135,7 +134,6 @@ class Interaction:
     def __init__(self, *, data: InteractionPayload, state: ConnectionState):
         self._state: ConnectionState = state
         self._session: ClientSession = state.http._HTTPClient__session
-        self._client: Client = state._get_client()
         self._original_message: Optional[InteractionMessage] = None
         self._from_data(data)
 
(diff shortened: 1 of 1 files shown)


nextcord/interactions.py
```diff
@@ -106,6 +107,11 @@ class Interaction:
         for 15 minutes.
     data: :class:`dict`
         The raw interaction data.
+    client: :class:`Client`
```
ooliver1, line 110: I don't think adding properties inside the attribute header works, properties docstrings should be enough

ooliver1 (changes requested):

nextcord/interactions.py
```diff
@@ -173,6 +181,16 @@ def _from_data(self, data: InteractionPayload):
             except KeyError:
                 pass
 
+    @property
+    def client(self):
+        """:class:`Client`: Returns the client that handled the interaction. An alias exists under ``bot``."""
+        return self._client
+
+    @property
+    def bot(self):
```
TAG-Epic, line 190: Lets just leave it be client

TAG-Epic (changes requested):

nextcord/interactions.py
DenverCoder1: Yeah, I think you're right about that. I was basing this on okay's PR which had this although their implementation was slightly different (adding it to the data rather than just getting it from state)

nextcord/interactions.py
```diff
@@ -173,6 +176,11 @@ def _from_data(self, data: InteractionPayload):
             except KeyError:
                 pass
 
+    @property
+    def client(self):
```
vcokltfre, line 180:
```suggestion
    def client(self) -> Client:
```

vcokltfre (changes requested):

vcokltfre (approved):

ooliver1 (approved):

nextcord/interactions.py
```diff
@@ -133,6 +135,7 @@ class Interaction:
     def __init__(self, *, data: InteractionPayload, state: ConnectionState):
         self._state: ConnectionState = state
         self._session: ClientSession = state.http._HTTPClient__session
+        self._client: Client = state._get_client()
```
TAG-Epic, line 138: This could probably be in the property itself?

TAG-Epic (changes requested):

PR #360, opened 2022-01-21
[User commands] Construct member if possible

Soheab:
## Summary

This PR rewrites `_handle_resolved_user` to construct a `Member` is the data if the data available.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -995,10 +995,26 @@ class ApplicationCommand(ApplicationSubcommand):
                 self._state._messages.append(message)
         return message
 
-    def _handle_resolved_user(self, user_data: dict) -> User:
-        """Takes the raw user data payload from Discord and adds it to the state cache."""
-        # This is what I prefer _handle_resolved_message to look like.
-        return self._state.store_user(user_data)
+    def _handle_resolved_user(self, resolved_payload: dict, guild: Optional[Union[Guild, int]] = None) -> Union[User, Member]:
+        """Takes the raw user data payload from Discord and adds it to the state cache.""" # needs changing?
+        user_id, user_payload = list(resolved_payload["users"].items())[0]
+        if not guild:
+            return self._state.store_user(user_payload)
+            
+        else:
+            if isinstance(guild, int):
+                guild = self._state.get_guild(guild)
+
+            member = guild.get_member(int(user_id))
+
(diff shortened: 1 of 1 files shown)


nextcord/application_command.py
```diff
@@ -995,10 +995,26 @@ def _handle_resolved_message(self, message_data: dict) -> Message:
...
-        """Takes the raw user data payload from Discord and adds it to the state cache."""
-        # This is what I prefer _handle_resolved_message to look like.
-        return self._state.store_user(user_data)
+    def _handle_resolved_user(self, resolved_payload: dict, guild: Optional[Union[Guild, int]] = None) -> Union[User, Member]:
+        """Takes the raw user data payload from Discord and adds it to the state cache.""" # needs changing?
+        user_id, user_payload = list(resolved_payload["users"].items())[0]
+        if not guild:
+            return self._state.store_user(user_payload)
+            
+        else:
```
onerandomusername, line 1004: you have an if that returns above, which is essentially a guard clause. That means there's no need to else this section which would make it easier to read.

nextcord/application_command.py
```diff
@@ -995,10 +995,26 @@ def _handle_resolved_message(self, message_data: dict) -> Message:
...
+            member = guild.get_member(int(user_id))
+
+            if not member and "members" in resolved_payload:
+                member_payload = list(resolved_payload["members"].values())[0]
+                # This is required to construct the Member.
+                member_payload["user"] = user_payload
+                member = Member(data=member_payload, guild=guild, state=self._state)  # type: ignore
+                guild._add_member(member)
+
+            return member
```
ooliver1, line 1017:
```suggestion
        if not guild:
            return self._state.store_user(user_payload)

        if isinstance(guild, int):
            guild = self._state.get_guild(guild)

        member = guild.get_member(int(user_id))

        if not member and "members" in resolved_payload:
            member_payload = list(resolved_payload["members"].values())[0]
            # This is required to construct the Member.
            member_payload["user"] = user_payload
            member = Member(data=member_payload, guild=guild, state=self._state)  # type: ignore
            guild._add_member(member)

        return member
```

ooliver1 (changes requested):

ooliver1 (approved):

Issue #373, opened 2022-01-25
default_permission unilaterally set to true for ApplicationCommands


hlieberman:
Regardless of the setting of default_permission when creating an ApplicationCommand, it will be unilaterally set to True internally during the creation of the object. 

https://github.com/nextcord/nextcord/blob/3ab52f348645e10369ddad45f409b311e000412a/nextcord/application_command.py#L812

Because False is falsy, a users' intent to disable the parameter instead results in the parameter being set to true.

For example:
```python3

@client.user_command(name="Test", default_permission=False)
async def test(interaction: nextcord.Interaction):
    await interaction.response.defer()
```

will be transformed into the ApplicationCommand object 
```
{'_callback': <function test at 0x7f7892592280>, 'parent_command': ..., 'type': <ApplicationCommandType.user: 2>, '_self_argument': ..., 'name': 'Test', '_description': ..., 'options': {}, 'children': {}, '_state': None, 'force_global': False, 'default_permission': True, '_guild_ids_to_rollout': set(), '_guild_ids': set(), '_command_ids': {}}
```

A one-liner to validate this (though ugly) is:
```python3
nextcord.application_command.ApplicationCommand(name="Test", default_permission=False, callback=test, cmd_type=nextcord.enums.ApplicationCommandType(2)).default_permission
```

Should probably be something like:
```python3
self.default_permission: bool = default_permission if default_permission is not MISSING else True
```

PR #383, opened 2022-01-26
Pass default_permission correctly (Closes: #373)

hlieberman:
## Summary
Correctly pass default_permission.

## Checklist

- [x] This PR fixes an issue.


hlieberman:
@alentoghostflame

TAG-Epic:
Maybe change from Missing to None as thats what its supposed to be?

hlieberman:
@TAG-Epic https://github.com/nextcord/nextcord/blob/2983dbcf88e5365eef3ebb319e7195a7f1ae1239/nextcord/application_command.py#L786 seems to indicate it will either be MISSING or True/False?

TAG-Epic:
Missing is for when setting None means something else. For example using None to set a users nick will reset it, thats why we use Missing there. None should be used otherwise

ooliver1:
im not sure what this solves, `MISSING` bool()'s to `False` doesnt it

hlieberman:
> Missing is for when setting None means something else. For example using None to set a users nick will reset it, thats why we use Missing there. None should be used otherwise

Converted to using None in f94b7b4

hlieberman:
PR failure is unrelated to changes; looks like a bug in darker itself.  Failed on master as well.

ooliver1:
linking an issue in title doesnt link properly

DenverCoder1:
Currently setting `default_permission` to `False` does nothing, since internally, it always overwrites the setting with `True` since it does 
```py
self.default_permission: bool = default_permission or True
```
This change makes `default_permission` configurable by the user since currently, it is not.

nextcord/application_command.py
```diff
@@ -809,7 +809,7 @@ def __init__(
         super().__init__(callback=callback, cmd_type=cmd_type, name=name, description=description)
         self._state: Optional[ConnectionState] = None
         self.force_global: bool = force_global
-        self.default_permission: bool = default_permission or True
+        self.default_permission: Optional[bool] = default_permission if default_permission is not None else True
```
DenverCoder1, line 812: Suggestion

See https://github.com/hlieberman/nextcord/blob/f94b7b44f8a43f948e833dae19b08f365b81de05/nextcord/application_command.py#L980-L981.

If default_permissions is None, it will use Discord's default, so this should not default to True at all.

```diff
- self.default_permission: Optional[bool] = default_permission if default_permission is not None else True
+ self.default_permission: Optional[bool] = default_permission
```

DenverCoder1, line 812:
```suggestion
        self.default_permission: Optional[bool] = default_permission
```

DenverCoder1:
The docstrings should have Optional for `default_permission` if it can be None
```diff
- default_permission: :class:`bool`
+ default_permission: Optional[:class:`bool`]
```

ooliver1 (approved):

ooliver1 (approved):

PR #388, opened 2022-01-26
perf: Reduce API calls for delete_after

DenverCoder1:
## Summary

Moves the delay code from `InteractionMessage.delete` to `Interaction.delete_original_message`. `InteractionMessage.delete` can still take a delay.

This reduces API calls for `InteractionResponse.send_message`, `InteractionResponse.edit_message` when using `delete_after`

`delay` was added to delete_original_interaction so that the message does not need to be fetched first.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -356,13 +356,19 @@ class Interaction:
             self._state.store_view(view, message.id)
         return message
 
-    async def delete_original_message(self) -> None:
+    async def delete_original_message(self, *, delay: Optional[float] = None) -> None:
         """|coro|
 
         Deletes the original interaction response message.
 
         This is a lower level interface to :meth:`InteractionMessage.delete` in case
         you do not want to fetch the message and save an HTTP request.
+
+        Parameters
+        -----------
+        delay: Optional[:class:`float`]
+            If provided, the number of seconds to wait before deleting the message.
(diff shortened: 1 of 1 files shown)


nextcord/interactions.py
```diff
@@ -372,12 +378,23 @@ async def delete_original_message(self) -> None:
...
         adapter = async_context.get()
-        await adapter.delete_original_interaction_response(
+        delete_func = adapter.delete_original_interaction_response(
             self.application_id,
             self.token,
             session=self._session,
         )
 
+        if delay is not None:
+            async def inner_call(delay: float = delay):
```
TAG-Epic, line 388: Maybe use call_later?

TAG-Epic (changes requested):

nextcord/interactions.py
DenverCoder1: This is the code from `InteractionMessage.delete` but repurposed.

I think I'd need the event loop to use call_later? Not sure how it would be changed.

DenverCoder1: I suppose it could be changed to use self._state.loop.call_later but it's not much different

```py
async def inner_call(delay: float = delay):
    try:
        await delete_func
    except HTTPException:
        pass
self._state.loop.call_later(delay, inner_call)
```

PR #394, opened 2022-01-28
[pending] feat: add ability to get message returned when sending an interaction response

DenverCoder1:
## Summary

This makes Interactions consistent with Webhooks by allowing the user to request the message be returned with `wait=True`.

Currently `Messageable.send` and `WebhookMessage.send` return their respective messages but `Interaction.response.send_message` never does.

For an interaction that was already sent, `Interaction.send` returns a `WebhookMessage`, but for an interaction response, nothing is returned. The docs were previously incorrect saying it would return a `Message` when really it returns `None`.

It would be useful if you could get the message at the same time as sending the interaction message.

So if you want the message, you could do:
```py
message = await interaction.send(..., wait=True)
```
Instead of
```py
await interaction.send(...)
message = await interaction.original_message()
```
This makes it easier for the user of the lib (similar to why we have `Interaction.send` in the first place)

**By adding a `wait` parameter, this allows requesting that the message be returned so that unnecessary API calls are not made unless specifically requested.**

## Checklist

- [ ] This PR fixes an issue.

image

Test code:
```py
async def send(interaction: Interaction):
    message = await interaction.send("Test", wait=True) # request message with wait=True
    await interaction.channel.send(f"A {type(message)} was sent")

async def defer_send(interaction: Interaction):
    await interaction.response.defer()
    await sleep(1)
    message = await interaction.send("Test") # this can already return the Message even if wait is False
    await interaction.channel.send(f"A {type(message)} was sent")

async def send_no_wait(interaction: Interaction):
    await interaction.send("Test") # returns None
```

### DIFF
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -406,6 +406,7 @@ class Interaction:
         view: View = MISSING,
         tts: bool = False,
         ephemeral: bool = False,
+        wait: bool = False,
         delete_after: Optional[float] = None,
     ) -> Optional[Union[InteractionMessage, WebhookMessage]]:
         """|coro|
@@ -423,7 +424,7 @@ class Interaction:
         """
 
         if not self.response.is_done():
-            await self.response.send_message(
+            return await self.response.send_message(
                 content=content,
                 embed=embed,
                 embeds=embeds,
@@ -432,9 +433,9 @@ class Interaction:
                 view=view,
                 tts=tts,
                 ephemeral=ephemeral,
+                wait=wait,
                 delete_after=delete_after,
             )
-            return await self.original_message()
         return await self.followup.send(
             content=content,  # type: ignore
             embed=embed,
@@ -444,6 +445,7 @@ class Interaction:
             view=view,
             tts=tts,
             ephemeral=ephemeral,
+            wait=wait,
             delete_after=delete_after,
         )
 
@@ -624,8 +626,9 @@ class InteractionResponse:
         view: View = MISSING,
         tts: bool = False,
         ephemeral: bool = False,
+        wait: bool = False,
         delete_after: Optional[float] = None,
-    ) -> None:
+    ) -> Optional[InteractionMessage]:
         """|coro|
 
         Responds to this interaction by sending a message.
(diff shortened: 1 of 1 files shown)


vcokltfre:
I don't think this should be making extra API calls behind the user's back, the reason `Messageable.send()` returns a message is that the API endpoint does, the same is not true of interaction responses, which is exactly why to this point the behaviour is this way. Implementing it like this is always going to catch people out who are expecting just one request to be made, and could cause issues in larger bots by hitting rate limits. In contrast, using the `send()` method and then fetching the original response separately makes the functionality very clear and people both gain knowledge of how the API works and are not caught out by the extra API calls being made. Furthermore, most of the time it is not necessary to have this message available, indeed in ephemeral messages, you don't get the option to edit, delete, or react to the message, and thus having it is near enough pointless in that context.

As a convenience method many people will use `Interaction.send()`, because of this I believe having unexpected behaviour in it is more harmful than good.

DenverCoder1:
@vcokltfre I've updated the code to have a `wait` parameter. This makes it consistent with Webhooks and does not make extra API calls unless the user specifically requests it

ooliver1 (approved):

nextcord/interactions.py
```diff
@@ -417,9 +418,10 @@ async def send(
 
         Returns
         -------
-        Optional[:class:`Message`, :class:`WebhookMessage`]
-            The :class:`Message` that was sent, a :class:`WebhookMessage` if the
-            interaction has been responded to before.
+        Optional[:class:`InteractionMessage`, :class:`WebhookMessage`]
```
TAG-Epic, line 421: Doesnt this have to be a union?

nextcord/interactions.py
```diff
@@ -729,6 +743,9 @@ async def send_message(
         if delete_after is not None:
             await self._parent.delete_original_message(delay=delete_after)
 
+        if wait:
+            return await self._parent.original_message()
```
TAG-Epic, line 747: Isnt there a wait=True thingy in send webhooks or does that not work here? I dont think we should be making requests without the knowledge of the user.

TAG-Epic (changes requested):

nextcord/interactions.py
DenverCoder1: Unfortunately, it seems like it won't work here. I've tried it, but when sending an interaction response, Discord responds with no content (204 response) meaning that it seems the only way to get the data is with an extra request

DenverCoder1:
I think I agree with the points above and we can hold off on this until Discord decides to return the message data in the response. It seems like this is the standard approach among most libs (eg. in djs, you need to use `.fetchReply()` to get the reply message.)

Djs does, however, return the MessagePayload from their [reply method](https://discord.js.org/#/docs/discord.js/stable/class/CommandInteraction?scrollTo=reply). Possibly instead of returning the InteractionMessage, we could make it return something similar, although I don't see how it would be all that helpful.

PR #405, opened 2022-01-29
Message embed parameter gets now send as embeds

veni-vidi-code:
## Summary

Discord Docs say embed field is depreceated so sending embed field as embeds
(https://discord.com/developers/docs/resources/channel#create-message)

All other parts of the libary i could find already do so

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -1335,7 +1335,8 @@ class Messageable:
             raise InvalidArgument('cannot pass both embed and embeds parameter to send()')
 
         if embed is not None:
-            embed = embed.to_dict()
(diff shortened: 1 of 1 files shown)


Soheab:
That's a local variable, what does that have to do with sending? embeds are send as "embeds" since discord 2.0 

See the sending part [here](https://github.com/nextcord/nextcord/blob/master/nextcord/http.py#L441-L442) and the commit where support for multiple embeds was added [here](https://github.com/Rapptz/discord.py/commit/47e6a754e496a811df0f29c49bf3fd9512d20021)

PR #410, opened 2022-01-31
feat: Alternate syntax for optional slash args

DenverCoder1:
## Summary

Allows creating optional arguments using `Optional` type annotations.

Previously, specifying `Optional` in a type annotation at all would give an error:
```
Type "typing.Optional[int]" isn't a supported typing for Application Commands.
```

This change not only allows annotating optional SlashOptions as `Optional`, but it also allows creating optional arguments without using the `SlashOption` class.

Example:

```py
async def enter_a_number(interaction: nextcord.Interaction, number: Optional[int] = None):
    if not number:
        await interaction.response.send_message("You need to specify a number!", ephemeral=True)
    else:
        await interaction.response.send_message(f"You chose {number}!")
```

image

image

## Checklist

- [x] This PR fixes an issue.


TAG-Epic (approved):
Not tested

ooliver1 (approved):

ooliver1:
on testing, ```py
async def slashe(
    interaction,
    number: Optional[int],
    qwerty: str = discord.SlashOption(name="letter", description="hm"),
):```
```py
  File "test.py", line 65, in <module>
    async def slashe(
  File "/home/ooliver/projects/TEST/env/lib/python3.8/site-packages/nextcord/client.py", line 2056, in decorator
  File "/home/ooliver/projects/TEST/env/lib/python3.8/site-packages/nextcord/application_command.py", line 1293, in decorator
    app_cmd = ApplicationCommand(
  File "/home/ooliver/projects/TEST/env/lib/python3.8/site-packages/nextcord/application_command.py", line 813, in __init__
  File "/home/ooliver/projects/TEST/env/lib/python3.8/site-packages/nextcord/application_command.py", line 428, in __init__
    self._from_callback(callback)
  File "/home/ooliver/projects/TEST/env/lib/python3.8/site-packages/nextcord/application_command.py", line 516, in _from_callback
    arg = CommandOption(value)
  File "/home/ooliver/projects/TEST/env/lib/python3.8/site-packages/nextcord/application_command.py", line 229, in __init__
  File "/home/ooliver/projects/TEST/env/lib/python3.8/site-packages/nextcord/application_command.py", line 267, in get_type
    elif typing._name == Optional.__name__ and (valid_type := self.option_types.get(typing.__args__[0], None)):
AttributeError: '_SpecialForm' object has no attribute '__name__'``` @DenverCoder1

**Comment by ooliver1** [comment]
in the repl too (vps 3.8 but locally with 3.9 too
```pycon
SyntaxError: invalid syntax
>>> Optional.__name__
  File "<stdin>", line 1, in <module>
AttributeError: '_SpecialForm' object has no attribute '__name__'
>>>
```

ooliver1:
@DenverCoder1 suggested fix, `Optional._name`

TAG-Epic (approved):
Not tested

DenverCoder1:
I've tested with both `Optional[...]` and `Union[..., None]` using Python version 3.8, 3.9 and 3.10

nextcord/application_command.py
```diff
@@ -209,6 +211,9 @@ def __init__(self, parameter: Parameter):
         # self.required = cmd_arg.required or MISSING
         # will cause self.required to be MISSING, not False.
         self.required = cmd_arg.required if cmd_arg.required is not MISSING else MISSING
+        # Set required to False if an Optional[...] or Union[..., None] type annotation is given.
+        if type(None) in typing.get_args(parameter.annotation):
+            self.required = False
```
alentoghostflame, line 216: What the `required` kwarg is manually set to (by the user) should always override this.
I recommend checking if `cmd_arg.required` is `MISSING` before checking the typing args.

alentoghostflame (changes requested):
Other than that one bit, it looks pretty good!

nextcord/application_command.py
DenverCoder1: I think it could go either way since there's no reason to annotate a parameter as Optional if the argument itself is actually required.

I can easily swap them though.

PR #423, opened 2022-02-04
Add check for confirming datetime instance, as well as adding a timeout method to member.py

Shlol762:
## Summary

Just checks if the argument `dt` is an instance of `datetime` to prevent confusion.
Adds a timeout method to member.py as a direct method instead of having to go through `Member.edit`

## Checklist

- [x ] If code changes were made then they have been tested.
    - [  ] I have updated the documentation to reflect the changes.
- [  ] This PR fixes an issue.
- [  ] This PR adds something new (e.g. new method or parameters).
- [  ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [  ] This PR is **not** a code change (e.g. documentation, README, ...)

### DIFF
--- a/nextcord/member.py
+++ b/nextcord/member.py
@@ -667,6 +667,13 @@ class Member(abc.Messageable, _UserTag):
 
         Times out this member.
 
+        .. note::
+
+            This is a more direct method of timing out a member.
+            You can also time out members using :meth:`Member.edit`.
+
+        .. versionadded:: 2.0
+
         Parameters
         -----------
         timeout: Optional[Union[:class:`~datetime.datetime`, :class:`~datetime.timedelta`]]
@@ -674,13 +681,6 @@ class Member(abc.Messageable, _UserTag):
             Set this to None to disable their timeout.
         reason: Optional[:class:`str`]
             The reason for editing this member. Shows up on the audit log.
-
-        .. note::
-
-            This is a more direct method of timing out a member.
-            You can also time out members using :meth:`Member.edit`.
-
-        .. versionadded:: 2.0
         """
         await self.edit(timeout=timeout, reason=reason)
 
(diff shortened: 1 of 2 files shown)


Skelmis (approved):
lgtm

nextcord/member.py
```diff
@@ -656,6 +656,29 @@ async def kick(self, *, reason: Optional[str] = None) -> None:
         Kicks this member. Equivalent to :meth:`Guild.kick`.
         """
         await self.guild.kick(self, reason=reason)
+        
+    async def timeout(self, timeout: Union[datetime.datetime, datetime.timedelta],
+                      *, reason: Optional[str] = None):
```
vcokltfre, line 661: This function formatting is inconsistent with other methods in this file, could it be changed to match them please.
```suggestion
    async def timeout(
        self,
        timeout: Union[datetime.datetime, datetime.timedelta],
        *,
        reason: Optional[str] = None,
    ) -> None:
```

vcokltfre (changes requested):

nextcord/member.py
```diff
@@ -656,6 +656,33 @@ async def kick(self, *, reason: Optional[str] = None) -> None:
...
+            Set this to None to disable their timeout.
+        reason: Optional[:class:`str`]
+            The reason for editing this member. Shows up on the audit log.
+
+        .. note::
+
+            This is a more direct method of timing out a member.
+            You can also time out members using :meth:`Member.edit`.
+
+        .. versionadded:: 2.0
```
ooliver1, line 683: this seems to render odd on the docs 
```suggestion
        """|coro|

        Times out this member.

        .. note::

            This is a more direct method of timing out a member.
            You can also time out members using :meth:`Member.edit`.

        .. versionadded:: 2.0

        Parameters
        -----------
        timeout: Optional[Union[:class:`~datetime.datetime`, :class:`~datetime.timedelta`]]
            The time until the member should not be timed out.
            Set this to None to disable their timeout.
        reason: Optional[:class:`str`]
            The reason for editing this member. Shows up on the audit log.
```

ooliver1 (approved):

PR #426, opened 2022-02-05
`from __future__ import annotations` Slash fix.

### LINKED ISSUE #328: Client.slash_command fails when annotations future is enabled

alentoghostflame:
## Summary

Fixes #328, creating an ApplicationCommand from a callback now properly converts string annotations into the correct objects.

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -24,6 +24,7 @@ DEALINGS IN THE SOFTWARE.
 
 from __future__ import annotations
 import asyncio
+import typing
 from inspect import signature, Parameter
 from typing import (
     Any,
@@ -190,7 +191,7 @@ class CommandOption(SlashOption):
         float: ApplicationCommandOptionType.number,
         Message: ApplicationCommandOptionType.integer  # TODO: This is janky, the user provides an ID or something? Ugh.
     }
-    """Maps Python typings to Discord Application Command typings."""
+    """Maps Python annotations/typehints to Discord Application Command type values."""
     def __init__(self, parameter: Parameter):
         super().__init__()
         self.parameter = parameter
@@ -225,8 +226,8 @@ class CommandOption(SlashOption):
             self.default = cmd_arg.default
 
         self.autocomplete_function: Optional[Callable] = MISSING
-
         self.type: ApplicationCommandOptionType = self.get_type(parameter.annotation)
+
         if cmd_arg._verify:
             self.verify()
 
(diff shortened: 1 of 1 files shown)


TAG-Epic (approved):
Not tested

ooliver1 (approved):

PR #444, opened 2022-02-11
Slash no description fix

alentoghostflame:
## Summary

Discord, seemingly silently, made it so having the description of slash commands and options only be a whitespace character will fail. This PR fixes that by setting the default description to "No description provided"

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -237,7 +237,7 @@ class CommandOption(SlashOption):
     def description(self) -> str:
         """If no description is set, it returns the bare minimum that Discord demands."""
         if not self._description:
-            return " "
+            return "No description provided"
         else:
             return self._description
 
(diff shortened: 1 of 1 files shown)


nextcord/application_command.py
```diff
@@ -237,7 +237,7 @@ def __init__(self, parameter: Parameter):
     def description(self) -> str:
         """If no description is set, it returns the bare minimum that Discord demands."""
```
Skelmis, line 238: This should be changed to not if it's not provided the new return is used

Skelmis (changes requested):

nextcord/application_command.py
alentoghostflame: Fixed!

Skelmis (approved):

nextcord/application_command.py
```diff
@@ -433,9 +433,11 @@ def error_name(self) -> str:
 
     @property
     def description(self) -> str:
-        """Returns the description of the command. If the description is MISSING, it returns the bare minimum needed."""
+        """
+        Returns the description of the command. If the description is MISSING, it returns "No description provided"
+        """
         if self._description is MISSING:  # Return Discords bare minimum for a command.
```
DenverCoder1, line 439:
```suggestion
        if self._description is MISSING: 
```

DenverCoder1 (approved):
These comments are not necessary

PR #448, opened 2022-02-12
feature: Allow passing an awaitable `after` function to `AudioPlayer`

anoadragon453:
## Summary

This enables awaitable (async) functions to be called after an audio file has finished playing in voice calls.

I have been long frustrated with needing to provide a non-awaitable function to `VoiceClient.play`, only to have that function need to use `asyncio.run_coroutine_threadsafe` to further call async code.

By checking whether we receive an awaitable when calling `self.after`, and awaiting on it, we add support for awaitable functions.

I have lightly tested this with success in my own Discord bot. Providing both async and non-async functions as the `after` arugment to `VoiceClient.play` has resulted in them both being called successfully.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -9,8 +9,12 @@ jobs:
       - uses: actions/checkout@v2
         with:
           fetch-depth: 0
-      - uses: akaihola/darker@github-action-v1.3.2-2
+      - uses: akaihola/darker@1.4.0
         with:
           options: "--check --verbose"
           src: "./nextcord"
-          version: "1.3.2"
+          # TODO: Remove "version" on the next darker release.
+          # By default, this should use the same version of darker as written above,
+          # but v1.4.0 of the github action had a bug where darker 1.3.2 was still
+          # being used. See https://github.com/akaihola/darker/pull/282 for details.
+          version: "1.4.0"
--- a/nextcord/message.py
+++ b/nextcord/message.py
@@ -1007,7 +1007,7 @@ class Message(Hashable):
         return self.type not in (
             MessageType.default,
             MessageType.reply,
-            MessageType.application_command,
+            MessageType.chat_input_command,
             MessageType.thread_starter_message,
         )
 


anoadragon453:
CI failure appears unrelated to this code change.

Skelmis (approved):
Lgtm, I haven't tested.

anoadragon453:
After digging into asyncio some more, I found that [`asyncio.run_coroutine_threadsafe`](https://docs.python.org/3/library/asyncio-task.html#asyncio.run_coroutine_threadsafe) will [only accept](https://github.com/python/cpython/blob/7a0486eaa98083e0407ff491872db6d7a0da2635/Lib/asyncio/tasks.py?q=%22Submit+a+coroutine+object+to+a+given+event+loop%22+org%3Apython#L870-L871) a coroutine, not *any* Awaitable type (Tasks, Futures...). Therefore, I've reduced the scope of what we'll run to only coroutines in https://github.com/nextcord/nextcord/pull/448/commits/e9954332e871253fe144fb724bdee863c636fe9f.

@Skelmis thank you for the suggestion! I've updated the documentation in https://github.com/nextcord/nextcord/pull/448/commits/fbfab79c8762143ee3890e40e2858d5b039e0c64 to note that coroutines may also be passed.

nextcord/voice_client.py
```diff
@@ -574,6 +574,7 @@ def play(self, source: AudioSource, *, after: Callable[[Optional[Exception]], An
             The finalizer that is called after the stream is exhausted.
             This function must have a single parameter, ``error``, that
             denotes an optional exception that was raised during playing.
+            If the function is a coroutine, it will be awaited when called.
```
Skelmis, line 577: Do we want to also change the types to include coros?

anoadragon453: What do you suggest?

If `Callable[[Optional[Exception]], Union[Coroutine, Any]]`: `Coroutine` is included in `Any`, so I don't think it's necessary to pass type checkers. It might be useful for humans, but that's what the docstring is for.

https://github.com/python/typing/issues/424 has some notes about this, though as discovered earlier we don't want a scope as large as `Awaitable`.

ooliver1: Yeah it would have some form of Any for sync functions, so type checkers ignore it

anoadragon453: I'm not really sure what further action to take here.

nextcord/player.py
```diff
@@ -683,7 +695,13 @@ def _call_after(self) -> None:
 
         if self.after is not None:
             try:
-                self.after(error)
+                # Run the after function
+                after_return = self.after(error)
+
+                # If what we got back was a coroutine, submit it to
+                # the main event loop for processing
+                if asyncio.coroutines.iscoroutine(after_return):
```
MaskDuck, line 703: why dont use `inspect.isawaitable`?

MaskDuck (changes requested):

nextcord/player.py
Skelmis: Jumping on this, since I missed that. Its been discussed in some help threads, but the usage of `asyncio.iscoroutinefunction` looked to be the way for cython based support as well.

MaskDuck: ah ok

MaskDuck (approved):

ooliver1 (approved):

PR #458, opened 2022-02-15
Add global scope for slash commands

Dysta:
## Summary

This PR add the ability for users to define a global scope for their slash commands instead of filling in the `guilds_id` parameter in each slash command decorator.

### Behavior: 
If user fill the `guild_ids` parameter for a slash command, it overwrite the `rollout_guild_ids` and the command is only available in the selected guild.

### Example:
```py

bot = commands.Bot(
    command_prefix="test ",
    rollout_guild_ids=[957587943717994586],
)


@bot.slash_command(name="test")
async def slashe(
    interaction,
    number: int,
    letter: str,
):
    await interaction.response.send_message(f"Helo {number} {letter}")


@bot.slash_command(name="testistos", guild_ids=[957598850691006544])
async def other_slashe(
    interaction,
    number: int,
):
    await interaction.response.send_message(f"Henllooooo {number}")

bot.run("token")
```
#### Expected behavior:
`957587943717994586` have the two command registered
image

`957598850691006544` have only the second command registered
image

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/__init__.py
+++ b/nextcord/__init__.py
@@ -14,7 +14,7 @@ __title__ = 'discord'
 __author__ = 'tag-epic & Rapptz'
 __license__ = 'MIT'
 __copyright__ = 'Copyright 2015-present Rapptz'
-__version__ = '2.0.0a7'
+__version__ = '2.0.0a8'
 
 __path__ = __import__('pkgutil').extend_path(__path__, __name__)
 


DenverCoder1:
I'd probably get rid of the word "test" since it's not necessarily for testing purposes (you could have a bot that only supports 1 or a few servers)

Maybe something like `rollout_guild_ids` to be consistent with the other slash kwargs?

DenverCoder1:
Does this work in Cogs?

Dysta:
> Does this work in Cogs?

See this point with Alento and it should with the last modification. 
Not tested actually btw

PR #463, opened 2022-02-15
Make utils.get content-aware

frissyn:
## Summary

This pull request aims to make `nextcord.utils.get` aware of the attribute names passed to it.

Prior to this change, `nextcord.utils.get` would blindly replace the `__` delimiter with `.` causing special variables (`__class__`, `__name__`, etc.) to throw an error. I've updated the `str.replace("__", "_")` to use regex to look for edge cases that contain special variables. If the name does not contain special variables, the regular expression will use a lookahead statement for underscores, and greedily replace them with `.`, assuming trailing underscores.

Previous behavior:
```py
>>> fmt("guild__name")
'guild.name'
>>> fmt("trailing___attr")
'trailing._attr'
>>> fmt("__class______name__")
'.class...name.'
```
New behavior:
```py
>>> fmt("guild__name")
'guild.name'
>>> fmt("trailing___attr")
'trailing_.attr'
>>> fmt("__class______name__")
'__class__.__name__'
>>> fmt("__class______name______type__")
'__class__.__name__.__type__'
```

Although these are pretty dramatic changes to behavior, it should be non-breaking. 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/utils.py
+++ b/nextcord/utils.py
@@ -393,10 +393,9 @@ def _key_fmt(key: str) -> str:
     # ex. "_privateattr___subattr" -> "_privateattr_.subattr"
 
     if key.startswith("__") and key.endswith("__"):
-        return re.sub(r"((__){1}__(__){1})", "__.__", key)
+        return re.sub(r"_{6}", "__.__", key)
     else:
-        slices = re.split(r"(?=__)", key)
-        return "".join([s.replace("__", ".") for s in slices])
+        return re.sub(r"__(?!_)", ".", key)
 
 
 def get(iterable: Iterable[T], **attrs: Any) -> Optional[T]:
--- a/nextcord/utils.py
+++ b/nextcord/utils.py
@@ -384,6 +384,21 @@ def find(predicate: Callable[[T], Any], seq: Iterable[T]) -> Optional[T]:
     return None
 
 
+def _key_fmt(key: str) -> str:
+    # Private helper for `nextcord.utils.get`. Formats the attribute
+    # names in a content aware manner. When special variables are
+    # provided, double trailing and leading underscores are ignored.
+    # ex. "__class______name__" -> "__class__.__name__"
+    # In case of underscore conflicts, it will assume trailing underscores.
+    # ex. "_privateattr___subattr" -> "_privateattr_.subattr"
+
+    if key.startswith("__") and key.endswith("__"):
+        return re.sub(r"((__){1}__(__){1})", "__.__", key)
+    else:
+        slices = re.split(r"(?=__)", key)
+        return "".join([s.replace("__", ".") for s in slices])
+
+
 def get(iterable: Iterable[T], **attrs: Any) -> Optional[T]:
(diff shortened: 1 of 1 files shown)


frissyn:
Ignore how messed up the author field is :S

TAG-Epic (approved):
Not tested

nextcord/utils.py
```diff
@@ -384,6 +384,21 @@ def find(predicate: Callable[[T], Any], seq: Iterable[T]) -> Optional[T]:
...
+def _key_fmt(key: str) -> str:
+    # Private helper for `nextcord.utils.get`. Formats the attribute
+    # names in a content aware manner. When special variables are
+    # provided, double trailing and leading underscores are ignored.
+    # ex. "__class______name__" -> "__class__.__name__"
+    # In case of underscore conflicts, it will assume trailing underscores.
+    # ex. "_privateattr___subattr" -> "_privateattr_.subattr"
+
+    if key.startswith("__") and key.endswith("__"):
+        return re.sub(r"((__){1}__(__){1})", "__.__", key)
```
DenverCoder1, line 396: What is the point of `{1}` here? Matching exactly 1 time seems kind of meaningless.

DenverCoder1: As the capturing groups are unused, I think simply `_{6}` would do the exact same thing as `((__){1}__(__){1})` here.

nextcord/utils.py
```diff
@@ -384,6 +384,21 @@ def find(predicate: Callable[[T], Any], seq: Iterable[T]) -> Optional[T]:
...
+    else:
+        slices = re.split(r"(?=__)", key)
+        return "".join([s.replace("__", ".") for s in slices])
```
DenverCoder1, line 399: I may be wrong, but can't this be simplified using a single regex.

It seems like you are trying to make it so that a `__` that's not followed by an underscore will be replaced with a `.`, so:

```py
re.sub(r"__(?!_)", ".", key)
```

This will make `trailing___attr` output `trailing_.attr`.

frissyn: I'll definitely admit that I only have a very basic understanding of regex, so there definitely is room to simply my expressions. You're right that I used a greedy lookahead to match `__` not followed by another underscore.

nextcord/utils.py
frissyn: Yup, the regular expression can definitely be simplified.

DenverCoder1: Regex might not even be necessary here since `str.replace("______", "__.__")` should also work. The regex may be more readable, though, since underscores can be hard to count. ¯\\_(ツ)\_/¯

frissyn: Fixed with my most recent commit, changes tested and work with the same behavior.

DenverCoder1 (approved):

ooliver1 (approved):
not tested, such a funny pr fix however

frissyn:
The triple underscores `______` is too good to pass up lol

PR #466, opened 2022-02-15
fix: kick using X Audit Reason header

veni-vidi-code:
Api V10 will not support the old reason param, so it needs to use the X-Audit-Log-Reason header aswell. 

Please note: This is not to the master branch. THe necessary changes for this to work properly will be added to the branch and will not be added in this pr

### DIFF
--- a/nextcord/guild.py
+++ b/nextcord/guild.py
@@ -2266,7 +2266,7 @@ class Guild(Hashable):
 
         payload['tags'] = emoji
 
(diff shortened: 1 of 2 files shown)


veni-vidi-code:
i can not tell why it wants to add the api bump again

DenverCoder1:
You could try doing a pull and rebase, but I don't think it matters that much since there is no conflict

DenverCoder1 (approved):

ooliver1 (approved):

PR #471, opened 2022-02-16
fix subscription is so so so so low

### LINKED ISSUE #416: Make 400: "Guild Premium subscription level too low" less ambiguous when creating a thread
### Summary

`start_thread_without_message` can raise an HTTPException which is not well documented

The core library

### The Problem

Currently, when creating a thread without a message, the default type for that thread is `ChannelType.private_thread`. This is consistent with the [official api](https://discord.com/developers/docs/resources/channel#start-thread-without-message).

However, if the guild in which this thread is being created is not boosted to level 2, this fails with the following exception: `nextcord.errors.HTTPException: 400 Bad Request (error code: 20035): Guild premium subscription level too low`

This is vague and not well documented, as it is the same error received when attempting to create a thread with too high of a `auto_archive_duration`. Also, despite `TextChannel.create_thread` being a seemingly simple function, its default value for the `type` parameter could be a common source of headache in guilds with a boost lower than 2. It is not mentioned in the nextcord documentation that this error could occur.

### The Ideal Solution

1. Mention in the documentation for `TextChannel.create_thread` that the default value should be changed for a guild with boost level lower than 2, or:

2. Raise an exception (perhaps `InvalidArgument`?) when a private_thread is created and "PRIVATE_THREADS" is not in `TextChannel.guild.features`

### The Current Solution

Nothing, user gets an exception: `nextcord.errors.HTTPException: 400 Bad Request (error code: 20035): Guild premium subscription level too low`

### Additional Context

- [Stack overflow post](https://stackoverflow.com/questions/70937404/starting-thread-in-discord-using-pycord/) where this issue arises.

- In APIv10, `start_thread_without_message` will have no default value for `type`.

MaskDuck:
## Summary

closes #416 
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/channel.py
+++ b/nextcord/channel.py
@@ -691,7 +691,7 @@ class TextChannel(abc.Messageable, abc.GuildChannel, Hashable):
 
         .. note::
 
-            This will raise :exc:`HTTPException` if the guild features does not include ``PRIVATE_THREADS`` and you are trying to create a private thread.
+            This will raise :exc:`HTTPException` if the guild features does not have the feature ``PRIVATE_THREADS`` and you are trying to create a private thread.
 
         Parameters
         -----------
--- a/nextcord/channel.py
+++ b/nextcord/channel.py
@@ -714,9 +714,7 @@ class TextChannel(abc.Messageable, abc.GuildChannel, Hashable):
         HTTPException
             Starting the thread failed.
 
-        .. note::
-        
-            This will raise :exc:`HTTPException` if the guild does not have the feature ``PRIVATE_THREADS`` and you are trying to create a private thread.
+            This can be raised if the guild does not have the feature ``PRIVATE_THREADS`` and you are trying to create a private thread.
 
         Returns
         --------
--- a/nextcord/channel.py
+++ b/nextcord/channel.py
@@ -689,10 +689,6 @@ class TextChannel(abc.Messageable, abc.GuildChannel, Hashable):
 
         .. versionadded:: 2.0
 
-        .. note::
-
-            This will raise :exc:`HTTPException` if the guild features does not have the feature ``PRIVATE_THREADS`` and you are trying to create a private thread.
-
         Parameters
         -----------
         name: :class:`str`
@@ -717,6 +713,11 @@ class TextChannel(abc.Messageable, abc.GuildChannel, Hashable):
             You do not have permissions to create a thread.
         HTTPException
             Starting the thread failed.
+
+        .. note::
+        
+            This will raise :exc:`HTTPException` if the guild does not have the feature ``PRIVATE_THREADS`` and you are trying to create a private thread.
+
 
         Returns
         --------
--- a/nextcord/channel.py
+++ b/nextcord/channel.py
@@ -715,7 +715,7 @@ class TextChannel(abc.Messageable, abc.GuildChannel, Hashable):
             Starting the thread failed.
 
             .. note:: 
-
+                
                 This can include if the guild isn't at least level two boosting yet and you are trying to create a private thread.
 
         Returns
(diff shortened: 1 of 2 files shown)


nextcord/channel.py
```diff
@@ -740,6 +743,7 @@ async def create_thread(
                 reason=reason,
             )
 
+
```
ooliver1, line 746: Unrelated change?

MaskDuck: it shouldnt be here

MaskDuck: removed =)

ooliver1 (approved):

DenverCoder1 (approved):

nextcord/channel.py
```diff
@@ -689,6 +689,10 @@ async def create_thread(
 
         .. versionadded:: 2.0
 
+        .. note::
+
```
TAG-Epic, line 693: Unneeded blank line?

nextcord/channel.py
```diff
@@ -689,6 +689,10 @@ async def create_thread(
...
+            This will raise :exc:`HTTPException` if the guild isn't at least level two boosting yet and you are trying to create a private thread.
```
TAG-Epic, line 694: This is technically not true as it checks for the PRIVATE_THREADS feature (which can also be given manually by discord staff). Also yet should be removed here

TAG-Epic (changes requested):
Not visible on Message.create_thread

nextcord/channel.py
ooliver1: No it is not, see https://github.com/nextcord/nextcord/blob/master/nextcord/guild.py#L188-L190 for example

nextcord/channel.py
MaskDuck: wdym?

nextcord/channel.py
```diff
@@ -689,6 +689,10 @@ async def create_thread(
 
         .. versionadded:: 2.0
 
+        .. note::
+
+            This will raise :exc:`HTTPException` if the guild features isn't contain ``PRIVATE_THREADS`` and you are trying to create a private thread.
```
DenverCoder1, line 694:
```suggestion
            This will raise :exc:`HTTPException` if the guild features does not include ``PRIVATE_THREADS`` and you are trying to create a private thread.
```

nextcord/message.py
```diff
@@ -1555,6 +1555,10 @@ async def create_thread(self, *, name: str, auto_archive_duration: ThreadArchive
 
         .. versionadded:: 2.0
 
+        .. note::
+
+            This will raise :exc:`HTTPException` if the guild features isn't contain ``PRIVATE_THREADS`` and you are trying to create a private thread.
```
DenverCoder1, line 1560:
```suggestion
        .. note::

            This will raise :exc:`HTTPException` if the guild features does not include ``PRIVATE_THREADS`` and you are trying to create a private thread.
```

DenverCoder1 (approved):

nextcord/message.py
```diff
@@ -1555,6 +1555,10 @@ async def create_thread(self, *, name: str, auto_archive_duration: ThreadArchive
 
         .. versionadded:: 2.0
 
+        .. note::
+
+            This will raise :exc:`HTTPException` if the guild features does not include ``PRIVATE_THREADS`` and you are trying to create a private thread.
```
TAG-Epic, line 1560:
```suggestion
            This will raise :exc:`HTTPException` if the guild does not have the feature ``PRIVATE_THREADS`` and you are trying to create a private thread.
```

TAG-Epic (changes requested):

ooliver1 (approved):
for the future, maybe instead of a billion commits instead make sure before committing many times

PR #497, opened 2022-02-24
fix: Return the correct timeout in _expires_at

Skelmis:
## Summary

This pull request fixes the `View._expires_at` decor so that it returns the correct timeout.

If you wish to check this yourself
```python
        from discord import ui
        import asyncio

        class MF(ui.View):
            @ui.button(label="reset or something")
            async def a(self, _, __):
                self.first = (self._expires_at, self._View__timeout_expiry)
                await asyncio.sleep(2)
                self.second = (self._expires_at, self._View__timeout_expiry)
                print(self.first, self.second, sep="\n")

        await ctx.send("a", view=MF())
```

They *should* be the same values both times, however, they are not. This PR resolves that.

Current output:
```
(34118.328, 34118.328)
(34120.343, 34118.328)
```

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/ui/view.py
+++ b/nextcord/ui/view.py
@@ -248,9 +248,7 @@ class View:
 
             None if no timeout is set.
         """
-        if self.timeout and self.__timeout_expiry is not None:
-            return self.__timeout_expiry
-        return None
+        return self.__timeout_expiry
 
     def add_item(self, item: Item) -> None:
         """Adds an item to the view.
--- a/nextcord/ui/view.py
+++ b/nextcord/ui/view.py
@@ -239,8 +239,17 @@ class View:
 
     @property
     def _expires_at(self) -> Optional[float]:
-        if self.timeout:
-            return time.monotonic() + self.timeout
+        """The monotonic time this view times out at.
+
+        Returns
+        -------
+        Optional[float]
+            When this view times out.
+
+            None if no timeout is set.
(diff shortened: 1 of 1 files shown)


TAG-Epic (approved):
Not tested

ooliver1 (approved):
just a code review, can't test as I don't have much idea what this change does

nextcord/ui/view.py
```diff
@@ -239,8 +239,17 @@ def from_message(cls, message: Message, /, *, timeout: Optional[float] = 180.0)
...
+        """The monotonic time this view times out at.
+
+        Returns
+        -------
+        Optional[float]
+            When this view times out.
+
+            None if no timeout is set.
+        """
+        if self.timeout and self.__timeout_expiry is not None:
```
TAG-Epic, line 251: Why are we checking for None on only one of the checks?
Is the `if self.timeout` check even necessary

TAG-Epic (changes requested):

nextcord/ui/view.py
DenverCoder1: Looks to me like this can simply return self.__timeout_expiry. If self.timeout is None, then self.__timeout_expiry will be None, and whether self.__timeout_expiry is None or not, returning it will give the correct output.

PR #500, opened 2022-02-25
Make SlashOption extend typing.Any

abrahammurciano:
Currently static type checkers will complain that a SlashOption object is not a valid default parameter for a parameter of type `int`, `str`, or anything. E.g.

```py
@nextcord.slash_command(description="bla bla bla")
async def foo(
	self,
	interaction: nextcord.Interaction,
	bar: str = SlashOption(description="bla bla bla"), # SlashOption is not a valid default for parameter of type str
):
	pass
```

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -37,7 +37,6 @@ from typing import (
     TYPE_CHECKING,
     Tuple,
     Union,
-    TYPE_CHECKING,
 )
 
 from .abc import GuildChannel
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -53,6 +53,9 @@ from .utils import MISSING
 
 if TYPE_CHECKING:
     from .state import ConnectionState
+    _SlashOptionMetaBase = Any
+else:
+    _SlashOptionMetaBase = object
 
 
 __all__ = (
@@ -98,7 +101,7 @@ class ClientCog:
 
 
 # Extends Any so that type checkers won't complain that it's a default for a parameter of a different type
-class SlashOption(Any if TYPE_CHECKING else object):
+class SlashOption(_SlashOptionMetaBase):
     """Provides Discord with information about an option in a command.
 
     When this class is set as the default argument of a parameter in an Application Command, additional information
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -96,7 +96,8 @@ class ClientCog:
         return self.__cog_to_register__
 
 
-class SlashOption:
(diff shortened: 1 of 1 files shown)


abrahammurciano:
@DenverCoder1 FYI

alentoghostflame (approved):
As long as you made sure it still functions, I say it's good :+1: 
Edit: Actually wait why are the checks failing?

DenverCoder1 (approved):

nextcord/application_command.py
```diff
@@ -96,7 +96,8 @@ def to_register(self) -> List[ApplicationCommand]:
         return self.__cog_to_register__
 
 
-class SlashOption:
+# Extends Any so that type checkers won't complain that it's a default for a parameter of a different type
+class SlashOption(Any):
```
alentoghostflame, line 100: This is breaking autodocs for some reason. We need to either find an alternative workaround for typing, or a way to have autodocs ignore this.

alentoghostflame (changes requested):

DenverCoder1 (changes requested):
Getting this error when importing the module (Python 3.10)

```
  File "test.py", line 8, in <module>
    import nextcord
  File "nextcord/__init__.py", line 24, in <module>
    from .client import *
  File "nextcord/client.py", line 54, in <module>
    from .application_command import message_command, slash_command, user_command
  File "nextcord/application_command.py", line 100, in <module>
    class SlashOption(Any):
  File "/usr/lib/python3.10/typing.py", line 377, in __mro_entries__
    raise TypeError(f"Cannot subclass {self!r}")
TypeError: Cannot subclass typing.Any
```

DenverCoder1:
Suggestion to fix:

```py
if TYPE_CHECKING:
    Base = Any
else:
    Base = object

...

class SlashOption(Base):
```

See: https://stackoverflow.com/a/68194494/11608064

nextcord/application_command.py
```diff
@@ -37,6 +37,7 @@
     TYPE_CHECKING,
     Tuple,
     Union,
+    TYPE_CHECKING,
```
DenverCoder1, line 40: TYPE_CHECKING was already imported, other than that, looks good

```suggestion
```

DenverCoder1 (changes requested):

MaskDuck (approved):
lgtm, hasnt test yet

DenverCoder1 (approved):
Looks good

PR #505, opened 2022-03-01
Add tag property to BaseUser

multiii:
Add tag property to BaseUser, displaying the user's entire tag (username + discriminator)

## Summary

It adds a new property to the BaseUser class, making it easier to print out the user's entire tag

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/user.py
+++ b/nextcord/user.py
@@ -256,6 +256,11 @@ class BaseUser(_UserTag):
         is returned instead.
         """
         return self.name
+    
+    @property
+    def tag(self) -> str:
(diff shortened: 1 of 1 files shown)


MaskDuck:
if i do remember correctly str(nextcord.User) should already return it.

multiii:
> if i do remember correctly str(nextcord.User) should already return it.

Oh I didn't actually know that. I guess this commit isn't needed then

MaskDuck:
aw these snippet of code do it
https://github.com/nextcord/nextcord/blob/master/nextcord/user.py#L96-L97

PR #515, opened 2022-03-07
Fix audio delay when calling after function

Dysta:
## Summary

Sometimes `source.cleanup()` may take several seconds to finish, which causes a noticeable delay when playing audio.

A possible solution is to call `self._call_after()` before `self.source.cleanup()` in `AudioPlayer.run()`.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/player.py
+++ b/nextcord/player.py
@@ -675,8 +675,8 @@ class AudioPlayer(threading.Thread):
             self._current_error = exc
             self.stop()
         finally:
-            self.source.cleanup()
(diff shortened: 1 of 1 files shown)


ooliver1 (approved):
not tested

Skelmis (approved):
Untested

DenverCoder1 (approved):
Looks good :+1: (Haven't tested)

ooliver1:
tested, that is so much better

PR #517, opened 2022-03-09
Cython compatibility

AstroFloof:
## Summary

This PR is supposed to fix some incompatibilities with Cython and Nextcord. 
The first of these is that Cython stores coros internally with the same `co_flags` as a regular function. Since `inspect.iscoroutinefunction` is used a bunch in nextcord to ensure listeners work right, "Pure Python" compiled with Cython cannot use nextcord. Cython *does* add a flag elsewhere that `asyncio.iscoroutinefunction` does check for (see [this](https://github.com/cython/cython/issues/2273#issuecomment-414032693)). I replaced all calls to inspect's impl with asyncio's impl, altering imports where necessary.

This is a draft for now, because I don't want to merge it until I can make a bot using Cython. Once the first obstacle is fixed, I'm sure there will be more.

## Checklist

- [x] This PR fixes an issue.

**I don't know how to test this yet, I was wondering if you guys would be able to point me in the right direction**

### DIFF
--- a/docs/extensions/attributetable.py
+++ b/docs/extensions/attributetable.py
@@ -3,6 +3,7 @@ from sphinx.locale import _
 from docutils import nodes
 from sphinx import addnodes
 
+import asyncio
 from collections import OrderedDict, namedtuple
 import importlib
 import inspect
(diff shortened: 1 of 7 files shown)


Skelmis (approved):
lgtm

ooliver1 (commented):
looks good so far, not approved as you said its untested, @ooliver#6969 in discord when tested though :)

Issue #468, opened 2022-02-15
Async setup


TAG-Epic:
For ext.commands allow setup to be async

MaskDuck:
actually not possible =((

ooliver1:
Nah it's possible, you just don't use asyncio.run for it

PR #536, opened 2022-03-17
feat: asynchronousifying setup function in cogs

MaskDuck:
## Summary

# NOTE: THIS WILL RAISE AN KEYBOARDINTERRUPT IF YOU ARE DOING AS SKELMIS SAYS AND IS ENDING THE BOT USING CONTROL + C

setup to be awaitable, fix issue #468 

## Checklist

- [x] If code changes were made then they have been tested. (pls help)
    - [x] I have updated the documentation to reflect the changes. (pls help)
- [x] This PR fixes an issue.
- [x] This PR is a breaking change (e.g. methods or parameters removed/renamed) (uh not really, but yeah)

### DIFF
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -422,7 +422,7 @@ Example: ::
 
     async def main():
         bot.load_extension('cog_name')
+        await bot.start("token")
     
     asyncio.run(main())
-    bot.run("token")
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -422,7 +422,7 @@ Example: ::
 
     async def main():
         bot.load_extension('cog_name')
-        await bot.start(token='your_token')
     
     asyncio.run(main())
+    bot.run("token")
 
--- a/docs/faq.rst
+++ b/docs/faq.rst
@@ -410,3 +410,19 @@ Example: ::
         await ctx.send(f'Pushing to {remote} {branch}')
 
 This could then be used as ``?git push origin master``.
+
+How do I make my setup function a coroutine and load it?
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+Create a new coroutine when you load the cog.
+
+Example: ::
+
+    import asyncio
+
+    async def main():
+        bot.load_extension('cog_name')
+        await bot.start(token='your_token')
+    
+    asyncio.run(main())
+
--- a/nextcord/ext/commands/bot.py
+++ b/nextcord/ext/commands/bot.py
@@ -695,7 +695,10 @@ class BotBase(GroupMixin):
         extras = extras or {}
         try:
             if asyncio.iscoroutinefunction(setup):
-                asyncio.create_task(setup(self, **extras))
+                try:
+                    asyncio.create_task(setup(self, **extras))
+                except RuntimeError:
+                    raise errors.ExtensionFailed(key, RuntimeError(f"Seems like this extension is having an async setup function. Please read our FAQ here: https://nextcord.readthedocs.io/faq.html#how-do-i-make-my-setup-function-a-coroutine-and-load-it"))
             else:
                 setup(self, **extras)
         except Exception as e:
--- a/nextcord/ext/commands/bot.py
+++ b/nextcord/ext/commands/bot.py
@@ -698,7 +698,7 @@ class BotBase(GroupMixin):
                 try:
                     asyncio.create_task(setup(self, **extras))
                 except RuntimeError:
-                    raise errors.ExtensionFailed(key, RuntimeError(f"Seems like this extension is having an async setup function. Please read our FAQ here: https://nextcord.readthedocs.io/faq.html#how-do-i-make-my-setup-function-a-coroutine-and-load-it"))
+                    raise RuntimeError(f"Seems like this extension is having an async setup function. Please read our FAQ here: https://nextcord.readthedocs.io/faq.html#how-do-i-make-my-setup-function-a-coroutine-and-load-it")
             else:
                 setup(self, **extras)
         except Exception as e:
--- a/nextcord/ext/commands/bot.py
+++ b/nextcord/ext/commands/bot.py
@@ -694,7 +694,10 @@ class BotBase(GroupMixin):
 
         extras = extras or {}
         try:
-            setup(self, **extras)
+            if asyncio.iscoroutinefunction(setup):
+                asyncio.create_task(setup(self, **extras))
+            else:
+                setup(self, **extras)
         except Exception as e:
             del sys.modules[key]
             self._remove_module_references(lib.__name__)
--- a/nextcord/ext/commands/bot.py
+++ b/nextcord/ext/commands/bot.py
@@ -698,7 +698,11 @@ class BotBase(GroupMixin):
                 try:
                     asyncio.create_task(setup(self, **extras))
                 except RuntimeError:
-                    raise RuntimeError(f"Seems like this extension is having an async setup function. Please read our FAQ here: https://nextcord.readthedocs.io/faq.html#how-do-i-make-my-setup-function-a-coroutine-and-load-it")
+                    raise RuntimeError(f"""
+                    Looks like your attempting to load an asynchronous setup function incorrectly.
+                    Please read our FAQ here:
+                    https://nextcord.readthedocs.io/faq.html#how-do-i-make-my-setup-function-a-coroutine-and-load-it
+                    """)
             else:
                 setup(self, **extras)
         except Exception as e:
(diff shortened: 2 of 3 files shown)


MaskDuck:
created in github.com, dont expect for lint to pass

MaskDuck:
ok so the only solution is make a breaking change with `Bot.load_extension`. 

it is great if we could make a branch for this.

nextcord/ext/commands/bot.py
```diff
@@ -694,7 +694,10 @@ def _load_from_module_spec(
 
         extras = extras or {}
         try:
-            setup(self, **extras)
+            if asyncio.iscoroutinefunction(setup):
+                asyncio.create_task(setup(self, **extras))
```
Skelmis, line 698: This will error if no loop exists, a case which is fairly familiar on bot startup. Please catch this exception and provide some form of helpful information, such as an FAQ entry or something along the lines of call `load_ext..` once your bot is running or has an event loop.
A FAQ entry is likely better as you can show example ways to load things correctly.
```py
async def main():
    bot.load_ext...
    await bot.start()

asyncio.run(main())
```

Skelmis (changes requested):
Please make your pull request name better, ideally following [cc](https://kapeli.com/cheat_sheets/Conventional_Commits.docset/Contents/Resources/Documents/index).
Something like `feat: Allow cog setup to be async`

Skelmis:
I personally disagree with making this a breaking change, especially given the overall use-case from what I have seen. Yes, some form of async loading is nice for a minority, however, enforcing the syntax proposed essentially removes ``.run`` which is what the majority of bots uses.

nextcord/ext/commands/bot.py
MaskDuck: seems like a better solution than mine...

MaskDuck:
your suggestion seems like a great idea!

nextcord/ext/commands/bot.py
MaskDuck: image
@Skelmis looks like ooliver disagreed with catching error

ooliver1: Or use asyncio.get_event_loop which I believe creates a new loop if it doesn't exist, and you can catch the error and create a new loop yourself if that works too

Astrea49: Psst: *don't* do this. `asyncio` as a whole discourages against making loops yourself unless you're *really* low level. As of Python 3.10, `asyncio.get_event_loop` is deprecated and will no longer make the event loop from 3.11+, making it an alias to `get_running_loop`.

Not catching this exception makes sense since there's no need, but there should also be a FAQ entry about it too.

Skelmis:
> You could use [nextcord.utils.maybe_coroutine()](https://github.com/nextcord/nextcord/blob/fa961ca79d852b5df29c42985b0993884a3cfcdf/nextcord/utils.py#L511-L516) for this.

This requires making `load_ext..` async, which I have already commented on.

nextcord/ext/commands/bot.py
Skelmis: We shouldn't be creating a loop every time we load a cog..? The event loop for the program relates to when the bot is ran

MaskDuck: @ooliver1 @Astrea49 @Skelmis how do you think about using `asyncio.new_event_loop()` or make a new helper method names `nextcord.utils.get_event_loop`?

Astrea49: > @ooliver1 @Astrea49 @Skelmis how do you think about using `asyncio.new_event_loop()` or make a new helper method names `nextcord.utils.get_event_loop`?

Don't do either. Let loading fail if there's no loop - that's honestly ideal. Add an FAQ entry about it instead.

Skelmis: My personal opinion is no. We shouldn't 'just make' a new event loop because it's a cheap fix to a problem rather then actually documenting in say, an faq, how to use your new feature.

I can't find things on mobile in the source itself, but we'd be creating atleast in my case up to 10 event loops which is absurd. 
The bot runs on a single event loop as is standard and we should be using that loop.

Skelmis: I agree with the comment above mine.

ooliver1 (changes requested):
make an faq entry on the error that may be risen as discussed

MaskDuck:
@ooliver1 i made an faq entry.

ooliver1 (approved):

DenverCoder1:
In extensions.rst this should be changed (remove "not a coroutine")

> An extension at its core is a python file with an entry point called ``setup``. This setup must be a plain Python function **(not a coroutine)**. It takes a single parameter -- the :class:`~.commands.Bot` that loads the extension.

ooliver1:
also perhaps handle error then raise an error which shows the faq entry perhaps

MaskDuck:
@ooliver1 handled

MaskDuck:
guys i have tested the code
# Files
### main.py
```python
import nextcord
import random

import asyncio

description = '''An example bot to showcase the nextcord.ext.commands extension
module.
There are a number of utility commands being showcased here.'''

intents = nextcord.Intents.default()
intents.members = True

bot = commands.Bot(command_prefix=')', description=description, intents=intents)

@bot.event
async def on_ready():
    print(f'Logged in as {bot.user} (ID: {bot.user.id})')

async def setup():
    bot.load_extension("testcog")

asyncio.run(setup())

bot.run('token')
```

### testcog.py
```python
from nextcord.ext.commands import Cog

class testCog(Cog):
    def __init__(self, bot):
        self.bot = bot
    
    @commands.command()
    async def test(self, ctx):
        await ctx.send("Test")

async def setup(bot):
    bot.add_cog(testCog(bot))
    print("test done")
```

Skelmis:
> guys i have tested the code
> 
> . . . . .

Have you also tested backwards compatibility?

MaskDuck:
wdym?

nextcord/ext/commands/bot.py
```diff
@@ -694,7 +694,13 @@ def _load_from_module_spec(
 
         extras = extras or {}
         try:
-            setup(self, **extras)
+            if asyncio.iscoroutinefunction(setup):
+                try:
+                    asyncio.create_task(setup(self, **extras))
+                except RuntimeError:
+                    raise RuntimeError(f"Seems like this extension is having an async setup function. Please read our FAQ here: https://nextcord.readthedocs.io/faq.html#how-do-i-make-my-setup-function-a-coroutine-and-load-it")
```
Skelmis, line 701: ``Seems like this extension has an async setup function. Please read our FAQ here: https://nextcord.readthedocs.io/faq.html#how-do-i-make-my-setup-function-a-coroutine-and-load-it``

Alternately:
``Looks like your attempting to load an asynchronous setup function incorrectly. Please read our FAQ here: https://nextcord.readthedocs.io/faq.html#how-do-i-make-my-setup-function-a-coroutine-and-load-it``


Further, please wrap this to be multiple lines, not just one big line

Skelmis (changes requested):

MaskDuck:
(if you mean plain python func cog then i have also tested)

Skelmis:
> (if you mean plain python func setup then i have also tested)

Sounds good

docs/faq.rst
```diff
@@ -410,3 +410,19 @@ Example: ::
...
+
+Example: ::
+
+    import asyncio
+
+    async def main():
+        bot.load_extension('cog_name')
+    
+    asyncio.run(main())
+    bot.run("token")
```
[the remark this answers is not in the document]
MaskDuck: pretty sure i tried `await bot.start` at first, but await bot.start gives error.

nextcord/ext/commands/bot.py
MaskDuck: done.

docs/faq.rst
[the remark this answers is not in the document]
MaskDuck: @Skelmis ^

MaskDuck:
@Skelmis this is the error when you trying to load async things without async loaders
image

Skelmis:
TIL `asyncio.run` is indeed not blocking. So ty for that

However, still not removing the requested changes however as it is bad practice since it's creating two loops (I think) and either of the two calls should be *the* entry point for the bot. 

```python
async def main():
    bot.load_extension("cog_name")
    await bot.start("TOKEN")


asyncio.run(main())
```

Tested on your fork works fine and as expected

MaskDuck:
kay let me test again

it give an error at my end

MaskDuck:
@Skelmis still have errors at my ends
image

Skelmis:
Code?

Alternately you can create a task instead, and then use `bot.run`

Skelmis (approved):
LGTM and I've tested on:
- 3.8.5
- 3.9.6
- 3.10.0

ooliver1 (approved):

PR #551, opened 2022-03-23
Fix issue removing embeds when editing with files and add `embeds` to `PartialMessage.edit`

DenverCoder1:
## Summary

Feature:
* Adds support for sending `embeds` in `PartialMessage.edit`

Fixes:
* Replaces use of deprecated `embed` field in `PartialMessage.edit` with `embeds`
* **Fixes issue when trying to remove content, embeds, or other attributes when editing a message with files**

## Bug fix details

This PR fixes a bug where if you pass `embed=None` or `embeds=[]` to an edit method while also sending files, the embeds do not get removed.

This bug also applies to sending `view=None`, a view with no components, an empty list of stickers, or empty string as content.

Example:

```py
@bot.command(name="pme")
async def partial_message_embeds(ctx: commands.Context):
    pm = ctx.channel.get_partial_message(956263847562477619)
    await pm.edit(content="Test 1", embeds=[nextcord.Embed(title="Test"), nextcord.Embed(title="Test 2")])
    await asyncio.sleep(2)
    # this should remove the embeds
    await pm.edit(content="Test 2", embed=None, files=[nextcord.File("nextcord-logo.png")])
    # this should also remove the embeds
    await pm.edit(content="Test 2", embeds=[], files=[nextcord.File("nextcord-logo.png")])
```

This was fixed by checking for `embeds` being `None` rather than checking `if embeds` (which would skip adding the embeds field if the list is empty.

## Testing

Code for testing changes:

https://paste.nextcord.dev/?id=1648148708552851

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/message.py
+++ b/nextcord/message.py
@@ -1862,10 +1862,10 @@ class PartialMessage(Hashable):
 
         if 'embed' in fields:
             embed = fields.pop('embed')
-            fields['embeds'] = [embed] if embed is not None else []
+            fields['embeds'] = [embed.to_dict()] if embed is not None else []
 
         elif 'embeds' in fields:
-            fields['embeds'] = [e.to_dict() for e in fields['embeds']]
+            fields['embeds'] = [embed.to_dict() for embed in fields['embeds']]
 
         try:
             suppress: bool = fields.pop('suppress')
--- a/nextcord/message.py
+++ b/nextcord/message.py
@@ -1804,6 +1804,11 @@ class PartialMessage(Hashable):
         embed: Optional[:class:`Embed`]
             The new embed to replace the original with.
             Could be ``None`` to remove the embed.
+        embeds: List[:class:`Embed`]
+            The new embeds to replace the original with. Must be a maximum of 10.
+            To remove all embeds ``[]`` should be passed.
+
+            .. versionadded:: 2.0
         suppress: :class:`bool`
             Whether to suppress embeds for the message. This removes
             all the embeds if set to ``True``. If set to ``False``
@@ -1835,6 +1840,8 @@ class PartialMessage(Hashable):
         Forbidden
             Tried to suppress a message without permissions or
             edited a message's content or embed that isn't yours.
+        ~nextcord.InvalidArgument
+            You specified both ``embed`` and ``embeds``
 
         Returns
         ---------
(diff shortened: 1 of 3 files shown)


nextcord/message.py
```diff
@@ -1850,13 +1857,14 @@ async def edit(self, **fields: Any) -> Optional[Message]:
...
-        except KeyError:
-            pass
-        else:
-            if embed is not None:
-                fields['embed'] = embed.to_dict()
+        if 'embed' in fields and 'embeds' in fields:
+            raise InvalidArgument('Cannot pass both embed and embeds parameter to edit()')
+
+        if 'embed' in fields and fields['embed'] is not None:
+            fields['embed'] = fields['embed'].to_dict()
```
MaskDuck, line 1864:
```suggestion
            fields['embed'] = [fields['embed'].to_dict()]
```
api v10 thing

MaskDuck (changes requested):

nextcord/message.py
DenverCoder1: The send_multipart_helper converts embed to embeds already so it doesn't need to be handled here

MaskDuck: k

MaskDuck (approved):

nextcord/message.py
DenverCoder1: Actually, it seems like it only gets converted if files are sent. I'll look into this further  :+1:

nextcord/message.py
```diff
@@ -1850,13 +1857,15 @@ async def edit(self, **fields: Any) -> Optional[Message]:
...
-            pass
-        else:
-            if embed is not None:
-                fields['embed'] = embed.to_dict()
+        if 'embed' in fields and 'embeds' in fields:
+            raise InvalidArgument('Cannot pass both embed and embeds parameter to edit()')
+
+        if 'embed' in fields:
+            embed = fields.pop('embed')
+            fields['embeds'] = [embed] if embed is not None else []
```
Skelmis, line 1865: Do you not required `.to_dict()` on this setup? It appears `embeds` and the previous implementation called it.

Skelmis (changes requested):
Looks like a regression from the change, however, if tested feel free to ignore it.

nextcord/message.py
DenverCoder1: Yeah, I think I messed that up after a few rewrites. It probably does need to be added back.

DenverCoder1:
@Skelmis Issue is fixed, all cases have been tested again

Skelmis (approved):

ooliver1 (approved):

Issue #349, opened 2022-01-11
"on_close" event for when a Client object is terminating its event loop


pmdevita:
### Summary

A new event that is called right before the Client is terminated.

### What is the feature request for?

The core library

### The Problem

I'm working with a database library ([tortoise-orm](https://github.com/tortoise/tortoise-orm/)) and in order to properly shutdown, Tortoise must be notified within the async loop so that it can terminate its database connections. Currently, when Nextcord is signaled to end, it stops everything and waits for the current async tasks to finish or kills them. Nextcord cannot kill the database connection tasks though (nor would that necessarily be the best way to end them) and this causes the process to hang and never exit.

This problem may also crop up with other database libraries and applications that need notice to gracefully shutdown certain tasks within the async loop.

### The Ideal Solution

[This](https://github.com/nextcord/nextcord/blob/e4197a8942c56308a73623dbbd9c47bc9a280183/nextcord/client.py#L719) is the core of Nextcord's current async loop manager.

```python
        try:
            loop.run_forever()
        except KeyboardInterrupt:
            _log.info('Received signal to terminate bot and event loop.')
        finally:
            future.remove_done_callback(stop_loop_on_completion)
            _log.info('Cleaning up tasks.')
            _cleanup_loop(loop)
```

I would suggest we add an "on_close" event that the Client and cogs can register for. We then call this event with `loop.run_until_complete(self.on_close())`. So it would look something like this

```python
        try:
            loop.run_forever()
        except KeyboardInterrupt:
            _log.info('Received signal to terminate bot and event loop.')
        finally:
            loop.run_until_complete(self.on_close())
            future.remove_done_callback(stop_loop_on_completion)
            _log.info('Cleaning up tasks.')
            _cleanup_loop(loop)
```

### The Current Solution

Currently, I've added my own run function to my Discord bot. It's more crude than Nextcord's but it works and calls the Bot's `on_close` function before terminating.

```python
class DiscordBot(commands.Bot):
    def run_with_close(self, *args):
        try:
            self.loop.run_until_complete(self.start(*args))
        except KeyboardInterrupt:
            print("Exiting...")
        finally:
            self.loop.run_until_complete(self.on_close())
            self.loop.close()
```

### Additional Context

My understanding of how asyncio works isn't super thorough so `loop.run_until_complete()` might not be the proper way to run the closing event.

pmdevita:
Anything I can do to help out with this?

MaskDuck:
[...]

PR #553, opened 2022-03-23
feat: Add close event (#349)

pmdevita:
## Summary

This adds the "close" event, which can then be listened for by library users to help shutdown parts of their bot, like database connections.

This ended up being a single line addition to the Client.close() function to just dispatch the `close` event but it looks like that's all that should be needed? Would greatly appreciate some feedback or confirmation.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -307,6 +307,10 @@ to handle it, which defaults to print a traceback and ignoring the exception.
     :param kwargs: The keyword arguments for the event that raised the
         exception.
 
+.. function:: on_close()
+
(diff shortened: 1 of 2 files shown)


ShowierData9978-Alt (approved):
Love This 

Somth That they shoulda done a while ago tbh

MaskDuck (approved):
acceptable

ooliver1 (approved):
tested (with KeyboardInterupt to gracefully close - as this will only fire on graceful close)

PR #557, opened 2022-03-28
docs: Remove incorrect attributes from docs

DenverCoder1:
## Summary

* Fix docs for `client` and `bot` in `nextcord.Interaction`

## Changes

client is documented as a property so should not be in attributes

bot never existed as an attribute so should not be in the documentation

_client does not exist as an attribute so should not be in slots

## Checklist

- [ ] This PR fixes an issue.


Skelmis (approved):

MaskDuck (approved):
nice

PR #560, opened 2022-03-29
Remove Store channel functionality

H4ckerxx44:
## Summary
Discord [discontinued](https://discord.com/developers/docs/game-and-server-management/special-channels#store-channels) selling games with Store channels a while ago so I figured we remove it.

Might be considered a `breaking change`.

## Checklist

- [ ] This PR fixes an issue.


ooliver1 (commented):
as this is a breaking change (since 1.7), may i suggest editing migrating_2.rst to mention its removal even though its unusable

ooliver1 (approved):

H4ckerxx44:
> Would it also be useful to remove [the enum for store channel](https://github.com/nextcord/nextcord/blob/6ac4187ee86310b5a2ea9d467a4908ed0c95d679/nextcord/enums.py#L188) and similar?

[That would definitely make sense.](https://nextcord.readthedocs.io/en/stable/api.html?highlight=channel%20type#nextcord.ChannelType.store)

ooliver1 (changes requested):
may be annoying but migrating_2 could do with those appinfo attr changes and the enum change

ooliver1 (approved):

PR #567, opened 2022-04-01
Add global scope for slash commands

Dysta:
## Summary

This PR add the ability for users to define a global scope for their slash commands instead of filling in the `guilds_id` parameter in each slash command decorator.

### Behavior: 
If user fill the `guild_ids` parameter for a slash command, it overwrite the `rollout_guild_ids` and the command is only available in the selected guild.

### Example:
```py

bot = commands.Bot(
    command_prefix="test ",
    rollout_guild_ids=[957587943717994586],
)


@bot.slash_command(name="test")
async def slashe(
    interaction,
    number: int,
    letter: str,
):
    await interaction.response.send_message(f"Helo {number} {letter}")


@bot.slash_command(name="testistos", guild_ids=[957598850691006544])
async def other_slashe(
    interaction,
    number: int,
):
    await interaction.response.send_message(f"Henllooooo {number}")

bot.run("token")
```
#### Expected behavior:
`957587943717994586` have the two command registered
image

`957598850691006544` have only the second command registered
image

## Checklist

- [ ] This PR fixes an issue.


nextcord/client.py
```diff
@@ -248,6 +248,10 @@ class Client:
         Warning: While enabling this will prevent "ghost" commands on guilds with removed code references, rolling out
         to ALL guilds with anything other than a very small bot will likely cause it to get rate limited.
 
+    rollout_guild_ids: `Optional[:class:`list`[:class:`int`]]`
```
ooliver1, line 251:
```suggestion
    rollout_guild_ids: Optional[List[:class:`int`]]
```
for consistency and also as the docs render like this

ooliver1 (changes requested):

ooliver1:
working in testing however

ooliver1 (approved):
more of a testing approval than a code approval

nextcord/client.py
```diff
@@ -1981,7 +1992,7 @@ def user_command(
         description: :class:`str`
             Description of the command that users will see. If not set, it defaults to the bare minimum Discord allows.
         guild_ids: Iterable[:class:`int`]
-            IDs of :class:`Guild`'s to add this command to. If unset, this will be a global command.
+            IDs of :class:`Guild`'s to add this command to. If unset, it will use the client rollout guild ids.
```
DenverCoder1, line 1995: This could be made more clear. It currently seems that setting client guild ids will add Guilds **in addition** to the set guild ids, not instead of.

I think it would make sense to use the command's `guild_ids` **instead of** the client `rollout_guild_ids` if they are set, but the way I'm reading the code, it seems like it is not that way right now.

Also, these descriptions should probably still mention that you can make it global by overriding guild_ids with the default or not setting either.

nextcord/client.py
```diff
@@ -1945,13 +1950,19 @@ async def rollout_application_commands(self) -> None:
 
     def _add_decorated_application_commands(self) -> None:
         for command in self._application_commands_to_add:
+            # we inject the rollout guild ids in the decorated commands
+            for i in self._rollout_guild_ids:
+                command.add_guild_rollout(i)
```
DenverCoder1, line 1955:
```suggestion
            # we inject the rollout guild ids in the decorated commands
            if not command.force_global and not command.guild_ids_to_rollout:
                for i in self._rollout_guild_ids:
                    command.add_guild_rollout(i)
```

nextcord/client.py
DenverCoder1, line 1995:
```suggestion
            IDs of :class:`Guild`'s to add this command to. If unset, ``rollout_guild_ids`` set on the
            :class:`~.ext.commands.Bot` or :class:`Client` will be used. If neither are present, this will be a global command.
```

nextcord/client.py
```diff
@@ -2013,7 +2024,7 @@ def message_command(
         description: :class:`str`
             Description of the command that users will see. If not set, it defaults to the bare minimum Discord allows.
         guild_ids: Iterable[:class:`int`]
-            IDs of :class:`Guild`'s to add this command to. If unset, this will be a global command.
+            IDs of :class:`Guild`'s to add this command to. If unset, it will use the client rollout guild ids.
```
DenverCoder1, line 2027:
```suggestion
            IDs of :class:`Guild`'s to add this command to. If unset, ``rollout_guild_ids`` set on the
            :class:`~.ext.commands.Bot` or :class:`Client` will be used. If neither are present, this will be a global command.
```

nextcord/client.py
```diff
@@ -2045,7 +2056,7 @@ def slash_command(
...
```
DenverCoder1, line 2059:
```suggestion
            IDs of :class:`Guild`'s to add this command to. If unset, ``rollout_guild_ids`` set on the
            :class:`~.ext.commands.Bot` or :class:`Client` will be used. If neither are present, this will be a global command.
```

nextcord/client.py
```diff
@@ -248,6 +248,10 @@ class Client:
         Warning: While enabling this will prevent "ghost" commands on guilds with removed code references, rolling out
         to ALL guilds with anything other than a very small bot will likely cause it to get rate limited.
 
+    rollout_guild_ids: Optional[List[:class:`int`]]
```
DenverCoder1, line 251:
```suggestion
    rollout_guild_ids: List[:class:`int`]
```

DenverCoder1 (changes requested):
This is my suggestion to make the default behavior for command settings to *override* the global settings rather than adding on

Also a few documentation suggestions

Skelmis:
> This is my suggestion to make the default behavior for command settings to *override* the global settings rather than adding on
> 
> Also a few documentation suggestions

I agree with overriding here.

Skelmis:
> This is my suggestion to make the default behavior for command settings to *override* the global settings rather than adding on
> 
> Also a few documentation suggestions

I agree with overriding here.

Dysta:
Ok after investigate, I fix with @DenverCoder1 suggestion and it should do what we expect.

Test code :
```py
from secret import BOT_TOKEN

import logging

logging.basicConfig(level=logging.DEBUG)

NC_SB_1 = 1234
NC_SB_2 = 5678

bot = commands.Bot(
    command_prefix="test ",
    rollout_guild_ids=[NC_SB_1, NC_SB_2],
)


@bot.slash_command(name="all_serv")
async def slashe(interaction, number: int, letter: str):
    await interaction.response.send_message(f"Helo {number} {letter}")

# should be only in NC_SB_2
@bot.slash_command(name="only_sb_2", guild_ids=[NC_SB_2])
async def other_slashe(interaction, number: int):
    await interaction.response.send_message(f"Henllooooo {number}")

bot.run(BOT_TOKEN)
```

ooliver1 (approved):
approval for new changes

nextcord/client.py
```diff
@@ -248,6 +248,10 @@ class Client:
         Warning: While enabling this will prevent "ghost" commands on guilds with removed code references, rolling out
         to ALL guilds with anything other than a very small bot will likely cause it to get rate limited.
 
+    rollout_guild_ids: List[:class:`int`]
```
TAG-Epic, line 251: I dislike the name here.

nextcord/client.py
```diff
@@ -248,6 +248,10 @@ class Client:
...
+        Allow user to set a global scope of guild ids instead of filling in `guild_ids` parameter for each application
+        command decorator. Defaults to ``[]``
```
TAG-Epic, line 253: This should be None?

nextcord/client.py
```diff
@@ -297,6 +301,7 @@ def __init__(
         self._rollout_register_new: bool = options.pop("rollout_register_new", True)
         self._rollout_update_known: bool = options.pop("rollout_update_known", True)
         self._rollout_all_guilds: bool = options.pop("rollout_all_guilds", False)
+        self._rollout_guild_ids: List[int] = options.pop("rollout_guild_ids", list())
```
TAG-Epic, line 304: If this is left as a list
```suggestion
        self._rollout_guild_ids: List[int] = options.pop("rollout_guild_ids", [])
```

TAG-Epic (changes requested):

nextcord/client.py
ooliver1: as they work as defaults, `default_guild_ids` may work better

also dont forget the docstrings if this is changed

nextcord/client.py
```diff
@@ -2045,7 +2060,8 @@ def slash_command(
         description: :class:`str`
             Description of the command that users will see. If not set, it defaults to the bare minimum Discord allows.
         guild_ids: Iterable[:class:`int`]
-            IDs of :class:`Guild`'s to add this command to. If unset, this will be a global command.
+            IDs of :class:`Guild`'s to add this command to. If unset, ``rollout_guild_ids`` set on the
```
ooliver1, line 2063:
```suggestion
            IDs of :class:`Guild`'s to add this command to. If unset, ``default_guild_ids`` set on the
```

nextcord/client.py
```diff
@@ -2013,7 +2027,8 @@ def message_command(
...
```
ooliver1, line 2030:
```suggestion
            IDs of :class:`Guild`'s to add this command to. If unset, ``default_guild_ids`` set on the
```

nextcord/client.py
```diff
@@ -1981,7 +1994,8 @@ def user_command(
...
```
ooliver1, line 1997:
```suggestion
            IDs of :class:`Guild`'s to add this command to. If unset, ``default_guild_ids`` set on the
```

ooliver1 (changes requested):

nextcord/client.py
DenverCoder1: To be consistent with application commands, defaulting to empty list is fine.

It was possibly done in a bit of an non-intuitive manner, but currently setting `guild_ids=[]` on a command is the same as not setting it.

nextcord/client.py
DenverCoder1: `rollout_guild_ids` is consistent with the naming of the other slash kwargs, but I'm fine with `default_guild_ids` as well.

DenverCoder1:
Something might need to be changed in the slash commands code to make it so that setting guild_ids to `[]` will make it global (and *not* use the client default).

It may need to differentiate between guild_ids being not being set and guild_ids being set to None/empty list.

Maybe we could have a flag to check if guild_ids was `MISSING` or not in the `ApplicationCommand` constructor?

Skelmis:
Any updates on this?

Skelmis:
Closing as this appears to be stalled.

PR #568, opened 2022-04-02
Add check_all function to application_checks

coderFlameyosFlow:
## Summary

This PR adds "check_all" function to nextcord.ext.application_checks
to make code slightly "nicer"?
How it works, is you can use check_all like:
```py
def guild_owner_only():
  def predicate(interaction: Interaction):
     return interaction.user == interaction.guild.owner
  return application_checks.check(predicate)

@client.slash_command()
@application_checks.check_all(
  guild_owner_only(),
  application_checks.is_owner()
)
async def owner_only(interaction: Interaction):
  await interaction.send("I know you, mister!")
```
You make a check (or not, just 2 or more checks together) so you can put it in check_all!
or you can do something like this as I said:
```py
@client.slash_command()
@application_checks.check_all(
  application_checks.bot_has_permissions(manage_messages=True),
  application_checks.has_permissions(manage_messages=True)
)
async def with_perms_only(interaction: Interaction):
  await interaction.send("We're both worthy of managing messages!")
```
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/ext/application_checks/core.py
+++ b/nextcord/ext/application_checks/core.py
@@ -303,8 +303,8 @@ def check_all(*checks: "ApplicationCheck") -> Callable[[T], T]:
         for func in unwrapped:
             try:
                 value = await func(interaction)
-            except:
-                raise
+            except ApplicationCheckFailure:
+                raise ApplicationCheckAllFailure from ApplicationCheckFailure
             if value:
                 return True
 
--- a/nextcord/ext/application_checks/core.py
+++ b/nextcord/ext/application_checks/core.py
@@ -296,22 +296,17 @@ def check_all(*checks: "ApplicationCheck") -> Callable[[T], T]:
             raise TypeError(
                 f"{wrapped!r} must be wrapped by application_checks.check decorator"
             ) from None
-        else:
-            unwrapped.append(pred)
+
+        unwrapped.append(pred)
 
     async def predicate(interaction: Interaction) -> bool:
         for func in unwrapped:
             try:
                 value = await func(interaction)
-            except ApplicationCheckFailure as e:
-                raise ApplicationCheckAllFailure(
-                    f"The checks for every check in check_all failed"
-                ) from None
-            else:
-                if value:
-                    return True
-        # if we're here, all checks failed
-        raise ApplicationCheckAllFailure(unwrapped, errors)
+            except:
+                raise
+            if value:
+                return True
 
     return check(predicate)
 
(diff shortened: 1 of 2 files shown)


ooliver1:
This looks like a weirdly indebted way of just using multiple decorators

nextcord/ext/application_checks/errors.py
```diff
@@ -73,7 +73,17 @@ def __init__(
...
+class ApplicationCheckAllFailure(ApplicationCheckFailure):
+    """Exception raised when at least 1 predicate in :func:`check_all` fails.
+
+    This inherits from :exc:`~.ApplicationCheckFailure`.
+    """
 
+    def __init__(
+        self,
+    ) -> None:
+        super().__init__("You do not have permission to run this command.")
```
Skelmis, line 86: This should ideally be similar to check any in terms of information. This is not enough to work with ideally.

nextcord/ext/application_checks/core.py
```diff
@@ -242,6 +243,72 @@ async def predicate(interaction: Interaction) -> bool:
...
+        unwrapped.append(pred)
+
+    async def predicate(interaction: Interaction) -> bool:
+        for func in unwrapped:
+            try:
+                value = await func(interaction)
+            except ApplicationCheckFailure:
+                raise ApplicationCheckAllFailure from ApplicationCheckFailure
+            if value:
+                return True
```
Skelmis, line 309: You've just copied ``check_any``...? This acts the same way, it *doesnt* check them all. It passes on the first successful check if I am reading this right.

nextcord/ext/application_checks/core.py
```diff
@@ -242,6 +243,72 @@ async def predicate(interaction: Interaction) -> bool:
...
+            ) from None
+
+        unwrapped.append(pred)
+
+    async def predicate(interaction: Interaction) -> bool:
+        for func in unwrapped:
+            try:
+                value = await func(interaction)
+            except ApplicationCheckFailure:
+                raise ApplicationCheckAllFailure from ApplicationCheckFailure
```
Skelmis, line 307: I haven't used this style specifically, but wouldn't you want to raise this from the error itself instead of the class? Please correct me if they do the same thing.

```python
 except ApplicationCheckFailure as e:
     raise ApplicationCheckAllFailure from e
```

Skelmis (changes requested):
Review for reviews sake, however I do agree that its likely unrequired as you can just use both decorators

ooliver1:
Gonna close this as you can use multiple decos, this is a copy of check_any, the error seems to only work how has_permission works etc

PR #586, opened 2022-04-10
Update application_commands/cog_example to ClientCog

ToasterUwU:
## Summary

This updates the example for application command Cogs to the actual ClientCog. 
It also shows basic usage of the cog_check equivalent for ClientCog.

## Checklist

- [ ] This PR fixes an issue.
- [X] This PR is **not** a code change (e.g. documentation, README, ...)

### DIFF
--- a/examples/application_commands/cog_example.py
+++ b/examples/application_commands/cog_example.py
@@ -1,12 +1,15 @@
 import nextcord
-from nextcord.ext import commands
 
 TESTING_GUILD_ID = 123456789  # Replace with your testing guild id
 
 
-class ApplicationCommandCog(commands.Cog):
-    def __init__(self, bot: commands.Bot):
-        self.bot = bot
+class ApplicationCommandCog(nextcord.ClientCog):
+    # Add a check for every application command in this ClientCog. Execute command if True, dont execute command if False.
+    def cog_application_command_check(self, interaction: nextcord.Interaction):
+        if interaction.user:
+            return interaction.user.id == 123456789 # Example User ID, replace with your own for testing
+        else:
+            return False
 
     @nextcord.slash_command(guild_ids=[TESTING_GUILD_ID], description="Test command")
     async def my_slash_command(self, interaction: nextcord.Interaction):
@@ -19,7 +22,3 @@ class ApplicationCommandCog(commands.Cog):
     @nextcord.message_command(guild_ids=[TESTING_GUILD_ID])
     async def my_message_command(self, interaction: nextcord.Interaction, message: nextcord.Message):
         await interaction.response.send_message(f"{message}")
-
-
-def setup(bot):
-    bot.add_cog(ApplicationCommandCog(bot))


MaskDuck:
what is the point of this?

koala9712:
It may deserve a separate file for itself, `cog_checks` or something similar.
the setup function was removed, huh?

Has this been tested?

ToasterUwU:
> what is the point of this?

Keeping examples updated helps people to understand better. Examples are of no use when outdated.

ToasterUwU:
> It may deserve a separate file for itself, `cog_checks` or something similar.
> the setup function was removed, huh?
> 
> Has this been tested?

1. I don't think so. The old version was plain outdated, using the wrong class and everything and I don't see a reason why a feature of the ClientCog should be separate from the rest. 

2. Yes. I checked the box that I did. Took the thing, changed the example IDs to actual IDs and added the cog and run the Client. Worked like intended.

MaskDuck:
> > what is the point of this?
> 
> Keeping examples updated helps people to understand better. Examples are of no use when outdated.

i havent heard of that breaking change. the old one still works and imo it is better to use `commands.Cog` instead of `ClientCog`. if there is any breaking change the author of the pr which made breaking change should also replace the old example with the new example as well.

ToasterUwU:
> > > what is the point of this?
> > 
> > Keeping examples updated helps people to understand better. Examples are of no use when outdated.
> 
> i havent heard of that breaking change. the old one still works and imo it is better to use `commands.Cog` instead of `ClientCog`. if there is any breaking change the author of the pr which made breaking change should also replace the old example with the new example as well.
> 

There was no breaking change. But normal Cogs are just simply not intended to be used for application commands, the ClientCog is.

The ClientCog was made for the application commands, the ext.commands.Cog was made for the old system.
Using a class made for something is always better when using something which is not intended for that (anymore).

Besides that: There are features that ClientCog brings like the cog-wide check, which are not possible with the old Cog. cog_check only works for the commands.command(s).

Positive: Using proper class for application commands (the topic of that example), having the check, having the pre and after invoke functions.
Negative: Nothing

DenverCoder1:
commands.Cog inherits from ClientCog, so anything that can be done with ClientCog should also be possible with commands.Cog.

ClientCog is just a stripped down version of Cogs that you can use if you really don't want to use the commands extension.

Besides "prefix commands", the commands extension does have some additional features that ClientCog doesn't have, such as listeners.

For simplicity, we've been mainly sticking to using ext.commands in examples to not confuse users with having multiple versions of everything. (Eg. You *can* use nextcord.Client instead of commands.Bot, but Client is essentially just a more stripped down version of Bot)

@alentoghostflame Do you have an opinion on this?

ToasterUwU:
In my opinion the stuff from the nextcord.ext namespace should only be used if necessary. Since it's the old/other way of doing stuff, I'm also not a fan of mixing the two systems just because people got used to it.

The example things should be a minimal example of what to do, and since this is the example for the new/builtin command system, I give the minimal example for the ClientCog. No mixing systems, cleaner example.

MaskDuck:
@ToasterUwU can you post the code you tried? since `bot.load_extension` does run setup and it is only available in `commands.Bot`

ToasterUwU:
> @ToasterUwU can you post the code you tried? since `bot.load_extension` does run setup func/coro and it is only available in `commands.Bot`

@MaskDuck 

That's not the way to use it. That would be mixing old and new systems.

Use the Client instead of Bot. Client.add_cog()

MaskDuck:
...

what if i need to init some more shit in the `__init__` function?

MaskDuck:
@ToasterUwU

ooliver1:
`add_cog(Class(bot/client))` calls the `__init__` as thats how classes works

MaskDuck:
ic.

i still want to init some async stuff so i usually do that in the setup func, but fine

ooliver1:
ive made a discussion #588 so to not clutter this page (so it can be used for reviews)

ToasterUwU:
> ...
> 
> what if i need to init some more shit in the `__init__` function of the cog?

@MaskDuck Than just do it. The add_cog wants a initialized object, not the class. So do all the stuff you need and than add it. `client.add_cog(CogName(arg1=69, arg2=420))`

ToasterUwU:
I think I should close this, since it's unneeded. 

Whenever the Cog class is moved, this example, most likely, will be wrong anyways.

PR #598, opened 2022-04-12
feat(commands): add bot.load_extensions

EmreTech:
## Summary

This PR adds 2 new functions in `ext.commands.bot` that do the following:
- load all extensions from a given list of their names
- load all submodule extensions found in a module

## Checklist

- [ ] This PR fixes an issue.


H4ckerxx44:
I am not sure if this is necessary

EmreTech:
> I am not sure if this is necessary

That is a good point, but the reason I wanted to open this PR is if you have many bots, you'll have to copy-paste the same code over and over again to load extensions and honestly I think it's better if it's in the library itself.

H4ckerxx44:
> That is a good point, but the reason I wanted to open this PR is if you have many bots, you'll have to copy-paste the same code over and over again to load extensions and honestly I think it's better if it's in the library itself.

My point is, that might not be usable for you if you have a different folder structure cause this kinda requires a given folder structure which might be "normal" to use but still might not be your personal taste.

I guess that lies in epic's hands.

Skelmis:
I feel it would maybe be better to load all extensions provided, rather then finding them itself. Where the input would be a list of paths to load in current format.

nextcord/ext/commands/bot.py
```diff
@@ -901,6 +902,88 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+        ExtensionFailed
+            An extension or its setup function had an execution error.
+        """
+        # TODO: Extra arguments
+
+        loaded_extensions: List[str] = []
+
+        for i_extension in range(len(names)):
+            try:
+                self.load_extension(names[i_extension], package=package)
```
Skelmis, line 945: This needs to support all the arguments of ``load_extension`` but your todo already shows it. Just noting for review purposes

nextcord/ext/commands/bot.py
```diff
@@ -901,6 +902,88 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+            An extension is already loaded.
+        NoEntryPointError
+            An extension does not have a setup function.
+        ExtensionFailed
+            An extension or its setup function had an execution error.
+        """
+        # TODO: Extra arguments
+
+        files: List[str] = [f[:-3] for f in os.listdir(folder_name) if f.endswith(".py")]
+        extensions: List[str] = self.load_extension(files, package=folder_name)
```
Skelmis, line 984: Need the call the plural method here?

nextcord/ext/commands/bot.py
```diff
@@ -901,6 +902,88 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+        ExtensionAlreadyLoaded
+            An extension is already loaded.
+        NoEntryPointError
+            An extension does not have a setup function.
+        ExtensionFailed
+            An extension or its setup function had an execution error.
+        """
+        # TODO: Extra arguments
+
+        files: List[str] = [f[:-3] for f in os.listdir(folder_name) if f.endswith(".py")]
```
Skelmis, line 983: I dislike this, what happens when people have an ``__init__.py`` file? What happens if people have nested cog setups (I see it somewhat regularly?

nextcord/ext/commands/bot.py
```diff
@@ -901,6 +902,88 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+        """
+        # TODO: Extra arguments
+
+        loaded_extensions: List[str] = []
+
+        for i_extension in range(len(names)):
+            try:
+                self.load_extension(names[i_extension], package=package)
+            except Exception:
+                raise
```
Skelmis, line 947: Possibly add an argument to suppress errors? Could be a nice feature like ``load_extensions([..,..,..,], suppress_errors=True)`` where it loads all the ones it can and returns those while logging exceptions. Idk, need some feedback from others

nextcord/ext/commands/bot.py
```diff
@@ -901,6 +902,88 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+        NoEntryPointError
+            An extension does not have a setup function.
+        ExtensionFailed
+            An extension or its setup function had an execution error.
+        """
+        # TODO: Extra arguments
+
+        loaded_extensions: List[str] = []
+
+        for i_extension in range(len(names)):
```
Skelmis, line 943: Why not just loop over the list? Why do we need the index
```python
for extension in names:
```

Skelmis (changes requested):

nextcord/ext/commands/bot.py
EmreTech: I remember I was going to do something else that required the index, but I guess I didn't implement that
I'll go revert it

nextcord/ext/commands/bot.py
EmreTech: I think that was a typo on my part. Sorry

nextcord/ext/commands/bot.py
EmreTech: Someone else pointed out the nested cog setups, and I said I'll try to work out some recursive solution.
Also, if the file is not for a cog, then there isn't really a good way to know unless we try to load it

nextcord/ext/commands/bot.py
```diff
@@ -901,6 +903,155 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+        _raw_files: List[Path] = list(path.rglob(filter))
+
+        packages: List[str] = [str(p.parent).replace(os.sep, ".") for p in _raw_files]
+        modules: List[str] = [str(p.stem) for p in _raw_files]
+        extensions: List[str] = self.load_extensions(modules, packages=packages)
+
+        return extensions
+
+    def find_load_module_extensions(self, folder_name: str, suffix: str = ".py") -> List[str]:
+        """Similar to :meth:`find_load_extensions` except it filters out any files that aren't module init files.
```
Skelmis, line 1028: Whats the use case for this method?

EmreTech: So if you have a certain cog setup where each cog has their own folder and they have multiple Python files inside. I added it because Denver suggested it.

Skelmis: Sure, but your current implement doesn't do what you just described. All it does is recursively attempt to load all ``__init__.py`` files. And it doesn't appear to be properly recursive either when tested locally. Using the method simply yields the first found ``__init__.py`` file when viewing the returned data.

nextcord/ext/commands/bot.py
```diff
@@ -901,6 +903,155 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+
+        Once an extension found in a folder has been loaded and did not throw
+        any exceptions, it will be added to a list of extension names that
+        will be returned.
+
+        Parameters
+        ----------
+        folder_name: :class:`str`
+            The name (or path) of the folder to look through.
+        filter: :class:`str`
```
Skelmis, line 999: This should provide a link to resources denoting what a filter can be.

nextcord/ext/commands/bot.py
```diff
@@ -901,6 +903,155 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+        path = Path(folder_name)
+        _raw_files: List[Path] = list(path.rglob(filter))
+
+        packages: List[str] = [str(p.parent).replace(os.sep, ".") for p in _raw_files]
+        modules: List[str] = [str(p.stem) for p in _raw_files]
+        extensions: List[str] = self.load_extensions(modules, packages=packages)
+
+        return extensions
+
+    def find_load_module_extensions(self, folder_name: str, suffix: str = ".py") -> List[str]:
```
Skelmis, line 1027: As alluded to in discord, I don't see a point for adding this. If its that's a deal breaker, make a note in ``find_load_extensions`` for the filter to load only ``__init__.py`` files.

nextcord/ext/commands/bot.py
```diff
@@ -901,6 +903,132 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+                elif packages:
+                    self.load_extension(extension, package=cur_package, extras=cur_extra)
+            except Exception:
+                raise
+            else:
+                loaded_extensions.append(extension)
+
+        return loaded_extensions
+
+    def find_load_extensions(self, folder_name: str, filter: str = "*.py") -> List[str]:
```
DenverCoder1, line 988: I can't get this function to work at all.

File structure:

```py
.
├── main.py
└── cogs
    ├── cog1.py
    └── cog2.py
```

With `bot.find_load_extensions("cogs"))` it doesn't find any extensions at all, it's searching in the wrong location.

With `bot.find_load_extensions(os.path.join(os.path.dirname(__file__), "cogs"))`, it finds the cog files, but the parent packages are way off (eg. `'.home.username.Documents.mybot.cogs'`) and when it gets to finding the spec, it does not work and raises `ExtensionNotFound`.

DenverCoder1 (dismissed):

DenverCoder1:
Can I offer an alternative solution for `find_load_extensions` that is working for me?

https://gist.github.com/DenverCoder1/49cc433c7bf5c9a70ff12ff01e4da6ed

This supports both of these file structures:

```
.
├── main.py
└── cogs
    ├── cog1.py
    └── cog2.py
 
.
├── main.py
└── cogs
    ├── cog1
    │   └── __init__.py
    └── cog2
        ├── __init__.py
        └── extras.py
```

And can be run with `bot.find_load_extensions("cogs")` where `"cogs"` is the parent module (not path).

EmreTech:
> What's the reason to return the list of successfully loaded extensions. Seems unnecessary as that is what you give as input and the function raises an exception if any extension fails.

That is a good point for load_extensions, yes, however for find_load_extensions, it serves a different purpose. It serves the purpose of telling you the names of the extensions found & loaded in the provided cog folder. I should definitely supply the parent folder path along with the extension name.

nextcord/ext/commands/bot.py
```diff
@@ -901,6 +903,132 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+
+            try:
+                if package:
+                    self.load_extension(extension, package=package, extras=cur_extra)
+                elif packages:
+                    self.load_extension(extension, package=cur_package, extras=cur_extra)
+            except Exception:
+                raise
+            else:
+                loaded_extensions.append(extension)
```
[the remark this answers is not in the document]
EmreTech: I could, but it would append the extensions that failed, correct? And I don't want to do that. But, the exception would raise at `load_extension` and stop the code from running, so it does maybe work.

EmreTech:
> Can I offer an alternative solution for `find_load_extensions` that is working for me?
> 
> 
> 
> https://gist.github.com/DenverCoder1/49cc433c7bf5c9a70ff12ff01e4da6ed
> 
> 
> 
> This supports both of these file structures:
> 
> 
> 
> ```
> 
> .
> 
> ├── main.py
> 
> └── cogs
> 
>     ├── cog1.py
> 
>     └── cog2.py
> 
>  
> 
> .
> 
> ├── main.py
> 
> └── cogs
> 
>     ├── cog1
> 
>     │   └── __init__.py
> 
>     └── cog2
> 
>         ├── __init__.py
> 
>         └── extras.py
> 
> ```
> 
> 
> 
> And can be run with `bot.find_load_extensions("cogs")` where `"cogs"` is the parent module (not path).

I don't like this solution. It doesn't use glob syntax and it hardcodes in a prefix, which was something you didn't want according to Discord. If you just want me to warn people that `find_load_extensions` only works via a relative path, then I will.

koala9712:
A useful optional argument in find_load_extensions can be a list called something like `ignore`, so if I have multiple cogs in `cogs/` - with a `utils.py` file which isn't a cog, I can simply pass in `ignore=["utils.py"]`

DenverCoder1:
> It doesn't use glob syntax and it hardcodes in a prefix, which was something you didn't want according to Discord. If you just want me to warn people that find_load_extensions only works via a relative path, then I will.

Not sure what you mean by hardcoding a prefix. This basically works similar to load_extension, but takes the parent module instead of the submodule, and loads all submodules in the specified parent module.

I like koala's idea of being able to exclude specific modules. In my opinion, it should take the extension names eg:

```py
bot.find_load_extensions("my.cog.module", exclude=["utils"])
```

Also, maybe there is a better name for this function? I don't know what the best name would be, but I'm thinking it would make more sense if it were called something like `load_extensions_from_module` or `load_submodule_extensions`. Something in that sense.

EmreTech:
After some thinking, I think I might rename `find_load_extensions` to `load_extensions_from_module`. It sounds a little more descriptive and less wordy.

nextcord/ext/commands/bot.py
```diff
@@ -901,6 +903,135 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
             # revert sys.modules back to normal and raise back to caller
             sys.modules.update(modules)
             raise
+    
+    def load_extensions(
+            self, 
+            names: List[str], 
+            *, 
+            package: Optional[str] = None,
```
Skelmis, line 911: Possible I missed this, but where is this used in the method?

EmreTech: Where is what used in the method? The `package` attribute?

EmreTech: As discussed on Discord, this was actually an oversight on my part.

nextcord/ext/commands/bot.py
```diff
@@ -905,6 +919,138 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+            An extension could not be imported.
+        ExtensionAlreadyLoaded
+            An extension is already loaded.
+        NoEntryPointError
+            An extension does not have a setup function.
+        ExtensionFailed
+            An extension or its setup function had an execution error.
+        """
+        path = Path(folder_name)
+        if path.is_absolute():
```
Skelmis, line 1039: Why does this have to be relative?

Besides, can't you just convert the path to relative from absolute

EmreTech: It could be absolute, but I don't think Python would like having a package input of `.home.Person.Documents.MyCoolBot.cogs.Moderation` for example.

Skelmis: I see.

Possibly could get funky and remove everything up to current directory instead? Although yea, possibly more effort then its worth

EmreTech: Yeah so I think it's better to do relative for this situation
I do see where you're coming from tho

Skelmis: Feel free to mark as resolved

Skelmis (approved):

ooliver1 (changes requested):

nextcord/ext/commands/bot.py
```diff
@@ -949,6 +963,138 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+
+        Parameters
+        ----------
+        folder_name: :class:`str`
+            The name (or path) of the folder to look through. This folder path is a
+            relative path based on the current folder.
+        filter: :class:`str`
+            The filter to use when looking for extensions. Defaults to ``*.py``.
+            To learn more about the syntax of the filter, check out the Python Docs
+            (https://docs.python.org/3/library/fnmatch.html#module-fnmatch).
```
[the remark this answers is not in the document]
EmreTech: Um what? I don't quite understand what to change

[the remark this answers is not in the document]
ooliver1: `` `Python Docs <https://docs.python.org/3/library/fnmatch.html#module-fnmatch<`_ ``

nextcord/ext/commands/bot.py
```diff
@@ -949,6 +963,137 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+            An extension could not be imported.
+        ExtensionAlreadyLoaded
+            An extension is already loaded.
+        NoEntryPointError
+            An extension does not have a setup function.
+        ExtensionFailed
+            An extension or its setup function had an execution error.
+        """
+        if package and packages:
+            raise ValueError("Cannot provide both package and packages.")
```
MaskDuck, line 1011: imo raise an InvalidArgument is better

```suggestion
            raise errors.InvalidArgument("Cannot provide both package and packages.")
```

nextcord/ext/commands/bot.py
```diff
@@ -949,6 +963,137 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+        packages: Optional[List[:class:`str`]]
+            A list of packages to resolve relative imports with.
+            This is required when loading an extension using a relative path, e.g ``.foo.test``.
+            Defaults to ``None``.
+        extras: Optional[List[Dict[:class:`str`, Any]]]
+            A list of extra arguments to pass to the extension's setup function.
+
+        Returns
+        --------
+        A List[:class:`str`] that contains the names of all of the extensions
```
ooliver1, line 993:
```suggestion
        List[:class:`str`]
            A list that contains the names of all of the extensions
```

nextcord/ext/commands/bot.py
```diff
@@ -949,6 +963,137 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+            Defaults to ``None``.
+        packages: Optional[List[:class:`str`]]
+            A list of packages to resolve relative imports with.
+            This is required when loading an extension using a relative path, e.g ``.foo.test``.
+            Defaults to ``None``.
+        extras: Optional[List[Dict[:class:`str`, Any]]]
+            A list of extra arguments to pass to the extension's setup function.
+
+        Returns
+        --------
```
ooliver1, line 992:
```suggestion
        Returns
        -------
```

nextcord/ext/commands/bot.py
```diff
@@ -949,6 +963,137 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+        A List[:class:`str`] that contains the names of all of the extensions
+        that loaded successfully.
+
+        Raises
+        --------
```
ooliver1, line 997:
```suggestion
        Raises
        ------
```

nextcord/ext/commands/bot.py
```diff
@@ -949,6 +963,137 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+        folder_name: :class:`str`
+            The name (or path) of the folder to look through. This folder path is a
+            relative path based on the current folder.
+        filter: :class:`str`
+            The filter to use when looking for extensions. Defaults to ``*.py``.
+            To learn more about the syntax of the filter, check out the
+            `Python Docs. <https://docs.python.org/3/library/fnmatch.html#module-fnmatch>`_
+
+        Returns
+        --------
```
ooliver1, line 1064:
```suggestion
        Returns
        -------
```

ooliver1 (changes requested):

nextcord/ext/commands/bot.py
```diff
@@ -949,6 +963,140 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+        -------
+        List[:class:`str`]
+            A list that contains the names of all of the extensions
+        that loaded successfully.
+
+        Raises
+        ------
+        ValueError
+            - If the ``package`` and ``packages`` parameters are both provided
+            - If the length of ``packages`` or the length of ``extras` is not equal to the length of ``names``
```
MaskDuck, line 1001:
```suggestion
        ValueError
            The length of ``packages`` or the length of ``extras` is not equal to the length of ``names``
        InvalidArgument
            You passed both ``package`` and ``packages``.
```

MaskDuck (changes requested):

EmreTech:
> Can I offer an alternative solution for `find_load_extensions` that is working for me?
> 
> https://gist.github.com/DenverCoder1/49cc433c7bf5c9a70ff12ff01e4da6ed
> 
> This supports both of these file structures:
> 
> ```
> .
> ├── main.py
> └── cogs
>     ├── cog1.py
>     └── cog2.py
>  
> .
> ├── main.py
> └── cogs
>     ├── cog1
>     │   └── __init__.py
>     └── cog2
>         ├── __init__.py
>         └── extras.py
> ```
> 
> And can be run with `bot.find_load_extensions("cogs")` where `"cogs"` is the parent module (not path).

Sorry about rejecting this approach earlier, it's clear to me now that it might be better than my current implementation

EmreTech:
Also, since that alternative implementation suggested removing the returned list of the names of the extensions that loaded successfully (in both `load_extensions` and `load_extensions_from_module`), I just want to justify why I didn't make that change:
When the developer is loading extensions via `load_extensions_from_module` for example, they cannot predict which modules loaded successfully and where they are from. Because of this, it would be advantageous to tell the developer which extensions loaded successfully. Why? Well maybe they want to inform the user about the extensions that were able to load. It'll save developers some time trying to figure out which extensions loaded successfully.

Skelmis (changes requested):

ooliver1 (changes requested):

ooliver1 (changes requested):

ooliver1 (changes requested):
tested both
- `load_extensions`
  - `names`
  - `extras`
  - `stop_at_error`
- `load_extensions_from_module`
  - `source_module`
  - `stop_at_error`
 
I'm not sure how `package(s)` are meant to be used.

ooliver1 (approved):
Sorry that was meant to be an approve.

EmreTech:
> I'm not sure how package(s) are meant to be used.

It's supposed to be the source package for a relative import, and the plural is for if each extension has a different source package for relative import.

DenverCoder1:
For the docstrings, maybe an example usage could be helpful to make it more likely people will understand how to use it (eg. a `Usage ::` section)?

Also, it should probably have `versionadded`

EmreTech:
> For the docstrings, maybe an example usage could be helpful to make it more likely people will understand how to use it (eg. a `Usage ::` section)?
> 
> Also, it should probably have `versionadded`

Resolved in f654c77

Skelmis (approved):
Looks good to me, test results can be found [here](https://nexttest.koldfusion.xyz/nextcord/598)

nextcord/ext/commands/bot.py
```diff
@@ -1038,6 +1041,237 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+                class MyCogTwo(commands.Cog):
+                    def __init__(self, bot):
+                        self.bot = bot
+
+                    # ...
+
+
+                def setup(bot):
+                    bot.add_cog(MyCogTwo(bot))
+
```
ooliver1, line 1110:
```suggestion


```
for docs rendering

nextcord/ext/commands/bot.py
```diff
@@ -1038,6 +1041,237 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+                    def __init__(self, bot, my_other_attribute):
+                        self.bot = bot
+                        self.my_other_attribute = my_other_attribute
+
+                    # ...
+
+
+                def setup(bot, my_other_attribute):
+                    bot.add_cog(MyCogTwo(bot, my_other_attribute))
+
```
ooliver1, line 1150:
```suggestion


```

nextcord/ext/commands/bot.py
```diff
@@ -1038,6 +1041,237 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+        package: Optional[str] = None,
+        packages: Optional[List[str]] = None,
+        extras: Optional[List[Dict[str, Any]]] = None,
+        stop_at_error: bool = False,
+    ) -> List[str]:
+        """Loads all extensions provided in a list.
+
+        .. note::
+
+            By default, any exceptions found while loading will not be raised but will be printed on screen.
```
ooliver1, line 1057:
```suggestion
            By default, any exceptions found while loading will not be raised but will be printed to console (standard error/`stderr`).
```

nextcord/ext/commands/bot.py
```diff
@@ -1038,6 +1041,237 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+    ) -> List[str]:
+        """Loads all extensions found in a module.
+
+        Once an extension found in a module has been loaded and did not throw
+        any exceptions, it will be added to a list of extension names that
+        will be returned.
+
+        .. note::
+
+            By default, any exceptions found while loading will not be raised but will be printed on screen.
```
ooliver1, line 1217:
```suggestion
            By default, any exceptions found while loading will not be raised but will be printed to console (standard error/`stderr`).
```

nextcord/ext/commands/bot.py
```diff
@@ -1038,6 +1041,239 @@ def reload_extension(self, name: str, *, package: Optional[str] = None) -> None:
...
+
+
+                def setup(bot):
+                    bot.add_cog(MyCogTwo(bot))
+
+
+        extras: Optional[List[Dict[:class:`str`, Any]]]
+            A list of extra arguments to pass to the extension's setup function.
+
+            Usage::
```
ooliver1, line 1115:
```suggestion
            Usage ::
```

ooliver1:
Sphinx isnt very compliant

PR #530, opened 2022-03-14
App Cmd Refactor + Feature Creep

alentoghostflame:
# Summary
----------

Greatly refactors how application commands are structured code-wise to fix some issues.
Also adds localization and permissions v2 support. 
Also feature creep.

# Refactor
-----------

While the old setup is useful in its simplicity, there's a lot of overlap in code usage to the point of being limiting.
For example, User and Message commands have to rely on Slash's typing, despite not having any actual options themselves.

To fix that, I reorganized the code from being in a small 5 classes to a total of 14. Code is now spread out more with
many more `__init__`'s unfortunately, but as a result the code is better sectioned and more straight forward.

## Breaking Changes

If you stick with the decorators and don't mess with the attributes of the app cmds at all, you will likely not notice 
any changes at all and can probably ignore this section.
If you are creating application commands manually via class instantiation, these changes are extremely breaking.

* Removed `CommandOption` and replaced it with `SlashCommandOption`.
* Removed `ApplicationSubcommand` and replaced it with `SlashApplicationSubcommand`. That is not subclassed
by `(Slash)ApplicationSubcommand`.
* Removed `ApplicationCommand` and split it into 3 different classes: `SlashApplicationCommand`, 
`UserApplicationCommand`, and `MessageApplicationCommand`.

## Maybe Breaking Changes

Changes that might break things, but people probably aren't doing or don't rely on.

* App cmds now only format themselves from the callback while the `Client` or `commands.Cog` is initializing, rather 
than formatting themselves immediately.
* Slash commands no longer accept `Message` as a valid typing.
* `ConnectionState.add_application_command` will now fully remove all references before re-adding it by default. This 
fixes some bad behavior, but may cause issues for people that rely on it. Why would you rely on that?
* `user_command` and `message_command` decorators and their associated classes no longer accept description as a kwarg.
This errored anyways when the bot tried syncing commands with Discord.

## Notes

Many methods in `Client` and `ConnectionState` regarding command discovery and syncing have been renamed. Deprecation 
notices have been added to old method names to make migrating over easier.

* `ConnectionState.remove_application_command` will no longer error if given a command that isn't in the state.
* Moved application command specific kwargs for `Client` into the `__init__`, rather than popping from options.

# Additions
------------

With the changes made to the application command code, new things are possible. Feature creep is both great and sucks.

## `OptionConverter`

Styled after `ext.commands` Converters, `OptionConverter` allows you to do custom typehinting that can modify both the 
[...]

PR #606, opened 2022-04-17
feat: fix type issues for the whole library

ooliver1:
# NOTE: this does not target master

## Summary

typehitns the whole library, albeit a little lazy with type ignores
does not fix application command related typings as to not conflict with #530 

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/__main__.py
+++ b/nextcord/__main__.py
@@ -21,6 +21,7 @@ LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
 FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
 DEALINGS IN THE SOFTWARE.
 """
+from __future__ import annotations
 
 import argparse
 import sys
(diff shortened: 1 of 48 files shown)


nextcord/__main__.py
```diff
@@ -155,7 +156,7 @@ async def cog_after_invoke(self, ctx):
 # certain file names and directory names are forbidden
 # see: https://msdn.microsoft.com/en-us/library/windows/desktop/aa365247%28v=vs.85%29.aspx
 # although some of this doesn't apply to Linux, we might as well be consistent
-_base_table = {
+_base_table: dict[str, str | None] = {
```
EmreTech, line 159: Wasn't this Union syntax introduced in Python 3.10?

ooliver1: Correct, note the future import that has been supported since 3.7 https://peps.python.org/pep-0563/#enabling-the-future-behavior-in-python-3-7

EmreTech: Oh right. I forgot about that.

ooliver1:
slotscheck decided to fail after passing

ooliver1:
okay i guess you have to trust me, see how those failing points are fixed by that commit

ooliver1:
targets a branch, can always be tested later

PR #607, opened 2022-04-18
feat: support more input types where files are used

RamzziSudip:
## Summary

- Accept multiple formats (eg. Asset, Attachment, or bytes) and read the object conditionally based on the type.
It adds extra time complexity only when users add Asset, Attachment, File for `await avatar.read()`  and `avatar.fp.read()` else it'll simply pass the bytes which takes no extra time.

## Checklist

- [ ] This PR fixes an issue.

 ## Testing
```py
    @group(pass_context=True, aliases=["t"])
    async def test(self, ctx: Context):
        """Testing Group commands for bot"""
        if ctx.invoked_subcommand is None:
            await ctx.send('Invalid sub command passed')

    @test.command(aliases=["pe"])
    async def pfp_edit(self, ctx: Context):
        old_avatar = self.bot.user.avatar
        await self.bot.user.edit(avatar=None)
        await sleep(20)
        await self.bot.user.edit(avatar=ctx.message.attachments[0])
        await sleep(20)
        await self.bot.user.edit(avatar=old_avatar)
        await sleep(20)
        await self.bot.user.edit(avatar=File("test/image.jpeg"))
        await ctx.message.add_reaction("👍")

    @test.command(aliases=["sec"])
    async def scheduled_event_create(self, ctx: Context, channel: VoiceChannel):
        await ctx.guild.create_scheduled_event(
            entity_type=ScheduledEventEntityType.voice,
            start_time=ctx.message.created_at+timedelta(minutes=1),
            name="Testing", channel=channel,
            image=ctx.message.attachments[0]
        )
        await ctx.message.add_reaction("👍")

    @test.command(aliases=["see"])
    async def scheduled_event_edit(self, ctx: Context, event_id: int):
        event = ctx.guild.get_scheduled_event(event_id)
        old_image = event.image
        await event.edit(image=None)
        await sleep(20)
        await event.edit(image=ctx.message.attachments[0])
        await sleep(20)
        await event.edit(image=old_image)
        await sleep(20)
        await event.edit(image=File("test/image.jpeg"))
        await ctx.message.add_reaction("👍")

    @test.command(aliases=["cr"])
    async def create_role(self, ctx: Context):
        await ctx.guild.create_role(
            name="TESTING",
            color=Color.brand_red(),
            icon=ctx.message.attachments[0]
        )
        await ctx.message.add_reaction("👍")

    @test.command(aliases=["gc"])
    async def guild_create(self, ctx: Context, template_code: str = None):
        guild = await self.bot.create_guild(
            name="TESTING",
            icon=ctx.message.attachments[0],
            code=template_code
        )
        channel = await guild.create_text_channel("demo")
        invite = await channel.create_invite()
        await ctx.author.send(invite.url)
        await ctx.message.add_reaction("👍")

    @test.command(aliases=["ge"])
    async def guild_edit(self, ctx: Context, guild: Guild):
        await guild.edit(icon=None)
        await sleep(20)
        await guild.edit(icon=ctx.message.attachments[0])
        await sleep(20)
        await guild.edit(icon=ctx.author.display_avatar)
        await sleep(20)
        await guild.edit(icon=File("test/image.jpeg"))

    @test.command(aliases=["gr"])
    async def guild_remove(self, ctx: Context, guild: Guild):
        await guild.delete()
        await ctx.message.add_reaction("👍")

    @test.command(aliases=["wc"])
    async def webhook_create(self, ctx: Context):
        await ctx.channel.create_webhook(
            name="Testing",
            avatar=ctx.message.attachments[0]
        )
        await ctx.message.add_reaction("👍")

    @test.command(aliases=["awe"])
    async def async_webhook_edit(self, ctx: Context):
        webhook = find(lambda wh: wh.user.id == self.bot.user.id, (await ctx.channel.webhooks()))
        await webhook.edit(avatar=None)
        await sleep(20)
        await webhook.edit(avatar=ctx.message.attachments[0])
        await sleep(20)
        await webhook.edit(avatar=ctx.author.display_avatar)
        await sleep(20)
        await webhook.edit(avatar=File("test/image.jpeg"))
        await ctx.message.add_reaction("👍")
```


DenverCoder1:
This is currently a breaking change since it changes the parameter type.

I don't agree with making the type Asset since it will be harder to use if the avatar is coming from another source such as a saved file or an attachment.

It may be acceptable if it were to accept multiple formats (eg. Asset, Attachment, or bytes) and read the object conditionally based on the type.

I don't know if it's all that necessary as it's not complicated to get bytes from those formats beforehand.

ooliver1:
I agree with denver, do remember that webhooks may be used outside of a bot and he standalone

koala9712:
^ Agreed, it could be made optional or something, perhaps

RamzziSudip:
@DenverCoder1  Tell me what you think now. #607

nextcord/channel.py
```diff
@@ -578,6 +584,8 @@ async def create_webhook(self, *, name: str, avatar: Optional[bytes] = None, rea
         from .webhook import Webhook
 
         if avatar is not None:
```
koala9712, line 586: This can be changed to simply `if avatar:`, just a suggestion, up to you depending on the scope of the PR.

koala9712:
Has this been tested? If yes, please share testing code used.

DenverCoder1:
This is no longer a breaking change since avatar still accepts bytes now.

If this is changed, it should probably also be changed in some or all of the following for consistency:

- Webhook.edit
- SyncWebhook.edit
- ClientUser.edit
- Bot.create_guild
- Guild.edit
- Guild.create_custom_emoji
- Guild.create_role?
- Guild.create_scheduled_event
- ScheduledEvent.edit

nextcord/channel.py
DenverCoder1: I don't think changing this is necessary. I would say checking that it is not None is better since you are explicitly checking if the parameter was set to something, rather than checking if it's truthy. In this case, it might not make a difference in practice, but theoretically, it could make a difference (eg. there could be some falsy bytes-like object the user might want to use as their default).

nextcord/guild.py
```diff
@@ -2513,34 +2503,6 @@ async def fetch_roles(self, *, cache: bool = False) -> List[Role]:
 
         return roles
 
-    @overload
-    async def create_role(
```
DenverCoder1, line 2517: Why remove the overloads?

nextcord/channel.py
```diff
@@ -577,8 +583,8 @@ async def create_webhook(self, *, name: str, avatar: Optional[bytes] = None, rea
 
         from .webhook import Webhook
 
-        if avatar is not None:
-            avatar = utils._bytes_to_base64_data(avatar)  # type: ignore
+        avatar = avatar if avatar is None else utils._bytes_to_base64_data(
+            avatar if isinstance(avatar, bytes) else await avatar.read())
```
DenverCoder1, line 587: It's best to avoid having multiple conditional operations in a single statement. This makes the code more difficult to comprehend.

I'd recommend spreading it out across multiple lines, or use a `utils` function to abstract the logic.

nextcord/guild.py
```diff
@@ -2621,14 +2585,14 @@ async def create_role(
             fields['name'] = name
 
         if icon is not MISSING:
-            if icon is None:
-                fields['icon'] = icon
-            elif isinstance(icon, str):
+            if isinstance(icon, str):
```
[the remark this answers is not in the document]
RamzziSudip: In Discord api docs I saw this.
image

By default, the icon is `null`. So we don't have to pass `None`. If we want no icon for our new role then we don't need to feed the icon field in the create_role() method, easy. It'll automatically check with `MISSING` and ignore the field.

But if still a user passes create_role(icon=None) it'll give a warning.
pycharm warning

And eventually, at runtime it'll fail because it'll hit the `await icon.read()` which is actually `await None.read()`
Error if passes None

Do you think we should do this? which doesn't make sense to me though.
```py
if icon is not MISSING:
    if icon is None:
        pass
    elif isinstance(icon, str):
        fields['unicode_emoji'] = icon
    elif isinstance(icon, bytes):
        fields['icon'] = utils._bytes_to_base64_data(icon)
    elif isinstance(icon, File):
        fields['icon'] = utils._bytes_to_base64_data(icon.fp.read())
    else:
        fields['icon'] = utils._bytes_to_base64_data(await icon.read())
```

Or this looks better to me without using the `MISSING`👇

```py
async def create_role(
    self,
    *,
    name: str = MISSING,
    permissions: Permissions = MISSING,
    color: Union[Colour, int] = MISSING,
    colour: Union[Colour, int] = MISSING,
    hoist: bool = MISSING,
    mentionable: bool = MISSING,
    icon: Union[str, bytes, File, Asset, Attachment] = None,
    reason: Optional[str] = None
) -> Role:
....

        if icon is not None:
            if isinstance(icon, str):
                fields['unicode_emoji'] = icon
            elif isinstance(icon, bytes):
                fields['icon'] = utils._bytes_to_base64_data(icon)
            elif isinstance(icon, File):
                fields['icon'] = utils._bytes_to_base64_data(icon.fp.read())
            else:
                fields['icon'] = utils._bytes_to_base64_data(await icon.read())
```

nextcord/guild.py
RamzziSudip: ## What is `@overload` ?

### **`@overload` allows describing methods that support multiple different combinations of argument types.**
I saw this from here... 
image

## Why it was there? (from my observation)
- These are two parameters called `color`  and  `colour`.
- These are basically the same, `colour` is an aliase of `color`. 
- For describing both in create_role() method `@overload` was there.

## Why it's not necessary then? 
- Every other argument and return types were exactly the same in those `@overload` create_role() only `color`  and  `colour` changed.
- We can just add those two arguments `color` and `colour` normally (which was already there) and I just added the docstring for `color` which was missing.
- `@overload` is not really useful for this case from my point of view.

ooliver1: Doesn't seem to describe why you removed it, overloads can show possible signatures, signatures do not include passing both color and colour

DenverCoder1: For hinting purposes it's better since you can specify `color` and not `colour`, or specify `colour` and not `color`, but you can't specify both.

While it's not mandatory to have, I don't see a reason to remove it.

nextcord/guild.py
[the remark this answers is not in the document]
DenverCoder1: I think it should actually still set it `None` if `None` is passed, since Discord accepts null (and they _could_ technically change their default).

It fits the API specification better if allows passing None explicitly.

Example:

```py
if icon is None or isinstance(icon, str)
    fields['unicode_emoji'] = icon
```

nextcord/guild.py
```diff
@@ -3204,8 +3208,8 @@ async def create_scheduled_event(
         privacy_level: ScheduledEventPrivacyLevel = ScheduledEventPrivacyLevel.guild_only,
         end_time: datetime.datetime = MISSING,
         description: str = MISSING,
-        image: bytes = MISSING,
-        reason: Optional[str] = None,
+        image: Union[bytes, Asset, Attachment] = None,
+        reason: Optional[str] = None
```
DenverCoder1, line 3212: Trailing commas are good. For one thing, it allows changing fewer lines when adding parameters.

```suggestion
        reason: Optional[str] = None,
```

nextcord/guild.py
```diff
@@ -3204,8 +3208,8 @@ async def create_scheduled_event(
         privacy_level: ScheduledEventPrivacyLevel = ScheduledEventPrivacyLevel.guild_only,
         end_time: datetime.datetime = MISSING,
         description: str = MISSING,
-        image: bytes = MISSING,
-        reason: Optional[str] = None,
+        image: Union[bytes, Asset, Attachment] = None,
```
DenverCoder1, line 3211: Denote Optional if it accepts None

```suggestion
        image: Optional[Union[bytes, Asset, Attachment]] = None,
```

nextcord/guild.py
```diff
@@ -3229,18 +3233,21 @@ async def create_scheduled_event(
             The description for the event
         entity_type: :class:`ScheduledEventEntityType`
             The type of event
-        image: :class:`bytes`
+        image: Union[:class:`bytes`, :class:`Asset`, :class:`Attachment`]
```
DenverCoder1, line 3236:
```suggestion
        image: Optional[Union[:class:`bytes`, :class:`Asset`, :class:`Attachment`]]
```

DenverCoder1 (changes requested):

DenverCoder1 (changes requested):
:+1: It's looking good. Just a few small things

RamzziSudip (commented):
Everything is ready here maybe? or anything else to change?

nextcord/guild.py
```diff
@@ -2523,7 +2532,7 @@ async def create_role(
         colour: Union[Colour, int] = ...,
         hoist: bool = ...,
         mentionable: bool = ...,
-        icon: Optional[Union[str, bytes, File]] = ...,
+        icon: Optional[Union[str, bytes, File, Asset, Attachment]] = ...,
```
DenverCoder1, line 2535: For consistency, maybe `File` could be removed from role icon parameters?

I would be fine with that breaking change as I don't see any reason someone would need to pass a `File`. It's more of just a utility class for sending files in messages.

I'm fine with leaving things the way they are though, and I'm open to hearing other opinions on this.

nextcord/guild.py
```diff
@@ -2580,10 +2589,10 @@ async def create_role(
         mentionable: :class:`bool`
             Indicates if the role should be mentionable by others.
             Defaults to ``False``.
+        icon: Optional[Union[:class:`str`, :class:`bytes`, :class:`Asset`, :class:`Attachment`]]
```
DenverCoder1, line 2592: If it is left alone, then `File` should be added here. If `File` is removed, as proposed above, then this is fine.

DenverCoder1 (approved):
Looks good, I haven't tested it

nextcord/guild.py
RamzziSudip: I also thought about it. I was confused why File was there in the first place...
Maybe because Asset and Attachment wasn't there

DenverCoder1:
It seems like `Role.edit` and `Template.create_guild` are a couple more that could be changed

RamzziSudip:
> 

How do you find those subtle changes 😂. Do you memorize everything object and classes of Nextcord?

DenverCoder1:
Mostly just from using Ctrl+F.

To make the original list of methods, I just did a quick search through the API reference docs. Second time, I did a Ctrl+F in the codebase for ": .*:class:`bytes`" to find all parameters that take bytes. There is one function that takes audio as bytes, but everything else was an image parameter.

RamzziSudip:
@ooliver1  Your turn to review the code. I think everything is ready for merging. I've tested everything and it's working. But you can also double-check.

DenverCoder1:
From a recent poll, it seems like the option to pass `File` is wanted as it does abstract the step of having to open the file as binary and reading the file, i.e. you'd be able to simply pass `File("file.png")` instead of `open("file.png", "rb").read()`.

image

So it seems like `SyncWebhook.edit` can probably at least take `File`, and everything else can have `File` added in addition to the other types.

ooliver1:
when you say you have tested, i would appreciate some code so its easier for me to confirm thats the case

nextcord/enums.py
```diff
@@ -177,6 +178,16 @@ def try_value(cls, value):
             except (KeyError, TypeError):
                 return value
 
+        @property
+        def name(self):
+            """The name of the Enum member."""
+            return self.name
+
+        @property
+        def value(self):
```
RamzziSudip, line 187: @DenverCoder1  
## Chech this... 
### It can fix pycharm warnings like following:
image
image

nextcord/enums.py
```diff
@@ -87,7 +88,7 @@ class EnumMeta(type):
         _enum_member_map_: ClassVar[Dict[str, Any]]
         _enum_value_map_: ClassVar[Dict[Any, Any]]
 
-    def __new__(cls, name, bases, attrs, *, comparable: bool = False):
+    def __new__(mcs, name, bases, attrs, *, comparable: bool = False):
```
RamzziSudip, line 91: This is a metaclass so `mcs` might "fix an inconsistency".

RamzziSudip:
> when you say you have tested, i would appreciate some code so its easier for me to confirm thats the case

### @ooliver1 I haven't tested the newly added commits like:

-  `Template.create_guild`,
-  `Role.edit` and `Guild.create_role` because my testing server is not boosted so can't do anything with `icon` kwarg
- `SyncWebhook.edit`
- and the `File` type support for all image/avatar/icon kwarg

### This is all of my testing code so far...
```py
@group(pass_context=True, aliases=["t"])
async def test(self, ctx: Context):
    """Testing Group commands for bot"""
    if ctx.invoked_subcommand is None:
        await ctx.send('Invalid sub command passed')
        
@test.command(aliases=["pe"])
async def pfp_edit(self, ctx: Context):
    old_avatar = self.bot.user.avatar
    await self.bot.user.edit(avatar=None)
    await sleep(20)
    await self.bot.user.edit(avatar=ctx.message.attachments[0])
    await sleep(20)
    await self.bot.user.edit(avatar=old_avatar)
    
@test.command(aliases=["sec"])
async def scheduled_event_create(self, ctx: Context, channel: VoiceChannel):
    await ctx.guild.create_scheduled_event(
        entity_type=ScheduledEventEntityType.voice,
        start_time=ctx.message.created_at+timedelta(minutes=1),
        name="Testing", channel=channel,
        image=ctx.message.attachments[0]
    )
    
@test.command(aliases=["see"])
async def scheduled_event_edit(self, ctx: Context, event_id: int):
    event = ctx.guild.get_scheduled_event(event_id)
    old_image = event.image
    await event.edit(image=None)
    await sleep(20)
    await event.edit(image=ctx.message.attachments[0])
    await sleep(20)
    await event.edit(image=old_image)
    
@test.command(aliases=["cr"])
async def create_role(self, ctx: Context):
    await ctx.guild.create_role(
        name="TESTING",
        color=Color.brand_red(),
        icon=ctx.message.attachments[0]
    )
@test.command(aliases=["gc"])
async def guild_create(self, ctx: Context):
    guild = await self.bot.create_guild(
        name="TESTING",
        icon=ctx.message.attachments[0]
    )
    channel = await guild.create_text_channel("demo")
    invite = await channel.create_invite()
    await ctx.author.send(invite.url)
    
@test.command(aliases=["ge"])
async def guild_edit(self, ctx: Context, guild: Guild):
    await guild.edit(icon=None)
    await sleep(20)
    await guild.edit(icon=ctx.message.attachments[0])
    await sleep(20)
    await guild.edit(icon=ctx.author.display_avatar)
    
@test.command(aliases=["wc"])
async def webhook_create(self, ctx: Context):
    await ctx.channel.create_webhook(
        name="Testing",
        avatar=ctx.message.attachments[0]
    )
    
@test.command(aliases=["awe"])
async def async_webhook_edit(self, ctx: Context):
    webhook = find(lambda wh: wh.user.id == self.bot.user.id, (await ctx.channel.webhooks()))
    await webhook.edit(avatar=None)
    await sleep(20)
    await webhook.edit(avatar=ctx.message.attachments[0])
    await sleep(20)
    await webhook.edit(avatar=ctx.author.display_avatar)
```

nextcord/enums.py
DenverCoder1: I think it can be left as `cls` as `mcs` would conflict with [PEP 8](https://peps.python.org/pep-0008/#function-and-method-arguments)

> Always use `cls` for the first argument to class methods.

Also, it's better not to change things unrelated to the main focus of the PR since it makes it harder to review.

It's better to create a separate branch and PR for each issue fixed.

A lot of refactoring and type stuff in one PR is fine, but it shouldn't be in the same PR as a new feature in my opinion.

nextcord/user.py
```diff
@@ -159,7 +160,11 @@ def avatar(self) -> Optional[Asset]:
 
     @property
     def default_avatar(self) -> Asset:
-        """:class:`Asset`: Returns the default avatar for a given user. This is calculated by the user's discriminator."""
+        """
+        :class:`Asset`: Returns the default avatar for a given user.
+
+        This is calculated by the user's discriminator.
+        """
```
DenverCoder1, line 167: Starting on the first line would be more consistent with the other docstrings

```suggestion
        """:class:`Asset`: Returns the default avatar for a given user.
        
        This is calculated by the user's discriminator.
        """
```

DenverCoder1 (changes requested):

nextcord/guild.py
```diff
@@ -2621,19 +2642,19 @@ async def create_role(
...
+            elif isinstance(icon, bytes):
                 fields['icon'] = utils._bytes_to_base64_data(icon)
+            else:
+                fields['icon'] = utils._bytes_to_base64_data(await icon.read())
 
         data = await self._state.http.create_role(self.id, reason=reason, **fields)
         role = Role(guild=self, data=data, state=self._state)
 
-        # TODO: add to cache
+        self._add_role(role)
```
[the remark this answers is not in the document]
RamzziSudip: Should I remove the TODO comment then?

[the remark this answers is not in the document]
DenverCoder1: Yeah, I think so

nextcord/template.py
```diff
@@ -166,7 +169,12 @@ def __repr__(self) -> str:
             f' creator={self.creator!r} source_guild={self.source_guild!r} is_dirty={self.is_dirty}>'
         )
 
-    async def create_guild(self, name: str, region: Optional[VoiceRegion] = None, icon: Any = None) -> Guild:
+    async def create_guild(
+        self,
+        name: str,
+        region: Union[VoiceRegion, str] = VoiceRegion.us_west,
```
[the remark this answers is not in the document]
RamzziSudip: This is actually a big breaking change
I have to change many internal requests in `http.py` and also have to edit `Client.create_guild`

nextcord/enums.py
ooliver1: How is this related to the pr?
Type issues should target other prs

nextcord/template.py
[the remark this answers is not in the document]
ooliver1: Or simply revert your change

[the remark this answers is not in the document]
DenverCoder1: I'd recommend just leaving `region` alone, it's unrelated to the feature this PR adds. It should preferably be removed in a separate PR.

nextcord/role.py
```diff
@@ -46,6 +45,8 @@
         RoleTags as RoleTagPayload,
     )
     from .types.guild import RolePositionUpdate
+    from .message import Attachment
+    from .file import File
```
DenverCoder1, line 49: I'm pretty sure the `File` imports need to be outside `if TYPE_CHECKING` if you are using it in `isinstance`.

If it works, you can ignore this, but I expect this would give a NameError since it's a runtime check.

nextcord/template.py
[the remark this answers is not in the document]
RamzziSudip: @DenverCoder1  I'll work on the region which is deprecated from guild object in a different pr in future

nextcord/role.py
RamzziSudip: image
 ## You were right.. Ima fix it

DenverCoder1:
What are your thoughts on reducing the repetition by making a method in utils to convert objects to base64?

Something like this, for example:
```py
async def _obj_to_base64_data(obj: Optional[Union[bytes, Attachment, Asset, File]]) -> Optional[str]:
    if obj is None:
        return obj
    if isinstance(obj, bytes):
        return _bytes_to_base64_data(obj)
    elif isinstance(obj, File):
        return _bytes_to_base64_data(obj.fp.read())
    else:
        return _bytes_to_base64_data(await obj.read())
```

That way, everywhere else in the code you'd be able to convert anything to base64 with one line instead of 6 or more.

RamzziSudip:
> What are your thoughts on reducing the repetition by making a method in `nextcord.utils` to convert objects to base64?
> 
> Something like this, for example:
> 
> ```python
> async def _obj_to_base64_data(obj: Optional[Union[bytes, Attachment, Asset, File]]) -> Optional[str]:
>     if obj is None:
>         return obj
>     if isinstance(obj, bytes):
>         return _bytes_to_base64_data(obj)
>     elif isinstance(obj, File):
>         return _bytes_to_base64_data(obj.fp.read())
>     else:
>         return _bytes_to_base64_data(await obj.read())
> ```
> 
> That way, everywhere else in the code you'd be able to convert anything to base64 with one line instead of 6 or more.

This is creating a circular loop for importing "File" in utils.py any solution?

nextcord/role.py
```diff
@@ -449,15 +450,9 @@ async def edit(
             payload['mentionable'] = mentionable
 
         if icon is not MISSING:
-            if icon is None:
-                payload['icon'] = icon
-            elif isinstance(icon, str):
+            if icon is None or isinstance(icon, str):
```
DenverCoder1, line 453: Don't need to set unicode emoji if icon is `None`

```suggestion
            if isinstance(icon, str):
```

nextcord/role.py
```diff
@@ -449,15 +450,9 @@ async def edit(
...
                 payload['unicode_emoji'] = icon
-                payload['icon'] = None
-            elif isinstance(icon, File):
-                payload['icon'] = _bytes_to_base64_data(icon.fp.read())
-            else:
-                payload['icon'] = _bytes_to_base64_data(icon)
+            payload['icon'] = await _obj_to_base64_data(icon)
```
DenverCoder1, line 455: This should probably be an else, since we don't want to set `icon` if it's a unicode emoji?

(Or possibly `icon` could be `None` if it's unicode?)

```suggestion
            else:
                payload['icon'] = await _obj_to_base64_data(icon)
```

nextcord/guild.py
```diff
@@ -2621,19 +2610,13 @@ async def create_role(
             fields['name'] = name
 
         if icon is not MISSING:
-            if icon is None:
-                fields['icon'] = icon
-            elif isinstance(icon, str):
+            if icon is None or isinstance(icon, str):
```
DenverCoder1, line 2613: Same as above
```suggestion
            if isinstance(icon, str):
```

nextcord/guild.py
```diff
@@ -2621,19 +2610,13 @@ async def create_role(
...
                 fields['unicode_emoji'] = icon
-            elif isinstance(icon, File):
-                fields['icon'] = utils._bytes_to_base64_data(icon.fp.read())
-            else:
-                fields['icon'] = utils._bytes_to_base64_data(icon)
+            fields['icon'] = await utils._obj_to_base64_data(icon)
```
DenverCoder1, line 2615: Same as above
```suggestion
            else:
                fields['icon'] = await _obj_to_base64_data(icon)
```

DenverCoder1 (changes requested):

nextcord/guild.py
```diff
@@ -1559,27 +1560,16 @@ async def edit(
 
         if icon is not MISSING:
             if icon is None:
-                fields['icon'] = icon
-            else:
-                fields['icon'] = utils._bytes_to_base64_data(icon)
+                fields['icon'] = await utils._obj_to_base64_data(icon)
```
DenverCoder1, line 1563: Pretty sure the `if icon is None:` line should not be there

```diff
-            if icon is None:
-                fields['icon'] = await utils._obj_to_base64_data(icon)
+            fields['icon'] = await utils._obj_to_base64_data(icon)
```

RamzziSudip:
Why pre-commit.ci is failing -_-

DenverCoder1 (approved):
Looks good, I haven't tested

ooliver1:
> Why pre-commit.ci is failing -_-

sorry, it shouldnt have been enabled until the relevant PR was merged

nextcord/enums.py
ooliver1: this was not resolved, why was it marked as so

if you believe it is resolved, tell me why, otherwise dont blank me

nextcord/guild.py
```diff
@@ -2621,14 +2628,14 @@ async def create_role(
             fields['name'] = name
 
         if icon is not MISSING:
-            if icon is None:
-                fields['icon'] = icon
-            elif isinstance(icon, str):
+            if isinstance(icon, str):
```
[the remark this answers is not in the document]
ooliver1: once again, this was never resolved, please add comments on why you believe this is resolved in this thread

nextcord/template.py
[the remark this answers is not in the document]
ooliver1: the code was never edited to revert your change

nextcord/enums.py
RamzziSudip: Ooo sorry, I didn't notice it...
You are right. That is not related to this pr... Should I revert the changes or let it be like that? 
It actually fixes warnings though.

nextcord/template.py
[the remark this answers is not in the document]
RamzziSudip: Now it's reverted

nextcord/guild.py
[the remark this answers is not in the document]
RamzziSudip: Now, this is resolved because we are now using utils._obj_to_base64_data which allows passing `None` explicitly to the icon kwarg.
```py
        if icon is not MISSING:
            if isinstance(icon, str):
                fields['unicode_emoji'] = icon
            else:
                fields['icon'] = await utils._obj_to_base64_data(icon)
```

nextcord/channel.py
```diff
@@ -556,7 +563,7 @@ async def create_webhook(self, *, name: str, avatar: Optional[bytes] = None, rea
         -------------
         name: :class:`str`
             The webhook's name.
-        avatar: Optional[:class:`bytes`]
+        avatar: Optional[Union[:class:`bytes`, :class:`Asset`, :class:`Attachment`, :class:`File`]]
             A :term:`py:bytes-like object` representing the webhook's default avatar.
```
koala9712, line 567: Docuemntation such as this should be fixed throughout the PR,
L566 says it accepts bytes/asset/attachment/file, but it's description does not match that

nextcord/utils.py
```diff
@@ -497,6 +500,17 @@ def _bytes_to_base64_data(data: bytes) -> str:
     return fmt.format(mime=mime, data=b64)
 
 
+async def _obj_to_base64_data(obj: Optional[Union[bytes, Attachment, Asset, File]]) -> Optional[str]:
```
koala9712, line 503: Fix return type, may be typehinted as `Optional[Union[str, None]]` instead

DenverCoder1: The way it is seems more correct to me.

`Optional[str]` means it returns a string or None (it means the same as `Union[str, None]` but more concise)

RamzziSudip: > Fix return type, may be typehinted as `Optional[Union[str, None]]` instead

`Optional[T]` already means `Optional[Union[T, None]]`

nextcord/enums.py
Skelmis: This change should be reverted and put into a new pull request which is specifically for resolving these kinds of issues

nextcord/webhook/sync.py
```diff
@@ -743,7 +743,7 @@ def edit(
         *,
         reason: Optional[str] = None,
         name: Optional[str] = MISSING,
-        avatar: Optional[bytes] = MISSING,
+        avatar: Optional[bytes, File] = MISSING,
```
DenverCoder1, line 746:
```suggestion
        avatar: Optional[Union[bytes, File]] = MISSING,
```

nextcord/utils.py
koala9712: Holy foof yes, I forgor typehinting, thanks

nextcord/client.py
```diff
@@ -1444,17 +1446,15 @@ async def create_guild(
...
             added to cache.
         """
-        if icon is not MISSING:
-            icon_base64 = utils._bytes_to_base64_data(icon)
-        else:
-            icon_base64 = None
+        icon_base64 = await utils._obj_to_base64_data(icon)
 
-        region_value = str(region)
+        if isinstance(region, VoiceRegion):
```
ooliver1, line 1451: this seems unrelated

nextcord/enums.py
```diff
@@ -180,6 +180,16 @@ def try_value(cls, value):
             except (KeyError, TypeError):
                 return value
 
+        @property
+        def name(self):
+            """The name of the Enum member."""
+            return self.name
```
ooliver1, line 186: also unrelated

ooliver1 (changes requested):

EmreTech (approved):
Besides the other suggestions, it looks fine. Also, make sure to resolve the Pyright errors!

nextcord/guild.py
```diff
@@ -2496,17 +2491,14 @@ async def create_custom_emoji(
         :class:`Emoji`
             The created emoji.
         """
-
-        img = utils._bytes_to_base64_data(image)
-
-        role_ids: SnowflakeList
```
DenverCoder1, line 2502: The removal of `role_ids: SnowflakeList` seems unrelated to the PR. This adds a type annotation for the variable.

nextcord/scheduled_events.py
```diff
@@ -24,28 +24,26 @@
...
 from .mixins import Hashable
-from .utils import MISSING, _bytes_to_base64_data, parse_time
+from .utils import MISSING, _obj_to_base64_data, parse_time
 
-__all__: Tuple[str, ...] = (
-    "EntityMetadata",
-    "ScheduledEventUser",
-    "ScheduledEvent",
-)
+__all__: Tuple[str, ...] = ("EntityMetadata", "ScheduledEventUser", "ScheduledEvent")
```
DenverCoder1, line 36: Removing this comma doesn't seem necessary
```suggestion
__all__: Tuple[str, ...] = (
    "EntityMetadata",
    "ScheduledEventUser",
    "ScheduledEvent",
)
```

nextcord/http.py
```diff
@@ -1469,9 +1469,9 @@ def create_custom_emoji(
         self,
         guild_id: Snowflake,
         name: str,
-        image: bytes | str,
+        image: Optional[str],
         *,
-        roles: Optional[SnowflakeList] = None,
+        roles: Optional[List[int]] = None,
```
DenverCoder1, line 1474: Any reason this can't be SnowflakeList?

DenverCoder1:
Probably the last thing. I hope it's not too much...

I think the docstrings might need some updating.

The attributes that were changed probably should not just reference bytes, for example:

```diff
- A :term:`py:bytes-like object` representing the webhook's default avatar.
+ A :term:`py:bytes-like object`, :class:`File`, :class:`Attachment`, or :class:`Asset` representing the webhook's default avatar.
```

And assuming it's targeting 2.1, change info would be nice, for example:

```
.. versionchanged:: 2.1
    The :attr:`avatar` attribute now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```

nextcord/guild.py
```diff
@@ -2458,46 +2455,48 @@ async def create_custom_emoji(
...
         name: str,
-        image: bytes,
+        image: Union[bytes, Asset, Attachment, File],
         roles: List[Role] = MISSING,
         reason: Optional[str] = None,
     ) -> Emoji:
         r"""|coro|
 
-        Creates a custom :class:`Emoji` for the guild.
+            Creates a custom :class:`Emoji` for the guild.
```
DenverCoder1, line 2464: The indentation seems inconsistent with this docstring

nextcord/role.py
```diff
@@ -409,7 +413,7 @@ async def edit(
         position: :class:`int`
             The new role's position. This must be below your top role's
             position or it will fail.
-        icon: Union[:class:`str`, :class:`bytes`, :class:`File`]
+        icon: Optional[Union[:class:`str`, :class:`bytes`, :class:`Asset`, :class:`Attachment`, :class:`File`]]
             The role's icon image
```
DenverCoder1, line 417: This can probably be made consistent with the other docstrings.

nextcord/webhook/sync.py
```diff
@@ -753,7 +753,7 @@ def edit(
         ------------
         name: Optional[:class:`str`]
             The webhook's new default name.
-        avatar: Optional[:class:`bytes`]
+        avatar: Optional[Union[:class:`bytes`, :class:`File`]]
             A :term:`py:bytes-like object` representing the webhook's new default avatar.
```
DenverCoder1, line 757: File was added for sync webhooks

DenverCoder1 (approved):

Skelmis (approved):

nextcord/channel.py
```diff
@@ -561,12 +566,16 @@ async def create_webhook(
         .. versionchanged:: 1.1
             Added the ``reason`` keyword-only parameter.
 
+        .. versionchanged:: 2.1
+            The :attr:`avatar` attribute now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```
ooliver1, line 570:
```suggestion
            The ``avatar`` parameter now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```

nextcord/client.py
```diff
@@ -1468,16 +1470,19 @@ async def create_guild(
 
         Bot accounts in more than 10 guilds are not allowed to create guilds.
 
+        .. versionchanged:: 2.1
+            The :attr:`icon` attribute now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```
ooliver1, line 1474:
```suggestion
            The ``icon`` parameter now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```

nextcord/guild.py
```diff
@@ -1490,31 +1497,33 @@ async def edit(
         .. versionchanged:: 2.0
             The newly updated guild is returned.
 
+        .. versionchanged:: 2.1
+            The :attr:`icon`, :attr:`banner`, :attr:`splash`, :attr:`discovery_splash`
+             attributes now accept :class:`File`, :class:`Attachment`, and :class:`Asset`.
```
ooliver1, line 1502:
```suggestion
            The ``icon``, ``banner``, ``splash``, ``discovery_splash``
            parameters now accept :class:`File`, :class:`Attachment`, and :class:`Asset`.
```

nextcord/guild.py
```diff
@@ -2472,13 +2469,16 @@ async def create_custom_emoji(
         You must have the :attr:`~Permissions.manage_emojis` permission to
         do this.
 
+        .. versionchanged:: 2.1
+            The :attr:`image` attribute now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```
ooliver1, line 2473:
```suggestion
            The ``image`` parameter now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```

nextcord/guild.py
```diff
@@ -2621,6 +2620,9 @@ async def create_role(
         .. versionchanged:: 1.6
             Can now pass ``int`` to ``colour`` keyword-only parameter.
 
+        .. versionchanged:: 2.1
+            The :attr:`icon` attribute now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```
ooliver1, line 2624:
```suggestion
            The ``icon`` parameter now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```

nextcord/guild.py
```diff
@@ -3269,13 +3266,16 @@ async def create_scheduled_event(
...
-        image: bytes = MISSING,
+        image: Optional[Union[bytes, Asset, Attachment, File]] = None,
         reason: Optional[str] = None,
     ) -> ScheduledEvent:
         """|coro|
 
         Create a new scheduled event object.
 
+        .. versionchanged:: 2.1
+        The :attr:`image` attribute now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```
ooliver1, line 3277:
```suggestion
        The ``image`` parameter now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```

nextcord/guild.py
```diff
@@ -3294,18 +3294,22 @@ async def create_scheduled_event(
             The description for the event
         entity_type: :class:`ScheduledEventEntityType`
             The type of event
-        image: :class:`bytes`
-            A :term:`py:bytes-like object` representing the cover image.
+        image: Optional[Union[:class:`bytes`, :class:`Asset`, :class:`Attachment`, :class:`File`]]
+            A :term:`py:bytes-like object`, :class:`File`, :class:`Attachment`, or :class:`Asset`
+            representing the cover image.
+        reason: Optional[:class:`str`]
+            The reason for creating this scheduled_event. Shows up in the audit logs.
```
ooliver1, line 3301:
```suggestion
            The reason for creating this scheduled event. Shows up in the audit logs.
```

nextcord/role.py
```diff
@@ -394,6 +395,9 @@ async def edit(
         .. versionchanged:: 2.0
             Edits are no longer in-place, the newly edited role is returned instead.
 
+        .. versionchanged:: 2.1
+            The :attr:`icon` attribute now accepts :class:`Attachment`, and :class:`Asset`.
```
ooliver1, line 399:
```suggestion
            The ``icon`` parameter now accepts :class:`Attachment`, and :class:`Asset`.
```

nextcord/scheduled_events.py
```diff
@@ -333,13 +335,16 @@ async def edit(
...
-        image: Optional[bytes] = MISSING,
+        reason: Optional[str] = None,
+        image: Optional[Union[bytes, Asset, Attachment, File]] = MISSING,
     ) -> ScheduledEvent:
         """|coro|
 
         Edit the scheduled event.
 
+        .. versionchanged:: 2.1
+            The :attr:`image` attribute now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```
ooliver1, line 346:
```suggestion
            The ``image`` parameter now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```

nextcord/scheduled_events.py
```diff
@@ -360,23 +365,25 @@ async def edit(
             The new type for the event.
         status: :class:`ScheduledEventStatus`
             The new status for the event.
+        reason: Optional[:class:`str`]
+            The reason for editing this scheduled_event. Shows up in the audit logs.
```
ooliver1, line 369:
```suggestion
            The reason for editing this scheduled event. Shows up in the audit logs.
```

nextcord/template.py
```diff
@@ -169,24 +172,30 @@ def __repr__(self) -> str:
...
+        icon: Optional[Union[bytes, Asset, Attachment, File]] = None,
     ) -> Guild:
         """|coro|
 
         Creates a :class:`.Guild` using the template.
 
         Bot accounts in more than 10 guilds are not allowed to create guilds.
 
+        .. versionchanged:: 2.1
+        The :attr:`icon` attribute now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```
ooliver1, line 187:
```suggestion
        .. versionchanged:: 2.1
            The :attr:`icon` attribute now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```

nextcord/user.py
```diff
@@ -363,13 +372,16 @@ async def edit(self, *, username: str = MISSING, avatar: bytes = MISSING) -> Cli
         .. versionchanged:: 2.0
             The edit is no longer in-place, instead the newly edited client user is returned.
 
+        .. versionchanged:: 2.1
+            The :attr:`avatar` attribute now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```
ooliver1, line 376:
```suggestion
        .. versionchanged:: 2.1
            The ``avatar`` parameter now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```

nextcord/webhook/async_.py
```diff
@@ -1245,20 +1245,24 @@ async def edit(
...
+        avatar: Optional[Union[bytes, Asset, Attachment, File]] = MISSING,
         channel: Optional[Snowflake] = None,
         prefer_auth: bool = True,
     ) -> Webhook:
         """|coro|
 
         Edits this Webhook.
 
+        .. versionchanged:: 2.1
+            The :attr:`avatar` attribute now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```
ooliver1, line 1257:
```suggestion
            The ``avatar`` parameter now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```

nextcord/webhook/sync.py
```diff
@@ -743,18 +743,21 @@ def edit(
...
         name: Optional[str] = MISSING,
-        avatar: Optional[bytes] = MISSING,
+        avatar: Optional[Union[bytes, File]] = MISSING,
         channel: Optional[Snowflake] = None,
         prefer_auth: bool = True,
     ) -> SyncWebhook:
         """Edits this Webhook.
 
+        .. versionchanged:: 2.1
+            The :attr:`avatar` attribute now accepts :class:`File`.
```
ooliver1, line 753:
```suggestion
            The ``avatar`` parameter now accepts :class:`File`.
```

ooliver1 (changes requested):
A few docs changes, otherwise looks good. Feel free to ping me in Discord once these changes have been made.

nextcord/guild.py
```diff
@@ -3269,13 +3266,16 @@ async def create_scheduled_event(
...
-        image: bytes = MISSING,
+        image: Optional[Union[bytes, Asset, Attachment, File]] = None,
         reason: Optional[str] = None,
     ) -> ScheduledEvent:
         """|coro|
 
         Create a new scheduled event object.
 
+        .. versionchanged:: 2.1
+        The ``image`` parameter now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```
ooliver1, line 3277:
```suggestion
        .. versionchanged:: 2.1
            The ``image`` parameter now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```

nextcord/template.py
```diff
@@ -169,24 +172,30 @@ def __repr__(self) -> str:
...
+        icon: Optional[Union[bytes, Asset, Attachment, File]] = None,
     ) -> Guild:
         """|coro|
 
         Creates a :class:`.Guild` using the template.
 
         Bot accounts in more than 10 guilds are not allowed to create guilds.
 
+        .. versionchanged:: 2.1
+            The :attr:`icon` attribute now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```
ooliver1, line 187:
```suggestion
            The ``icon`` parameter now accepts :class:`File`, :class:`Attachment`, and :class:`Asset`.
```

ooliver1 (changes requested):

ooliver1 (approved):

PR #620, opened 2022-04-25
feat(VocalGuildChannel): add `VocalGuildChannel.members_count`

MaskDuck:
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/channel.py
+++ b/nextcord/channel.py
@@ -836,6 +836,11 @@ class VocalGuildChannel(abc.Connectable, abc.GuildChannel, Hashable):
                 if member is not None:
                     ret.append(member)
         return ret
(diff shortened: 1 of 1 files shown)


koala9712:
What's a `VocalGuildChannel` ?
What is the aim of this PR?

The function right above where you edited also supposedly returns the same thing:
"""List[:class:`Member`]: Returns all members that are currently inside this voice channel."""

It's like making an extra property for len(Channel.members) - like making an extra property for len(guild.members)

ooliver1:
1. `member_count` not `members_count`
2. `Guild.member_count` exists because that is provided by discord - is used especially when members intent is not enabled

PR #621, opened 2022-04-25
fix(Message): RawMessage edits throwing key errors

veni-vidi-code:
Fixes Error thrown for Raw Message edits due to attachemnts key missing

The possibility of it missing is also mentioned in the docs https://discord.com/developers/docs/topics/gateway#message-update

I am getting this error regulary since a former release so i wanted to fix it:
```python
"/venv/lib/python3.10/site-packages/nextcord/message.py", line 663, in __init__
    self.attachments: List[Attachment] = [Attachment(data=a, state=self._state) for a in data['attachments']]
KeyError: 'attachments'
```

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/message.py
+++ b/nextcord/message.py
@@ -686,8 +686,9 @@ class Message(Hashable):
         self.id: int = int(data['id'])
         self.webhook_id: Optional[int] = utils._get_as_snowflake(data, 'webhook_id')
         self.reactions: List[Reaction] = [Reaction(message=self, data=d) for d in data.get('reactions', [])]
-        self.attachments: List[Attachment] = \
-            [Attachment(data=a, state=self._state) for a in data.get('attachments', [])]
+        self.attachments: List[Attachment] = [
+            Attachment(data=a, state=self._state) for a in data.get('attachments', [])
+        ]
         self.embeds: List[Embed] = [Embed.from_dict(a) for a in data.get('embeds', [])]
         self.application: Optional[MessageApplicationPayload] = data.get('application')
         self.activity: Optional[MessageActivityPayload] = data.get('activity')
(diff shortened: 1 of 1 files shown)


nextcord/message.py
```diff
@@ -686,8 +686,9 @@ def __init__(
         self.id: int = int(data['id'])
         self.webhook_id: Optional[int] = utils._get_as_snowflake(data, 'webhook_id')
         self.reactions: List[Reaction] = [Reaction(message=self, data=d) for d in data.get('reactions', [])]
-        self.attachments: List[Attachment] = [Attachment(data=a, state=self._state) for a in data['attachments']]
-        self.embeds: List[Embed] = [Embed.from_dict(a) for a in data['embeds']]
+        self.attachments: List[Attachment] = \
+            [Attachment(data=a, state=self._state) for a in data.get('attachments', [])]
+        self.embeds: List[Embed] = [Embed.from_dict(a) for a in data.get('embeds', [])]
```
DenverCoder1, line 691: Formatting suggestion

```suggestion
        self.attachments: List[Attachment] = [
            Attachment(data=a, state=self._state) for a in data.get("attachments", [])
        ]
        self.embeds: List[Embed] = [Embed.from_dict(a) for a in data.get("embeds", [])]
```

DenverCoder1 (approved):
As the docs say that everything except id and channel id may be left out, maybe there should be defaults for everything else as well (eg. type, pinned, content, etc.) just in case? Although it's unclear when that could happen.

Skelmis (approved):

DenverCoder1:
Maybe default to `None`, or maybe just ignore it since there doesn't seem to be any issue for now? It could possibly be documented somewhere with a warning.

DenverCoder1:
Doing `msg.channel.get_message(msg.id)` / `state._get_message(msg.id)`? could possibly solve the issue as that would be the most up-to-date version of the message the cache is aware of.

DenverCoder1:
The combined information could be used I suppose, if you take the older message and update it with the raw data.

veni-vidi-code:
Ig that would be another solution... Actually i think it was a problem on my side, it seems as if the libary doesnt do the creation of the message itself, just me... Sry

PR #626, opened 2022-04-27
feat(Embed): return the new embed at `Embed.clear_fields()`

### LINKED ISSUE #624: Return embed for `Embed.clear_fields()`
### Summary

`clear_fields()` should return the modified `Embed` instance to allow for fluent-style chaining

The core library

### The Problem

Almost all other methods of `Embed` return the embed itself after modifying it, but `clear_fields()` doesn't. It doesn't matter too much, but considering all the others do, it's a small inconsistency that should be fixed.

### The Ideal Solution

It would allow for more compact usage, like so:

```py
async def relay(interaction, embed: nextcord.Embed):
  await interaction.send(embed = copy.deepcopy(embed).clear_fields())
```

### The Current Solution

_No response_

### Additional Context

_No response_

MaskDuck:
## Summary

close #624

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/embeds.py
+++ b/nextcord/embeds.py
@@ -618,10 +618,12 @@ class Embed:
 
     def clear_fields(self) -> Embed:
         """Removes all fields from this embed.
-        
-        Returns
-        --------
-        :class:`Embed`"""
+
+        This function returns the class instance to allow for fluent-style
+        chaining.
+
+        .. versionchanged:: 2.0
+            This method returns ``self``
         try:
             self._fields.clear()
         except AttributeError:
--- a/nextcord/embeds.py
+++ b/nextcord/embeds.py
@@ -616,8 +616,12 @@ class Embed:
 
         return self
 
-    def clear_fields(self) -> None:
-        """Removes all fields from this embed."""
+    def clear_fields(self) -> Embed:
+        """Removes all fields from this embed.
+        
+        Returns
+        --------
+        :class:`Embed`"""
         try:
             self._fields.clear()
         except AttributeError:
--- a/nextcord/embeds.py
+++ b/nextcord/embeds.py
@@ -623,6 +623,8 @@ class Embed:
         except AttributeError:
             self._fields = []
 
+        return self
+
     def remove_field(self, index: int) -> None:
         """Removes a field at a specified index.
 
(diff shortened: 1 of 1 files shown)


koala9712:
Doc can be updated to show return type.

DenverCoder1 (changes requested):
Might as well do the same for `remove_field`?

nextcord/embeds.py
```diff
@@ -616,13 +616,19 @@ def insert_field_at(self: E, index: int, *, name: Any, value: Any, inline: bool
 
         return self
 
-    def clear_fields(self) -> None:
-        """Removes all fields from this embed."""
+    def clear_fields(self) -> Embed:
```
DenverCoder1, line 619: For consistency

```suggestion
    def clear_fields(self: E) -> E:
```

nextcord/embeds.py
```diff
@@ -616,13 +616,21 @@ def insert_field_at(self: E, index: int, *, name: Any, value: Any, inline: bool
...
-    def clear_fields(self) -> None:
-        """Removes all fields from this embed."""
+    def clear_fields(self: E) -> E:
+        """Removes all fields from this embed.
+
+        This function returns the class instance to allow for fluent-style
+        chaining.
+
+        .. versionchanged:: 2.0
+            This method returns ``self``
```
DenverCoder1, line 626: I actually might change my mind about the versionchanged. Since it's not a breaking change, I don't know if it's necessary.
```suggestion
```

nextcord/embeds.py
```diff
@@ -616,13 +616,18 @@ def insert_field_at(self: E, index: int, *, name: Any, value: Any, inline: bool
 
         return self
 
-    def clear_fields(self) -> None:
-        """Removes all fields from this embed."""
+    def clear_fields(self: E) -> E:
+        """Removes all fields from this embed.
+
+        This function returns the class instance to allow for fluent-style
+        chaining.
```
DenverCoder1, line 623: Small typo missing end quotes.

```suggestion
        chaining.
        """
```

Also, `remove_field` could return self as well maybe?

ooliver1 (commented):
what happened to the return type

DenverCoder1:
> what happened to the return type

All of the existing methods in this lib that follow the builder design pattern seem to just have a comment saying:

"This function returns the class instance to allow for fluent-style chaining."

and they are not documented with "Returns".

ooliver1:
oh i see i see

ooliver1 (approved):

PR #637, opened 2022-05-06
Added Member.disconnect function - fixes #636

### LINKED ISSUE #636: Method to disconnect a member from the VC.
### Summary

member.disconnect() method to disconnect the member from VC

The core library

### The Problem

As of right now, you have to do `await member.move_to(None)` to disconnect a member from their voice channel. I believe this could very well use an own method `await member.disconnect()` for example, just for the sake of simplicity.

### The Ideal Solution

Create a method `.disconnect()` that disconnects a member from their vc

### The Current Solution

Do it manually using `await member.move_to(None)`

### Additional Context

https://discord.com/channels/881118111967883295/972136018012696626

ToasterUwU:
## Summary

This adds a disconnect function to the Member class so that move_to(None) is not the only option anymore.
This fixes #636

## Checklist

- [X] This PR fixes an issue.


nextcord/member.py
```diff
@@ -887,6 +887,23 @@ async def move_to(self, channel: Optional[VocalGuildChannel], *, reason: Optiona
...
 
+    async def disconnect(self, *, reason: Optional[str] = None) -> None:
+        """|coro|
+
+        Disconnects a member from the voice channel they are connected to.
+
+        You must have the :attr:`~Permissions.move_members` permission to
+        use this.
+
+        This raises the same exceptions as :meth:`edit`.
```
Skelmis, line 898: I feel it would be better to have these here regardless

ToasterUwU: > I feel it would be better to have these here regardless

have what here regardless?

ooliver1: the errors listed

ToasterUwU: i just copy pasted the line from the move_to thing, so if its not specific enough here, it shouldn't be at move_to either.

Skelmis: Oki doki, feel free to resolve

Skelmis (approved):

ooliver1 (approved):

PR #638, opened 2022-05-07
Changed str to int in the SlashOption docs

E0lNt3xbMLCY45pNScOMb2qR1zl5fTcKljZMgu2:
This code will not run with str, however will with int.

## Summary

Allows the sample code in the documentation to run in its current form. 
## Checklist

- [ ] This PR fixes an issue.
- [x ] This PR is **not** a code change (e.g. documentation, README, ...)

### DIFF
--- a/docs/interactions.rst
+++ b/docs/interactions.rst
@@ -70,7 +70,7 @@ Nextcord's implementation of slash commands has fields and is very simple. in th
     @bot.slash_command()
     async def choose_a_number(
         interaction: Interaction,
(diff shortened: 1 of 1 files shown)


ooliver1 (approved):

alentoghostflame:
IMO, it should be changed to be a list of numbers. (IIRC lists weren't supported when this was added)

DenverCoder1:
The branch you are targeting is not up to date.

It has already been fixed in the master branch for the next release.

https://github.com/nextcord/nextcord/blob/master/docs/interactions.rst

ooliver1:
I should really pay attention to the target, thanks

PR #649, opened 2022-05-17
fix: guild iterator pagination + limits

### LINKED ISSUE #648: nextcord.ext.commands.Bot.fetch_guilds - "limit=None" doesnt work. Limit still defaults to 100
### Summary

bot.fetch_guilds(limit=None) returns not all, but only 100 guilds

### Reproduction Steps

Just look at the Code example.

### Minimal Reproducible Code

```python

bot = commands.Bot()

@bot.event
async def on_ready():
    guilds = await bot.fetch_guilds(limit=None).flatten() # Will return up to 100 guilds
    guilds = await bot.fetch_guilds(limit=150).flatten() # Will return up to 150 guilds

bot.run("TOKEN")
```

### Expected Results

"limit=None" should return all guilds → `limit: Optional[int]
    The number of guilds to retrieve. If None, it retrieves every guild you have access to. Note, however, that this would make it a slow operation. Defaults to 100.`

### Actual Results

"limit=None" returns only 100 guilds

### Intents

default, thats it.

### System Information
```
- Python v3.8.10-final
- nextcord v2.0.0-alpha
    - nextcord pkg_resources: v2.0.0a10
- aiohttp v3.8.0
- system info: Linux 5.4.0-110-generic #124-Ubuntu SMP Thu Apr 14 19:46:19 UTC 2022
```

### Checklist

### Additional Context

_No response_

shiftinv:
## Summary

Updates the default limit and page size to 200 ([docs](https://discord.com/developers/docs/resources/user#get-current-user-guilds)), and fixes several issues in `GuildIterator` that could break pagination:
- Only use `before` strategy if `before` parameter was specified
- Don't stop after first page if `limit=None` is set
- Always return results in correct order

Resolves #648

<sup>related: DisnakeDev/disnake#526</sup>

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/client.py
+++ b/nextcord/client.py
@@ -1257,7 +1257,7 @@ class Client:
     def fetch_guilds(
         self,
         *,
-        limit: Optional[int] = 100,
+        limit: Optional[int] = 200,
         before: Optional[SnowflakeTime] = None,
         after: Optional[SnowflakeTime] = None,
     ) -> GuildIterator:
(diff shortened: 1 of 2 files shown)


ooliver1 (approved):

ToasterUwU (approved):
Seems to work perfectly

PR #652, opened 2022-05-18
fix: getting guild when no channel type

### LINKED ISSUE #651: Client.get_partial_messageable().send() is not sending a message
### Summary

The PartialMessageable object send method does not send message.

### Reproduction Steps

Steps that I did:
1. partial_channel = Client.get_partial_messageable(my dm_channel ID, type = ChannelType.private)
2. await partial_channel.send(embed = some embed)

In the source code, this looks like
1. `bot.get_partial_messageable` creates a `PartialMessageable` object
2. The `PartialMessageable` has an attribute channel which is initialized as `Object(id = id)`
3. `PartialMessageable.send` calls `ConnectionState.create_message(channel, data)` where channel is the Object and data is stuff from `await state.http.send_message`
4. `ConnectionState.create_message` creates a new Message object with `Message(state, channel, data)`
5. `Message.__init__` attempts to get the guild attribute and type attribute from channel (which is an Object object).

Traceback can be found here: https://paste.nextcord.dev/?id=1652901053415284

### Minimal Reproducible Code

```python
bot = commands.Bot()
user = bot.get_user(MY ID)
dm_channel = user.dm_channel if user.dm_channel else await user.create_dm()
partial_channel = bot.get_partial_messageable(dm_channel.id, type = ChannelType.private)
await partial_channel.send(embed = some embed)
```

### Expected Results

I expected the message to be sent with the `PartialMessageable` object.

### Actual Results

The `Message.__init__` attempts to check the guild attribute and type attribute of the "channel" object, but `channel` is an Object object.

### Intents

nextcord.Intents.default() with intent.members = True

### System Information

- Python v3.10.1-final
- nextcord v2.0.0-alpha
    - nextcord pkg_resources: v2.0.0a10
- aiohttp v3.8.1
- system info: Windows 10 10.0.19043

### Checklist

- [ ] I have searched the open issues for duplicates.

### Additional Context

_No response_

DenverCoder1:
## Summary

Fixes #651

An attribute occurs when the channel used to create a message is partial (i.e. an Object with an id, but without other attributes such as type). This can be caused by the lib, for example, using `PartialMessageable.send()`:

```py
  File "nextcord/ext/commands/core.py", line 165, in wrapped
    ret = await coro(*args, **kwargs)
  File "test.py", line 33, in test
    ret = await partial_channel.send("Hello World")
  File "nextcord/dist/../nextcord/abc.py", line 1465, in send
    ret = state.create_message(channel=channel, data=data)
  File "nextcord/dist/../nextcord/state.py", line 1957, in create_message
    return Message(state=self, channel=channel, data=data)
  File "nextcord/dist/../nextcord/message.py", line 750, in __init__
    if channel.type is not ChannelType.private and channel.type is not ChannelType.group:
AttributeError: 'Object' object has no attribute 'type'
```

To fix this, `getattr(channel, "type", None)` can be used instead of `channel.type`.

### Testing (code from issue #651)

```py
@bot.command(aliases=["t"])
async def test(ctx):
    user = ctx.author
    dm_channel = user.dm_channel if user.dm_channel else await user.create_dm()
    partial_channel = bot.get_partial_messageable(dm_channel.id, type=nextcord.ChannelType.private)
    ret = await partial_channel.send("Hello World")
    await ctx.send(f"Sent message {ret}.\n\nGuild ID: {ret.guild}")
```

image

### Additional info

Prior to #461, this if-else structure was simply:

```py
self.guild = state._get_guild(utils._get_as_snowflake(data, "guild_id"))
```

While this works and does not cause any issues, it was changed in #461 to bring a small performance improvement.

I have changed it to use `getattr()` rather than reverting to the simpler approach in order to keep the "performance improvement" mentioned in the above mentioned PR.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/message.py
+++ b/nextcord/message.py
@@ -747,7 +747,10 @@ class Message(Hashable):
             # if the channel doesn't have a guild attribute, we handle that
             self.guild = channel.guild  # type: ignore
         except AttributeError:
-            self.guild = state._get_guild(utils._get_as_snowflake(data, "guild_id"))
+            if getattr(channel, "type", None) in (ChannelType.group, ChannelType.private):
+                self.guild = state._get_guild(utils._get_as_snowflake(data, "guild_id"))
+            else:
+                self.guild = None
 
         if thread_data := data.get("thread"):
             if not self.thread and self.guild:
--- a/nextcord/message.py
+++ b/nextcord/message.py
@@ -747,7 +747,7 @@ class Message(Hashable):
             # if the channel doesn't have a guild attribute, we handle that
             self.guild = channel.guild  # type: ignore
         except AttributeError:
-            if getattr(channel, "type", None) in (ChannelType.group, ChannelType.private):
+            if getattr(channel, "type", None) not in (ChannelType.group, ChannelType.private):
                 self.guild = state._get_guild(utils._get_as_snowflake(data, "guild_id"))
             else:
                 self.guild = None
--- a/nextcord/message.py
+++ b/nextcord/message.py
@@ -747,10 +747,7 @@ class Message(Hashable):
             # if the channel doesn't have a guild attribute, we handle that
             self.guild = channel.guild  # type: ignore
         except AttributeError:
-            if channel.type is not ChannelType.private and channel.type is not ChannelType.group:
-                self.guild = state._get_guild(utils._get_as_snowflake(data, "guild_id"))
-            else:
-                self.guild = None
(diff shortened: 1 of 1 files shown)


DenverCoder1:
Note: this effectively reverts #461. Possibly another solution can be made.

Skelmis (approved):

ooliver1 (approved):

PR #653, opened 2022-06-01
feat: Add the ability to set local images in embeds

Skelmis:
## Summary

This pull request adds a simplistic interface for attaching local images to embeds. 

Test code can be found [here](https://paste.nextcord.dev/?id=1654098173570265) and all cases in the code work as expected

## Checklist

  - [X] I have run `task pyright` and fixed the relevant issues. (i think)
- [ ] This PR fixes an issue.


nextcord/embeds.py
```diff
@@ -215,6 +222,9 @@ def __init__(
         if timestamp:
             self.timestamp = timestamp
 
+        self._local_image: Optional[File] = None
```
ooliver1, line 225: type issue here is caused by
https://github.com/nextcord/nextcord/pull/653/files#diff-223279fd9bce07b3afa63f9f5d80fdaf27d8bc90b6fa894cf4fdaeca7037a49aR484

nextcord/embeds.py
```diff
@@ -215,6 +222,9 @@ def __init__(
...
+        self._local_thumbnail: Optional[File] = None
```
ooliver1, line 226: type issue here is caused by
https://github.com/nextcord/nextcord/pull/653/files#diff-223279fd9bce07b3afa63f9f5d80fdaf27d8bc90b6fa894cf4fdaeca7037a49aR560
potential fix: do not re-declare the type in that method

nextcord/embeds.py
```diff
@@ -434,13 +444,47 @@ def set_image(self: E, url: MaybeEmpty[Any]) -> E:
...
 
         return self
 
+    def set_local_image(
+        self: E,
+        image_path: Union[str, bytes, os.PathLike, io.BufferedIOBase],
+        *,
+        image_name: Optional[str] = None,
+    ) -> E:
+        """
```
ooliver1, line 462: could these docstrings follow pep257 of
```py
"""short desc

long desc
"""
``` 
instead of the current
```py
"""
short

long
"""
```

ooliver1 (changes requested):

nextcord/abc.py
```diff
@@ -1366,10 +1366,20 @@ async def send(
...
+            if embed._local_thumbnail:
+                local_embed_files.append(embed._local_thumbnail)
+
             embed = embed.to_dict()
 
         elif embeds is not None:
+            local_embed_files.extend([e._local_image for e in embeds if bool(e._local_image)])
+            local_embed_files.extend(
+                [e._local_thumbnail for e in embeds if bool(e._local_thumbnail)]
+            )
```
DenverCoder1, line 1382: Is the cast to bool necessary?
```suggestion
            local_embed_files.extend([e._local_image for e in embeds if e._local_image])
            local_embed_files.extend(
                [e._local_thumbnail for e in embeds if e._local_thumbnail]
            )
```

nextcord/abc.py
```diff
@@ -1366,10 +1366,20 @@ async def send(
         if embed is not None and embeds is not None:
             raise InvalidArgument("Cannot pass both embed and embeds parameter to send()")
 
+        local_embed_files = []
         if embed is not None:
+            if embed._local_image:
```
DenverCoder1, line 1371: A lot of this logic can be simplified if we refactor the code to do:

`embeds = [embed]` and `files = [file]` early on (when `embed` or `file` are set, respectively.

The cases for single embed and single file can then be removed since the list case can cover both.

It would be similar to what is done in `interactions.py`.

nextcord/embeds.py
```diff
@@ -434,13 +444,46 @@ def set_image(self: E, url: MaybeEmpty[Any]) -> E:
...
                 "url": str(url),
             }
 
         return self
 
+    def set_local_image(
+        self: E,
+        image_path: Union[str, bytes, os.PathLike, io.BufferedIOBase],
+        *,
+        image_name: Optional[str] = None,
```
DenverCoder1, line 460: Instead of both of these, why not make a single parameter of type `File` similar to other places.

Also related: #607 (can make it take `Union[bytes, Asset, Attachment, File]`)

nextcord/embeds.py
```diff
@@ -360,7 +368,11 @@ def footer(self) -> _EmbedFooterProxy:
         return EmbedProxy(getattr(self, "_footer", {}))  # type: ignore
 
     def set_footer(
-        self: E, *, text: MaybeEmpty[Any] = EmptyEmbed, icon_url: MaybeEmpty[Any] = EmptyEmbed
+        self: E,
+        *,
+        file: Optional[File] = None,
+        text: MaybeEmpty[Any] = EmptyEmbed,
+        icon_url: MaybeEmpty[Any] = EmptyEmbed,
```
DenverCoder1, line 375: Perhaps EmptyEmbed should be used for consistency, and maybe name it icon_file to match with icon_url?

```suggestion
        text: MaybeEmpty[Any] = EmptyEmbed,
        icon_url: MaybeEmpty[Any] = EmptyEmbed,
        icon_file: MaybeEmpty[File] = EmptyEmbed,
```

nextcord/embeds.py
```diff
@@ -427,17 +455,79 @@ def set_image(self: E, url: MaybeEmpty[Any]) -> E:
...
             The source URL for the image. Only HTTP(S) is supported.
+        file: Optional[File]
+            A file to use for the image.
+
+            .. versionadded:: 2.0
         """
+        from . import InvalidArgument
 
-        if url is EmptyEmbed:
+        if url is EmptyEmbed and not file:
```
DenverCoder1, line 465: This might need to be a separate if-block, since you can set the image to a url after setting it to a file for example, in which case it should delete only the file one.

nextcord/embeds.py
```diff
@@ -427,17 +455,79 @@ def set_image(self: E, url: MaybeEmpty[Any]) -> E:
...
+            The URL of the author icon. Only HTTP(S) is supported.
+        file: Optional[File]
+            A file to use for the image.
+
+            .. versionadded:: 2.0
+        """
+        from . import InvalidArgument
+
+        self._author = {
+            "name": str(name),
```
DenverCoder1, line 516: Can probably fit this on 1 line
```suggestion
            "name": str(name)
```

nextcord/embeds.py
```diff
@@ -456,7 +546,12 @@ def thumbnail(self) -> _EmbedMediaProxy:
         """
         return EmbedProxy(getattr(self, "_thumbnail", {}))  # type: ignore
 
-    def set_thumbnail(self: E, url: MaybeEmpty[Any]) -> E:
+    def set_thumbnail(
+        self: E,
+        url: MaybeEmpty[Any] = EmptyEmbed,
+        *,
+        file: Optional[File] = None,
```
DenverCoder1, line 553: I want to say these should both be keyword arguments.

It's a tough call since that will be a breaking change, as it would mean you would no longer be able to make the url positional.

nextcord/embeds.py
```diff
@@ -469,17 +564,32 @@ def set_thumbnail(self: E, url: MaybeEmpty[Any]) -> E:
...
             try:
                 del self._thumbnail
             except AttributeError:
                 pass
+            try:
+                del self._local_files["thumbnail"]
+            except:
+                pass
+
+        elif url and file:
```
DenverCoder1, line 584: Should probably be using `is not EmptyEmbed`

nextcord/embeds.py
Skelmis: I was thinking `file` for consistency with the other 3 messages

nextcord/embeds.py
```diff
@@ -414,7 +438,11 @@ def image(self) -> _EmbedMediaProxy:
         """
         return EmbedProxy(getattr(self, "_image", {}))  # type: ignore
 
-    def set_image(self: E, url: MaybeEmpty[Any]) -> E:
+    def set_image(
+        self: E,
+        url: MaybeEmpty[Any] = None,
+        file: Optional[File] = None,
```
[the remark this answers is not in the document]
Skelmis: Agreed it needs a refactor, but this isn't the pr.

nextcord/embeds.py
DenverCoder1: In the other places there is `url` and `file`, since this has `icon_url` instead, then `icon_file` would match. Naming it wirh icon makes it more clear what the "footer file" / "author file" is for.

nextcord/embeds.py
Skelmis: Closing as I haven't seen anything further on this + it maintains backwards compat + the note above

ooliver1: this *would* be a breaking change yet i would say this is good

nextcord/embeds.py
```diff
@@ -427,17 +456,79 @@ def set_image(self: E, url: MaybeEmpty[Any]) -> E:
...
             except AttributeError:
                 pass
+
+        if file is EmptyEmbed:
+            try:
+                del self._local_files["image"]
+            except:
+                pass
+
+        elif url is not EmptyEmbed and file is not EmptyEmbed:
```
EmreTech, line 478: This `elif` doesn't look right. Maybe change it to an `if`?

EmreTech (commented):
Looks fine otherwise. Also you should resolve the Pyright errors.

nextcord/embeds.py
```diff
@@ -427,17 +456,79 @@ def set_image(self: E, url: MaybeEmpty[Any]) -> E:
...
 
         if url is EmptyEmbed:
             try:
                 del self._image
             except AttributeError:
                 pass
+
+        if file is EmptyEmbed:
+            try:
+                del self._local_files["image"]
```
ooliver1, line 474: this might be worth it to use [`dict.pop(item, default)`](https://docs.python.org/3/library/stdtypes.html#dict.pop) where `default` is `None` so a `ValueError` is not raised

nextcord/embeds.py
```diff
@@ -427,17 +456,79 @@ def set_image(self: E, url: MaybeEmpty[Any]) -> E:
...
-            }
+            self._image = {"url": str(url)}
+
+        return self
+
+    def set_author(
+        self: E,
+        *,
+        name: Any,
+        url: MaybeEmpty[Any] = EmptyEmbed,
```
ooliver1, line 494: shouldnt this be `MaybeEmpty[str]` as urls are strings

ooliver1 (changes requested):

nextcord/embeds.py
Skelmis: Outside the scope of this pr. Just accidentally moved things which is why its green

nextcord/embeds.py
```diff
@@ -373,15 +385,27 @@ def set_footer(
             The footer text.
         icon_url: :class:`str`
             The URL of the footer icon. Only HTTP(S) is supported.
+        icon_file: Optional[File]
```
ooliver1, line 388:
```suggestion
        icon_file: Optional[:class:`File`]
```

nextcord/embeds.py
```diff
@@ -427,17 +456,76 @@ def set_image(self: E, url: MaybeEmpty[Any]) -> E:
         -----------
         url: :class:`str`
             The source URL for the image. Only HTTP(S) is supported.
+        file: Optional[File]
```
ooliver1, line 459:
```suggestion
        file: Optional[:class:`File`]
```

nextcord/embeds.py
```diff
@@ -427,17 +456,76 @@ def set_image(self: E, url: MaybeEmpty[Any]) -> E:
...
+
+        Parameters
+        -----------
+        name: :class:`str`
+            The name of the author.
+        url: :class:`str`
+            The URL for the author.
+        icon_url: :class:`str`
+            The URL of the author icon. Only HTTP(S) is supported.
+        icon_file: Optional[File]
```
ooliver1, line 508:
```suggestion
        icon_file: Optional[:class:`File`]
```

nextcord/embeds.py
```diff
@@ -469,17 +562,31 @@ def set_thumbnail(self: E, url: MaybeEmpty[Any]) -> E:
         -----------
         url: :class:`str`
             The source URL for the thumbnail. Only HTTP(S) is supported.
+        file: Optional[File]
```
ooliver1, line 565:
```suggestion
        file: Optional[:class:`File`]
```

ooliver1 (approved):

nextcord/abc.py
```diff
@@ -1366,10 +1366,14 @@ async def send(
         if embed is not None and embeds is not None:
             raise InvalidArgument("Cannot pass both embed and embeds parameter to send()")
 
+        local_embed_files = []
```
DenverCoder1, line 1369:
```suggestion
        local_embed_files: List[File] = []
```

nextcord/embeds.py
```diff
@@ -42,6 +44,9 @@
 from . import utils
 from .colour import Colour
 
+if TYPE_CHECKING:
+    from . import File, InvalidArgument
```
DenverCoder1, line 48:
```suggestion
    from .errors import InvalidArgument
    from .file import File
```

nextcord/embeds.py
```diff
@@ -373,15 +385,27 @@ def set_footer(
...
+            .. versionadded:: 2.1
         """
+        from . import InvalidArgument
 
         self._footer = {}
         if text is not EmptyEmbed:
             self._footer["text"] = str(text)
 
+        if icon_url is not EmptyEmbed and icon_file is not EmptyEmbed:
+            raise InvalidArgument("Can't pass both icon_url and file.")
```
DenverCoder1, line 400:
```suggestion
            raise InvalidArgument("Cannot pass both icon_url and icon_file")
```

nextcord/embeds.py
```diff
@@ -373,15 +385,27 @@ def set_footer(
             The footer text.
         icon_url: :class:`str`
             The URL of the footer icon. Only HTTP(S) is supported.
+        icon_file: Optional[:class:`File`]
+            A file to use for the image.
+
+            .. versionadded:: 2.1
         """
+        from . import InvalidArgument
```
DenverCoder1, line 393: Can this not be imported at the top?

nextcord/embeds.py
```diff
@@ -373,15 +385,27 @@ def set_footer(
...
         if text is not EmptyEmbed:
             self._footer["text"] = str(text)
 
+        if icon_url is not EmptyEmbed and icon_file is not EmptyEmbed:
+            raise InvalidArgument("Can't pass both icon_url and file.")
+
         if icon_url is not EmptyEmbed:
             self._footer["icon_url"] = str(icon_url)
 
+        if icon_file is not EmptyEmbed:
```
DenverCoder1, line 405: Could check if isinstance of File to avoid type ignore?

nextcord/embeds.py
```diff
@@ -427,17 +456,76 @@ def set_image(self: E, url: MaybeEmpty[Any]) -> E:
...
         if url is EmptyEmbed:
             try:
                 del self._image
             except AttributeError:
                 pass
+
+        if file is EmptyEmbed:
+            self._local_files.pop("image", None)
+
+        elif url is not EmptyEmbed and file is not EmptyEmbed:
```
DenverCoder1, line 475: This is already checked by elif
```suggestion
        elif url is not EmptyEmbed:
```

nextcord/embeds.py
```diff
@@ -427,17 +456,76 @@ def set_image(self: E, url: MaybeEmpty[Any]) -> E:
...
+            A file to use for the image.
+
+            .. versionadded:: 2.1
+        """
+        from . import InvalidArgument
+
+        self._author = {"name": str(name)}
+
+        if icon_url is not EmptyEmbed and icon_file is not EmptyEmbed:
+            raise InvalidArgument("Can't pass both icon_url and file.")
```
DenverCoder1, line 518:
```suggestion
            raise InvalidArgument("Cannot pass both icon_url and icon_file")
```

DenverCoder1 (commented):
Do the `edit` methods need to be changed as well?

nextcord/embeds.py
Skelmis: I'll change it to an `if`, as they are actually meant to be 3 separate logical blocks.
1) Reset existing url image
2) Reset existing file image
3) Set a new image. You can only set 1 image, thus the check

nextcord/embeds.py
Skelmis: Currently not, but I can try find a work around

nextcord/abc.py
```diff
@@ -1366,10 +1366,14 @@ async def send(
...
             raise InvalidArgument("Cannot pass both embed and embeds parameter to send()")
 
+        local_embed_files: List[File] = []
         if embed is not None:
+            local_embed_files = list(embed._local_files.values())
             embed = embed.to_dict()
 
         elif embeds is not None:
+            for embed in embeds:
+                local_embed_files.extend(list(embed._local_files.values()))
```
teaishealthy, line 1376:
```suggestion
                local_embed_files.extend(embed._local_files.values())
```
The `list` can be removed as `list.extend` takes an iterable

nextcord/interactions.py
```diff
@@ -814,7 +814,10 @@ async def send_message(
         if embed is not MISSING:
             embeds = [embed]
 
+        local_embed_files = []
```
teaishealthy, line 817:
```suggestion
        local_embed_files: List[File] = []
```

nextcord/interactions.py
```diff
@@ -814,7 +814,10 @@ async def send_message(
...
         if embeds:
+            for embed in embeds:
+                local_embed_files.extend(list(embed._local_files.values()))
```
teaishealthy, line 820:
```suggestion
                local_embed_files.extend(embed._local_files.values())
```
Same here

teaishealthy (changes requested):

nextcord/abc.py
```diff
@@ -1367,10 +1367,15 @@ async def send(
         if embed is not None and embeds is not None:
             raise InvalidArgument("Cannot pass both embed and embeds parameter to send()")
 
+        local_embed_files: List[File] = []
```
EmreTech, line 1370: After this line, you should set the `embeds` list to a list that only contains `embed` if `embed` is set
This should eliminate the repetitive code below

nextcord/embeds.py
```diff
@@ -517,40 +613,6 @@ def author(self) -> _EmbedAuthorProxy:
         """
         return EmbedProxy(getattr(self, "_author", {}))  # type: ignore
 
-    def set_author(
```
EmreTech, line 520: Isn't this out of scope for this PR?

Skelmis: Not sure what you mean? Outside of accidentally moving it I'm not sure what this comment is referring to

EmreTech: Oh nevermind I didn't spot that it was moved

teaishealthy (approved):

nextcord/embeds.py
```diff
@@ -373,15 +382,25 @@ def set_footer(
...
             self._footer["text"] = str(text)
 
+        if icon_url is not EmptyEmbed and icon_file is not EmptyEmbed:
+            raise InvalidArgument("Cannot pass both icon_url and icon_file")
+
         if icon_url is not EmptyEmbed:
             self._footer["icon_url"] = str(icon_url)
 
+        if isinstance(icon_file, File):
+            self._local_files["footer"] = icon_file
```
DenverCoder1, line 401: footer image file can be removed from local files if icon url is set afterwards similar to other methods?

nextcord/embeds.py
```diff
@@ -427,17 +451,72 @@ def set_image(self: E, url: MaybeEmpty[Any]) -> E:
...
+        self._author = {"name": str(name)}
+
+        if icon_url is not EmptyEmbed and icon_file is not EmptyEmbed:
+            raise InvalidArgument("Cannot pass both icon_url and icon_file")
+
+        if url is not EmptyEmbed:
+            self._author["url"] = str(url)
+
+        if isinstance(icon_file, File):
+            self._local_files["author"] = icon_file
```
DenverCoder1, line 515: Same as above - removing author?

docs/faq.rst
```diff
@@ -314,17 +314,15 @@ How do I use a local image file for an embed image?
...
+To do so, create a file object and pass it as a ``file`` kwarg.
 
 
 Quick example: ::
 
     file = nextcord.File("path/to/my/image.png", filename="image.png")
     embed = nextcord.Embed()
-    embed.set_image(url="attachment://image.png")
-    await channel.send(file=file, embed=embed)
+    embed.set_image(file=file)
```
DenverCoder1, line 324: Maybe set_thumbnail, set_author, and set_footer examples should also be shown?

DenverCoder1 (commented):
The test code in the PR description seems outdated

I tested with this code - https://paste.nextcord.dev/?id=1659541555914935

nextcord/embeds.py
```diff
@@ -360,7 +365,11 @@ def footer(self) -> _EmbedFooterProxy:
         return EmbedProxy(getattr(self, "_footer", {}))  # type: ignore
 
     def set_footer(
-        self: E, *, text: MaybeEmpty[Any] = EmptyEmbed, icon_url: MaybeEmpty[Any] = EmptyEmbed
+        self: E,
+        *,
+        text: MaybeEmpty[Any] = EmptyEmbed,
+        icon_url: MaybeEmpty[Any] = EmptyEmbed,
+        icon_file: MaybeEmpty[File] = EmptyEmbed,
```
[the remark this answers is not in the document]
Skelmis: I don't think so as 607 added support for items where they were cast to bytes originally, however, we directly send the files using `http.send_files` which takes `Sequence[File]`. Admittedly this could be reworked, but I feel it is outside the scope of this pull request. What do you think?

Skelmis:
```python
    embed.set_footer(text="3")
    embed.set_author(name="4")
```
Will remove the images, which is inline with how other methods remove images.

Skelmis:
As much as I would love for this to be in nextcord, I do not currently have the time required to resolve the reviews on this pull request to a standard both I and the maintainers would be happy with.

This may be reopened if I get to a stage where I have found time to make the required changes/test and streamline them.


As noted in discord, all methods where an embed can be sent should be supported. This includes `message.edit / interactionresponse.edit_message`, webhooks, etc..

PR #662, opened 2022-06-05
fix(app-checks): change type aliases to allow for normal funcs

ooliver1:
## Summary

Fixes the application checks to allow for normal `async def` functions before they are decorated as a command.
Also allows for cog versions of the above.

(adding `CoroFunc` to the callable union)

the other partially related change is so app checks can be used on cogs w/o type errors

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/types/checks.py
+++ b/nextcord/types/checks.py
@@ -34,7 +34,9 @@ T = TypeVar("T")
 Coro = Coroutine[Any, Any, T]
 MaybeCoro = Union[T, Coro[T]]
 CoroFunc = Callable[..., Coro[Any]]
-ApplicationCheck = Callable[[Interaction], MaybeCoro[bool]]
+ApplicationCheck = Union[
+    Callable[[ClientCog, Interaction], bool], Callable[[Interaction], MaybeCoro[bool]]
+]
 ApplicationHook = Union[
     Callable[[ClientCog, Interaction], Coro[Any]], Callable[[Interaction], Coro[Any]]
 ]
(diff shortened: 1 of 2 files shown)


nextcord/types/checks.py
```diff
@@ -34,7 +34,9 @@
 Coro = Coroutine[Any, Any, T]
 MaybeCoro = Union[T, Coro[T]]
 CoroFunc = Callable[..., Coro[Any]]
-ApplicationCheck = Callable[[Interaction], MaybeCoro[bool]]
+ApplicationCheck = Union[
+    Callable[[ClientCog, Interaction], bool], Callable[[Interaction], MaybeCoro[bool]]
```
DenverCoder1, line 38: Should the first bool also be in `MaybeCoro`?

ooliver1, line 38:
```suggestion
    Callable[[ClientCog, Interaction], MaybeCoro[bool]], Callable[[Interaction], MaybeCoro[bool]]
```

DenverCoder1 (approved):
Looks good, I haven't tested

nextcord/types/checks.py
```diff
@@ -34,7 +34,9 @@
 Coro = Coroutine[Any, Any, T]
 MaybeCoro = Union[T, Coro[T]]
 CoroFunc = Callable[..., Coro[Any]]
-ApplicationCheck = Callable[[Interaction], MaybeCoro[bool]]
+ApplicationCheck = Union[
+    Callable[[ClientCog, Interaction], MaybeCoro[bool]], Callable[[Interaction], MaybeCoro[bool]]
```
Skelmis, line 38: Do these nested brackets not require a Union?

ooliver1: They represent the arguments to the function

Skelmis: In that case, shouldnt it be `List[..]`?

Skelmis (approved):

PR #663, opened 2022-06-05
remove $ from codeblock

H4ckerxx44:
## Summary

On copying, you would also copy the `$` which I think should be fixed

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -70,9 +70,9 @@ To install the development version, do the following:
 
 .. code:: sh
 
-    $ git clone https://github.com/nextcord/nextcord/
-    $ cd nextcord
-    $ python3 -m pip install -U .[voice]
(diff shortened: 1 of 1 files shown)


toolifelesstocode:
As discussed on the nextcord server, I think it would be best to add more `$`'s to the shell codeblocks, it's best for consistency and generally how the shell usually has those.

ooliver1:
^, `$` signifies you are typing this, no `$` is usually the response from the shell

H4ckerxx44:
> As discussed on the nextcord server, I think it would be best to add more `$`'s to the shell code blocks, it's best for consistency and generally, the shell has those.

I'll close this PR and make a new one for that idea, I think this is the best

PR #656, opened 2022-06-02
chore(readme): add an example slash command

ArjunSharda:
## Summary

_not given_

## Checklist

- [ ] If code changes were made then they have been tested.
  - [ ] I have updated the documentation to reflect the changes.
  - [ ] I have run `task pyright` and fixed the relevant issues.
- [ ] This PR fixes an issue.
- [ ] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [x] This PR is **not** a code change (e.g. documentation, README, ...)

PR #675, opened 2022-06-11
docs: make README not as outdated

toolifelesstocode:
## Summary

This pull request basically makes the README.rst file a bit more updated.

- Uses double quotations(quotation marks) instead of single quotations(apostrophes) in the Quick Example
- Uses the nextcord vanity URL instead of a custom invite
- Uses pypi.org instead of pypi.python.org

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -2,16 +2,16 @@
    :alt: Nextcord
 
 .. image:: https://img.shields.io/discord/881118111967883295?color=blue&label=discord
-   :target: https://discord.gg/ZebatWssCB
+   :target: https://discord.gg/nextcord
    :alt: Discord server invite
 .. image:: https://img.shields.io/pypi/v/nextcord.svg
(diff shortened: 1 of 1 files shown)


DenverCoder1:
I don't know if these links *need* to be changed since the existing ones still work.

Also #656 is touching the Quick Example, so it's probably best to avoid modifying it.

toolifelesstocode:
> I don't know if these links _need_ to be changed since the existing ones still work.

Well, _some_ of the links in the README file do actually use PyPI's new URL, so maybe for consistency, the nextcord URL's could be changed too.

> Also #656 is touching the Quick Example, so it's probably best to avoid modifying it.

Fixed

Skelmis (approved):

ooliver1 (approved):

PR #685, opened 2022-06-16
fix(ext-cmds): copy context before dispatching on_command

### LINKED ISSUE #105: the "is_owner" check used on cog commands is conflicting with "on_command"
### Link

https://github.com/Rapptz/discord.py/issues/1789

ooliver1:
## Summary

fixes #105
copies ctx that is dispatched for `on_command` so a race condition on the race track of `ext.commands` doesnt take over

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/ext/commands/core.py
+++ b/nextcord/ext/commands/core.py
@@ -1177,18 +1177,14 @@ class Command(_BaseCommand, Generic[CogT, P, T]):
                 if local_check is not None:
                     ret = await nextcord.utils.maybe_coroutine(local_check, ctx)
                     if not ret:
-                        ctx.command = original
                         return False
 
             predicates = self.checks
             if not predicates:
                 # since we have no checks, then we just return True.
-                ctx.command = original
                 return True
 
-            result = await nextcord.utils.async_all(predicate(ctx) for predicate in predicates)  # type: ignore
-            ctx.command = original
-            return result
+            return await nextcord.utils.async_all(predicate(ctx) for predicate in predicates)  # type: ignore
         finally:
             ctx.command = original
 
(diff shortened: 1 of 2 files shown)


nextcord/ext/commands/core.py
```diff
@@ -1177,13 +1177,16 @@ async def can_run(self, ctx: Context) -> bool:
...
                         return False
 
             predicates = self.checks
             if not predicates:
                 # since we have no checks, then we just return True.
+                ctx.command = original
                 return True
 
+            ctx.command = original
             return await nextcord.utils.async_all(predicate(ctx) for predicate in predicates)  # type: ignore
```
Skelmis, line 1190: Would this not change the behavior of these predicates? I'm not entirely sure how this functions but it appears like this is a logic change and you were possibly aiming for something more in line with:
```python
result =  await nextcord.utils.async_all(predicate(ctx) for predicate in predicates)  # type: ignore
ctx.command = original
return result
```

Which I think would keep the same behavior, while also resetting the command after its been used by the help command. However, this is an untested theory.

ooliver1: it seems as though this is more complicated than i thought, doing it this way seems to cause the command invoked to run, maybe the command should be `copy.copy()`'d before dispatching so the original one is dispatched?

ooliver1: `copy.copy` and this suggestion has been applied

nextcord/ext/commands/core.py
```diff
@@ -1177,14 +1177,18 @@ async def can_run(self, ctx: Context) -> bool:
                 if local_check is not None:
                     ret = await nextcord.utils.maybe_coroutine(local_check, ctx)
                     if not ret:
+                        ctx.command = original
```
DenverCoder1, line 1180: Doesn't this happen already do to the finally block?

ooliver1: yes it does, `copy.copy` now actually does the work of getting around the race condition

Skelmis (approved):
Untested, lgtm

toolifelesstocode (approved):
Works fine for me! 
image
image

DenverCoder1 (approved):

PR #695, opened 2022-06-21
refactor!(client): remove the use of **options for options

ooliver1:
## Summary

References the opinions of #279
> Discord.py had a prolific use of **kwargs where it is unnecessary. This makes it both hard for new developers and experienced developers alike to get arguments right in what should be as simple as your editor pointing out the arguments.

This is breaking as it removes the ability to use unused options.
This however is better than having this silently ignored, assuming your code does something.

## Checklist

- [ ] This PR fixes an issue.


Skelmis (approved):
Untested, but assuming this works as expected lgtm

PR #696, opened 2022-06-22
refactor!: move cogs to the main library

### LINKED ISSUE #393: Move cogs to Client

EmmmaTech:
## Summary

This PR moves many cog features like names, descriptions, and listeners from ext.commands over to the core library. This also migrates cog functionality from `commands.Bot` to `nextcord.Client`.
However, even though listeners were brought in from `commands.Cog`, it has 0 functionality in `nextcord.Client`. I plan to either add that functionality into this PR or open another PR to add it in.

Closes #393.

## Checklist

- [ ] This PR fixes an issue.


docs/cogs.rst
```diff
@@ -0,0 +1,148 @@
...
+            channel = member.guild.system_channel
+            if channel is not None:
+                await channel.send(f'Welcome {member.mention}.')
+
+        @commands.command()
+        async def hello(self, interaction, member: nextcord.Member = None):
+            """Says hello"""
+            member = member or interaction.user
+            if self._last_member is None or self._last_member.id != member.id:
+                await interaction.response.send_message(f'Hello {member.name}~')
```
ooliver1, line 41:
```suggestion
                await interaction.response.send_message(f"Hello {member.name}~")
```

docs/cogs.rst
```diff
@@ -0,0 +1,148 @@
...
+            else:
+                await interaction.response.send_message(f'Hello {member.name}... This feels familiar.')
```
ooliver1, line 43:
```suggestion
                await interaction.response.send_message(f"Hello {member.name}... This feels familiar.")
```

nextcord/cog.py
```diff
@@ -0,0 +1,324 @@
...
+
+    .. note::
+
+        When passing an attribute of a metaclass that is documented below, note
+        that you must pass it as a keyword-only argument to the class creation
+        like the following example:
+
+        .. code-block:: python3
+
+            class MyCog(nextcord.Cog, name='My Cog'):
```
ooliver1, line 92:
```suggestion
            class MyCog(nextcord.Cog, name="My Cog"):
```

nextcord/cog.py
```diff
@@ -0,0 +1,324 @@
...
+                pass
+
+    Attributes
+    -----------
```
ooliver1, line 96:
```suggestion
    Attributes
    ----------
```

nextcord/cog.py
```diff
@@ -0,0 +1,324 @@
...
+    name: :class:`str`
+        The cog name. By default, it is the name of the class with no modification.
+    description: :class:`str`
+        The cog description. By default, it is the cleaned docstring of the class.
+
+        .. versionadded:: 1.6
```
ooliver1, line 102:
```suggestion
        The cog description. By default, it is the cleaned docstring of the class.
```

nextcord/cog.py
```diff
@@ -0,0 +1,324 @@
...
+
+    @property
+    def listeners(self) -> List[Tuple[str, Callable[..., Any]]]:
+        """Provides the list of listeners in this cog with the name and callback in a tuple."""
+        return [(name, getattr(self, method_name)) for name, method_name in self.__cog_listeners__]
+
+    def get_listeners(self) -> List[Tuple[str, Callable[..., Any]]]:
+        """Returns a :class:`list` of (name, function) listener pairs that are defined in this cog.
+        Returns
+        --------
```
ooliver1, line 227:
```suggestion
        """Returns a :class:`list` of (name, function) listener pairs that are defined in this cog.

        Returns
        -------
```

nextcord/cog.py
```diff
@@ -0,0 +1,324 @@
...
+        """:class:`str`: Returns the cog's description, typically the cleaned docstring."""
+        return self.__cog_description__
+
+    @description.setter
+    def description(self, description: str) -> None:
+        self.__cog_description__ = description
+
+    @property
+    def application_commands(self) -> List[BaseApplicationCommand]:
+        """Provides the list of application commands in this cog. Subcommands are not included."""
```
ooliver1, line 216:
```suggestion
        """List[:class:`BaseApplicationCommand`]: Provides the list of application commands in this cog. Subcommands are not included."""
```

nextcord/cog.py
```diff
@@ -0,0 +1,324 @@
...
+            The listeners defined in this cog.
+        """
+        return [(name, getattr(self, method_name)) for name, method_name in self.__cog_listeners__]
+
+    @classmethod
+    def listener(cls, name: str = MISSING):
+        """A decorator that marks a function as a listener.
+
+        Parameters
+        ------------
```
ooliver1, line 238:
```suggestion
        Parameters
        ----------
```

nextcord/cog.py
```diff
@@ -0,0 +1,324 @@
...
+        name: :class:`str`
+            The name of the event being listened to. If not provided, it
+            defaults to the function's name.
+
+        Raises
+        --------
```
ooliver1, line 244:
```suggestion
        Raises
        ------
```

nextcord/cog.py
```diff
@@ -0,0 +1,324 @@
...
+    @_cog_special_method
+    async def cog_application_command_before_invoke(self, interaction: Interaction) -> None:
+        """A special method that acts as a cog local pre-invoke hook.
+
+        This is similar to :meth:`.ApplicationCommand.before_invoke`.
+
+        This **must** be a coroutine.
+
+        Parameters
+        -----------
```
ooliver1, line 305:
```suggestion
        Parameters
        ----------
```

nextcord/cog.py
```diff
@@ -0,0 +1,324 @@
...
+    @_cog_special_method
+    async def cog_application_command_after_invoke(self, interaction: Interaction) -> None:
+        """A special method that acts as a cog local post-invoke hook.
+
+        This is similar to :meth:`.Command.after_invoke`.
+
+        This **must** be a coroutine.
+
+        Parameters
+        -----------
```
ooliver1, line 320:
```suggestion
        Parameters
        ----------
```

nextcord/ext/commands/cog.py
```diff
@@ -161,27 +112,9 @@ def __new__(cls: Type[CogMeta], *args: Any, **kwargs: Any) -> CogMeta:
...
-            for listener_name in listener.__cog_listener_names__:
-                # I use __name__ instead of just storing the value so I can inject
-                # the self attribute when the time comes to add them to the bot
-                listeners_as_list.append((listener_name, listener.__name__))
-
-        new_cls.__cog_listeners__ = listeners_as_list
-        return new_cls
+
+        new_cls.__cog_commands__ = list(commands.values())  # type: ignore
+        return new_cls  # type: ignore
```
ooliver1, line 117: consider adding comments here explaining why `# type: ignore` is necessary and the type issues cannot be resolved

nextcord/ext/commands/cog.py
```diff
@@ -191,21 +124,18 @@ def qualified_name(cls) -> str:
...
-    When inheriting from this class, the options shown in :class:`CogMeta`
-    are equally valid here.
+    This class inherits from :class:`Cog` and uses the :class:`CommandCogMeta`
+    metaclass. More information on them can be found on the
+    :ref:`ext_commands_cogs` page.
     """
 
     __cog_name__: ClassVar[str]
     __cog_settings__: ClassVar[Dict[str, Any]]
     __cog_commands__: ClassVar[List[Command]]
```
ooliver1, line 137:
```suggestion
    __cog_name__: str
    __cog_settings__: Dict[str, Any]
    __cog_commands__: List[Command]
```
cog name, settings and commands should have this change too

ooliver1 (changes requested):

ooliver1:
I am not too sure on if `commands.Cog` -> `commands.CommandCog` is necessary, both are in separate modules to `nextcord.Cog`, seems like an unnecessary breaking change.

EmreTech:
> I am not too sure on if `commands.Cog` -> `commands.CommandCog` is necessary, both are in separate modules to `nextcord.Cog`, seems like an unnecessary breaking change.

I thought it would be confusing if I had two classes with the same name, but since they are in "different" modules I realize now that it should be fine

docs/cogs.rst
```diff
@@ -0,0 +1,148 @@
...
+                await interaction.response.send_message(f"Hello {member.name}~")
+            else:
+                await interaction.response.send_message(f"Hello {member.name}... This feels familiar.")
+            self._last_member = member
+
+A couple of technical notes to take into consideration:
+
+- All listeners must be explicitly marked via the :meth:`nextcord.Cog.listener` decorator.
+- The name of the cog is automatically derived from the class name but can be overridden. See :ref:`cogs_meta_options`.
+- All commands must now take a ``self`` parameter to allow usage of instance attributes that can be used to maintain state.
```
Skelmis, line 50: Do we want to add the setup function to this as a kind of "full example"?

docs/ext/commands/cogs.rst
```diff
@@ -120,21 +55,16 @@ They are as follows:
...
-At the heart of a cog resides a metaclass, :class:`.commands.CogMeta`, which can take various options to customise some of the behaviour. To do this, we pass keyword arguments to the class definition line. For example, to change the cog name we can pass the ``name`` keyword argument as follows:
-
-.. code-block:: python3
-
-    class MyCog(commands.Cog, name='My Cog'):
-        pass
+Meta options are exactly like what they are in :class:`nextcord.Cog`, however there is an additional option.
 
-To see more options that you can set, see the documentation of :class:`.commands.CogMeta`.
+If you would like to see what this extra option is, see the documentation of :class:`.CogMeta`.
```
DenverCoder1, line 67: I think the wording here could be improved
```suggestion
For more information, see the documentation of :class:`.CogMeta`.
```

docs/api.rst
```diff
@@ -4833,7 +4833,7 @@ Decorators
 Cogs
 ~~~~
 
-.. autoclass:: ClientCog
+.. autoclass:: Cog
```
DenverCoder1, line 4836: Maybe this should stay as ClientCog to avoid breaking changes?

EmreTech: I think an alias would be a better idea
But I think I should give out a deprecation warning (maybe in `__init_subclass__`) when someone uses it

docs/cogs.rst
EmreTech: We could, but the problem is extensions are only in `ext.commands.Bot` and they haven't been implemented in `nextcord.Client` at all

nextcord/cog.py
```diff
@@ -0,0 +1,338 @@
...
+    """The base class that all cogs must inherit from.
+
+    A cog is a collection of commands, listeners, and optional state to
+    help group commands together. More information on them can be found on
+    the :ref:`cogs` page.
+
+    When inheriting from this class, the options shown in :class:`nextcord.CogMeta`
+    are equally valid here.
+
+    There is an alias for this called :class:`ClientCog`.
```
Skelmis, line 135: I feel like there's no need to state this given its deprecated and will be removed?

Skelmis (changes requested):
This fails to load a simplistic cog I'd expect to work:
Main: https://workbin.dev/?id=1664633956624160
Cog: https://workbin.dev/?id=166463398386748
Error: https://workbin.dev/?id=166463399413683

The ``listener`` it errors on is ``<function Cog.cog_application_command_before_invoke at 0x7faa9a5c5f30>``

EmreTech:
> This fails to load a simplistic cog I'd expect to work: Main: https://workbin.dev/?id=1664633956624160 Cog: https://workbin.dev/?id=166463398386748 Error: https://workbin.dev/?id=166463399413683
> 
> The `listener` it errors on is `<function Cog.cog_application_command_before_invoke at 0x7faa9a5c5f30>`

Fixed in 519fbfd

Skelmis (approved):
This has been tested under a simplistic cog case using `nextcord.Client`, I havent tested in bots using `commands.Bot` or a more complicated setup

docs/ext/commands/cogs.rst
```diff
@@ -5,15 +5,8 @@
 Cogs
 ====
 
-There comes a point in your bot's development when you want to organize a collection of commands, listeners, and some state into one class. Cogs allow you to do just that.
-
-The gist:
-
-- Each cog is a Python class that subclasses :class:`.commands.Cog`.
-- Every command is marked with the :func:`.commands.command` decorator.
```
DenverCoder1, line 13: > - Every command is marked with the :func:`.commands.command` decorator.

This should probably be documented somewhere, it seems like it's not.

Also, maybe regular Cog docs should link to commands.Cog telling people to look there for ext commands. Because of the naming, I could see it confusing some people.

nextcord/cog.py
```diff
@@ -0,0 +1,336 @@
...
+        Parameters
+        ----------
+        interaction: :class:`.Interaction`
+            The invocation interaction.
+        """
+        pass
+
+
+class ClientCog(Cog):
+    """An alias of :class:`Cog` for backwards compatibility reasons."""
```
DenverCoder1, line 328: This might be better in nextcord.application_command because moving it here would break imports like `from nextcord.application_command import ClientCog` if people are currently doing that.

Feel free to ignore if you think it's not necessary.

EmreTech: I could include the alias and add it to the `__all__` of `application_command.py`

DenverCoder1: That should also work

EmreTech:
a655d3e is a little messy, but it's the only solution I could come up with that doesn't have static type checker errors or runtime errors.

docs/ext/commands/cogs.rst
```diff
@@ -45,70 +38,13 @@ This example cog defines a ``Greetings`` category for your commands, with a sing
...
-            ...
-
-    class Gambling(commands.Cog):
-        def __init__(self, bot):
-            self.bot = bot
-
-        def coinflip(self):
-            return random.randint(0, 1)
-
-        @commands.command()
```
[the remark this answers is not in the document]
EmreTech: There is already a quick example that contains a `commands.Command`.

nextcord/types/checks.py
```diff
@@ -36,12 +36,10 @@
 MaybeCoro = Union[T, Coro[T]]
 CoroFunc = Callable[..., Coro[Any]]
 ApplicationCheck = Union[
-    Callable[[ClientCog, Interaction], MaybeCoro[bool]], Callable[[Interaction], MaybeCoro[bool]]
-]
-ApplicationHook = Union[
-    Callable[[ClientCog, Interaction], Coro[Any]], Callable[[Interaction], Coro[Any]]
+    Callable[[Cog, Interaction], MaybeCoro[bool]], Callable[[Interaction], MaybeCoro[bool]]
```
DenverCoder1, line 39: Things that use Cog should be in TYPE_CHECKING if Cog is imported in TYPE_CHECKING

(Basically the issue from #888 which could be fixed here instead or in addition depending on which is merged first)

EmreTech: bca6ff6

nextcord/ext/commands/bot.py
```diff
@@ -639,8 +644,10 @@ def add_cog(self, cog: Cog, *, override: bool = False) -> None:
                 raise nextcord.ClientException(f"Cog named {cog_name!r} already loaded")
             self.remove_cog(cog_name)
 
-        cog = cog._inject(self)
-        self.__cogs[cog_name] = cog
+        if isinstance(cog, Cog):
+            cog = cog._inject(self)
+            self.__cogs[cog_name] = cog
```
DenverCoder1, line 649: This probably can still happen since you already allowed __cogs to contain nextcord.Cog types and it's already using it to check if it already exists.

Maybe some of this can be moved to nextcord.Client so it can check for duplicates and consolidate `Bot.__cogs` and `Client._client_cogs`?

```suggestion
        self.__cogs[cog_name] = cog
```

EmreTech:
@TAG-Epic @ItsAleph Can you explain why you disliked this PR? Just disliking it without any message on why is not helpful.

nextcord/client.py
```diff
@@ -2543,23 +2553,116 @@ def _add_decorated_application_commands(self) -> None:
...
-    def add_cog(self, cog: Cog) -> None:
-        # cog.process_app_cmds()
+    def add_cog(self, cog: Cog, *, override: bool = False) -> None:
+        """Adds a "cog" to the client.
+
+        A cog is a class that has its own commands.
+
+        .. versionchanged:: 2.3
+
+            This method is now documented.
```
DenverCoder1, line 2568: I don't think it's necessary to document when it was documented
```suggestion
```

nextcord/client.py
```diff
@@ -2543,23 +2553,116 @@ def _add_decorated_application_commands(self) -> None:
...
+        Optional[:class:`nextcord.Cog`]
+             The cog that was removed. ``None`` if not found.
+        """
+        stored_cog: Optional[Cog] = None
+        if isinstance(cog, Cog):
+            for name, value in self.__cogs.items():
+                if value is cog:
+                    stored_cog = self.__cogs.pop(name)
+        else:
+            stored_cog = self.__cogs.pop(cog, None)
```
DenverCoder1, line 2657: Rather than iterating, can't it be done more efficiently in a single operation by checking for its unique cog name?
```suggestion
        if isinstance(cog, Cog):
            cog = cog.__cog_name__
        stored_cog = self.__cogs.pop(cog, None)
```

EmreTech: How did I not think of that lmao

nextcord/cog.py
```diff
@@ -0,0 +1,325 @@
...
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
```
ooliver1, line 23:
```suggestion
# SPDX-License-Identifier: MIT

```
#930

elenakrittik:
> @ItsAleph Can you explain why you disliked this PR? Just disliking it without any message on why is not helpful.

If i understand correctly, the original point for moving cogs outside of ext.commands was due to forcing users to use ext.commands. I agree with that point *as-is*, but i don't quite see where it comes from? Can't you just not use all `@commands.*command` decorators and utilize `@commands.Cog.listener` for handling interactions/messages manually (like you would without cogs, i suppose)? If that's a matter of namespaces, then why not move it to, e.g., ext.cogs? I am strongly against merging this directly into the main lib, as (to me, at least) it's purpose is to implement the API itself, while extensions are, well, extensions to it, whatever they're "extending".

ooliver1:
Maybe just `nextcord.cogs`? I do agree for not filling the main namespace, but `ext` is a bit misleading and I personally dislike the idea of ext and how it happens to also have "extensions" 'by default'.

EmreTech:
> Maybe just `nextcord.cogs`? I do agree for not filling the main namespace, but `ext` is a bit misleading and I personally dislike the idea of ext and how it happens to also have "extensions" 'by default'.

This is a better idea, and it resolves the issue with making `ClientCog` an alias to this new cog. Because we don't want to import an extension inside of the main library.

ooliver1 (commented):
I think I looked at this a while back, looks good but I will check again once conflicts are solved.

Skelmis:
@EmmmaTech care to take a crack at the merge conflicts?

EmmmaTech:
I haven't touched this PR because I wasn't sure if I wanted to keep cogs in 3.0 or replace them with a different solution like plugins. I'm not sure if we should go this route or just keep cogs and make a separate extension for plugins (exactly like disnake.ext.plugins, or get that to support nextcord natively).

Skelmis:
This is very true. We could just remove cogs in 3.0 and replace them with something better

EmmmaTech:
Yup, I created a [draft branch](https://github.com/EmmmaTech/nextcord/tree/feat/plugins) for the plugin concept, but I didn't open a PR for it since I wasn't sure whether we should go in the plugins direction.

Skelmis:
Provided it is actually advantageous compared to cogs and well documented enough I don't see why not

EmmmaTech:
I have decided to finally close this pr in favor of my in-progress plugins implementation.

PR #700, opened 2022-06-23
refactor(types): use NotRequired for optional fields

ooliver1:
## Summary

use `typing_extensions.NotRequired` for optional `TypedDict` fields instead of complicated inheritance

## Checklist

- [ ] This PR fixes an issue.


nextcord/types/appinfo.py
```diff
@@ -40,30 +43,23 @@ class BaseAppInfo(TypedDict):
...
-    hook: bool
-    terms_of_service_url: str
-    privacy_policy_url: str
-    max_participants: int
-    flags: int
-
-
-class PartialAppInfo(_PartialAppInfoOptional, BaseAppInfo):
-    pass
+class PartialAppInfo(BaseAppInfo):
```
EmreTech, line 58: I'm not sure if this would affect the inherited fields, but couldn't you use `total=False` here?

EmreTech (commented):
Looks good otherwise

ArjunSharda (commented):
Why is the license copyright year until 2021? It should be changed to 2022 / present.

ooliver1:
Why should Danny continue to have copyright?

ooliver1:
Completely refactored and redone for all the changes made since this PR was originally created.

ooliver1:
Sorry! Did not realise there's a max pending review count

h.py
```diff
@@ -0,0 +1,25 @@
+import re
```
DenverCoder1, line 1: This doesn't seem like it's supposed to be here?

DenverCoder1 (approved):
Looks good besides what looks like an unrelated file

nextcord/types/activity.py
```diff
@@ -26,6 +27,8 @@
 
 from typing import List, Literal, Optional, TypedDict
```
EmreTech, line 28: Wouldn't it make sense to use `TypedDict` from `typing_extensions`? Because I don't think the older `TypedDict` can handle `Required` and `NotRequired`

EmreTech (commented):
Looks fine otherwise

nextcord/types/activity.py
DenverCoder1: All the checks pass, so it seems like it's fine either way

EmreTech: Ah yeah that does make sense

teaishealthy (approved):

ArjunSharda (approved):

Skelmis (approved):
Untested, lgtm

PR #701, opened 2022-06-23
feat(help)!: Improve the looks of help command

### LINKED ISSUE #690: new default help command
### Summary

title

The core library

### The Problem

the right-now help command is not really good, its made when buttons isnt released

### The Ideal Solution

design a new default help command. maybe using an embed or button/dropdown?

### The Current Solution

_No response_

### Additional Context

_No response_

H4ckerxx44:
## Summary

Refactors `DefaultHelpCommand` to look less outdated and ugly.

You could consider this a breaking change due to `DefaultHelpCommand` loosing some functionality. Though, `DefaultHelpCommand` should not be subclassed to begin with. 

An image of how everything looks is in the comments.

resolves #690 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/ext/commands/core.py
+++ b/nextcord/ext/commands/core.py
@@ -64,7 +64,6 @@ if TYPE_CHECKING:
 
 __all__ = (
     "Command",
-    "Cog",
     "Group",
     "GroupMixin",
     "command",
--- a/nextcord/ext/commands/core.py
+++ b/nextcord/ext/commands/core.py
@@ -64,6 +64,7 @@ if TYPE_CHECKING:
 
 __all__ = (
     "Command",
+    "Cog",
     "Group",
     "GroupMixin",
     "command",
--- a/nextcord/ext/commands/help.py
+++ b/nextcord/ext/commands/help.py
@@ -32,11 +32,13 @@ from typing import TYPE_CHECKING
 
 import nextcord.utils
 
-from .core import Command, Group
+from .core import Command, Group, Cog
 from .errors import CommandError
+from ... import Embed, Message
+
 
 if TYPE_CHECKING:
-    from typing import Any, Callable, Optional
+    from typing import Any, Callable, Optional, Union, Mapping
 
     from .context import Context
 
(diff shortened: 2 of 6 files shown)


H4ckerxx44:
Main help
main_help

Cog help
cog_help

Group in cog
group_in_cog

Command in cog
command_in_cog

Group outside cog
group_outside_cog

Command outside cog
command_outside_cog

Command chaining
command_chaining

nextcord/ext/commands/core.py
```diff
@@ -64,6 +64,7 @@
 
 __all__ = (
     "Command",
+    "Cog",
```
ooliver1, line 67: why is this re-exported here? cog is in `commands/cog.py`

nextcord/ext/commands/help.py
```diff
@@ -32,11 +32,12 @@
 
 import nextcord.utils
 
-from .core import Command, Group
+from ... import Embed, Message
```
ooliver1, line 35:
```suggestion
from nextcord import Embed, Message
```
a bit cleaner

nextcord/ext/commands/help.py
```diff
@@ -887,203 +888,108 @@ class DefaultHelpCommand(HelpCommand):
...
 
-        note = self.get_ending_note()
-        if note:
-            self.paginator.add_line()
-            self.paginator.add_line(note)
+        return await self.send_embed(emb)
 
-        await self.send_pages()
+    # help <command>
+    async def send_command_help(self, cmd: Command) -> Union[Message, None]:
```
ooliver1, line 984: could maybe the full `command` word be used over `cmd`

ooliver1 (changes requested):

lol219:
Looks good

nextcord/ext/commands/core.py
H4ckerxx44: done

nextcord/ext/commands/help.py
H4ckerxx44: done

DenverCoder1:
Have you considered the possibility of subclassing a help command to add an "EmbedHelpCommand" for example, rather than overwriting the existing default?

That way it would be less of a breaking change since you could always choose to revert to the non-embed version.

H4ckerxx44:
> Have you considered the possibility of subclassing a help command to add an "EmbedHelpCommand" for example, rather than overwriting the existing default?
> 
> That way it would be less of a breaking change since you could always choose to revert to the non-embed version.

The point of this PR is to _improve_ the existing one and not make another default one. Personally, I think the current `DefaultHelpCommand` is _very_ outdated and it shows.

And in terms of breaking, as I have said in the summary, I don't know how to really evalute of the "breaking" status of that PR, because `HelpCommand` is supposed to be subclassed, not `DefaultHelpCommand`. Therefor I doubt that this is much breaking in a user-wide sense.

DenverCoder1:
While HelpCommand or minimalhelpcommand is more often subclassed, people do use defaulthelpcommand and it is documented. Doing a search turns up some results, mostly within jishaku it seems. I have not tested it yet, but it may be a good idea to test some existing use cases to make sure it's not going to immediately break a lot of bots.

https://github.com/search?q="Defaulthelpcommand"&type=code

H4ckerxx44:
> I have not tested it yet, [...]

@EmreTech as well as myself tested that and it should be golden.

> test some existing use cases to make sure it's not going to immediately break a lot of bots.

What I also could do is, still override `DefaultHelpCommand` but as an addition add the previous `DefaultHelpCommand` as something like `LegacyHelpCommand` or something.

This way, the old help commands gets replaced but would be a pretty easy fix but just changing the class.

H4ckerxx44:
I implemented a `LegacyHelpCommand` class which is basically the old ´DefaultHelpCommand´, that should be easy enough for devs that subclass the `DefaultHelpCommand` to fix, it's a simple class change.

DenverCoder1:
The current implementation of HelpCommand is meant to be customized by the user as the lib wants the user to be given the choice to decide the text, colors, and formatting. Help command preferences can be different for everyone and some designs may be better suited for different bots.

Currently there are 2 built-in choices for help commands (`DefaultHelpCommand` and `MinimalHelpCommand`) and it is easy to select the other in the bot's constructor (eg. `Bot(..., help_command=MinimalHelpCommand()`)

I think a collection of pre-built custom HelpCommands could be useful, though. Maybe in `/examples` or a third-party library that HelpCommands can be imported from.

As of now, I think sticking to Discord.py's decision may be best (https://github.com/Rapptz/discord.py/issues/5081).

H4ckerxx44:
In my opinion, the DefaultHelpCommand was never really intended to support a lot of customization, hence it being the default help command. Makes sense, why would the library do your work on being crazy customiziable?

Moreover, I don't agree with you on the `meant to be customized` part. The things you are allowed to customize are not really big things.
- `send to DM instead of channel` - not really customization but just a simple switch
- `indent` - that is bare customization at max
- `paginator` - that is indeed customization
- `width` - that falls off due to the embed

Also, you can change the text and heading with my implementation, too. 

Just by the looks of the command alone, you are very limited in what you can do. I personally don't think much people would spent a lot of time customizing `DefaultHelpCommand`, especially just be the not-great look of it alone, just does not look proper and / or not something that got the littliest amount of love in the development process of a bot.

Just like:
Would you really spent much money customizing a car which has 4 tires and a rusty body? Probably not.

I also wonder why you first suggest implementing some backwards-compatability, and 8 days after I did that suggesting something different.
What made you come to this different conclusion now? Genuinely curious.

DenverCoder1:
I'm still thinking about the best way to go about this. I personally don't think this change is necessary as the current help command system is the way it's been for years and it's working perfectly.

My original thought was that instead of changing the default, you could make a class that people could optionally use instead of the default. I did suggest it could be a default that could be reverted, but after thinking about it longer, I think changing the help command from the current default should be an optional opt-in.

I recently had the idea recently of creating a library of many custom help commands people can import to use in their project. Along the lines of that I remembered this PR and thought that might be a good alternative solution so I came back here to suggest it. If others think it's a good idea, we could go forward with it and I'd be happy to have collaboration on that.

I'm still open to suggestions on how your idea could best be implemented, for example, whether custom help commands like the embed version should be added in `nextcord.ext.commands.help` similar to `MinimalHelpCommand` or whether it should be a separate pypi download (I'm leaning toward the latter).

H4ckerxx44:
> Have you considered the possibility of subclassing a help command to add an "EmbedHelpCommand" for example, rather than overwriting the existing default?
> 
> That way it would be less of a breaking change since you could always choose to revert to the non-embed version.

Coming back on this, this seems like a nice solution, especially with the context of this dicussion here.

I can implement the `EmbedHelpCommand` and let `DefaultHelpCommand` as it is right now.
With it just being a new class, that would be a nice-looking help command which you can opt-in if you want to

Skelmis:
I like the idea of keeping the existing help command, with the idea of this moving to a separate ext with other help commands as was partially discussed on discord under the nextcord repo.

PR #682, opened 2022-06-16
feat: automod support 

MaskDuck:
## Summary

discord/discord-api-docs#4860

## adds
- `enums.TriggerType`
- `enums.EventType`
- `enums.KeywordPresetRule`
- `Intents.automod_configuration`
- `Intents.automod_execution`
- `Intents.automod`
- `automod.AutoModerationRule`
- `enums.ActionType`
- `enums.KeywordPresetType`
- `enums.EventType`
- `enums.TriggerType`
- `http.HTTPClient.list_guild_automod_rules`
- `http.HTTPClient.get_automod_rules`
- `http.HTTPClient.create_automod_rule`
- `http.HTTPClient.modify_automod_rule`
- `http.HTTPClient.delete_automod_rule`
- `MessageType.automod_action`
- `AuditLogAction.automod_rule_create`
- `AuditLogAction.automod_rule_update`
- `AuditLogAction.automod_rule_delete`
- `AuditLogAction.automod_block_message`
- `Guild.get_automod_rule`
- `Guild.get_automod_rules`

# edit
changed `Intents.all` to account for the skip in intents values because discord has not released the bitfields inbetween

## Checklist

- [ ] If code changes were made then they have been tested.
  - [x] I have updated the documentation to reflect the changes.
  - [x] I have run `task pyright` and fixed the relevant issues.
- [x] This PR fixes an issue.
- [x] This PR adds something new (e.g. new method or parameters).
- [x] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [x] This PR is **not** a code change (e.g. documentation, README, ...)

PR #706, opened 2022-06-26
feat: add remove_view

### LINKED ISSUE #425: Implement remove_view in Client
### Summary

Implement remove_view that reverse of add_view in Client.

The core library

### The Problem

Client class has method `add_view`.
This is useful, but if you do it in the constructor of the extension's cog, it will cause a memory leak unless you delete the registered view when the extension is reloaded.
However, `remove_view`, which is the opposite of `add_view` to remove a view, is not implemented.
This is very inconvenient.
Not only does this cause a memory leak, but if you pass an instance of the cog when instantiating the view, the old instance will remain and will be used.
This means that the old version of the cog will be used in the View.
So I thought it would be good to have `remove_view`.

### The Ideal Solution

Example on cog
```python
class ExampleCog(commands.Cog):
  def __init__(self, bot):
    self.view = MyView()
    bot.add_view(self.view)

  def cog_unload(self):
    bot.remove_view(self.view)
```

### The Current Solution

It's a little aggressive, but that's how I'm dealing with it now.
```python
# Remove view added
for key, value in list(bot._connection._view_store._views.items()):
  if key[2] == "CustomID on the view":
    del bot._connection._view_store._views[key]
```

### Additional Context

_No response_

MaskDuck:
## Summary

close #425 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/client.py
+++ b/nextcord/client.py
@@ -1824,7 +1824,7 @@ class Client:
         .. versionadded:: 2.0
 
         Parameters
-        ------------
+        ----------
         view: :class:`nextcord.ui.View`
             The view to remove from dispatching.
 
--- a/nextcord/client.py
+++ b/nextcord/client.py
@@ -1819,7 +1819,7 @@ class Client:
         self._connection.store_view(view, message_id)
 
     def remove_view(self, view: View) -> None:
-        """Remove a :class:`nextcord.ui.View` for persistant listening.
+        """Remove a :class:`nextcord.ui.View` from persistent listening.
 
         .. versionadded:: 2.0
 
--- a/nextcord/client.py
+++ b/nextcord/client.py
@@ -1826,7 +1826,7 @@ class Client:
         Parameters
         ------------
         view: :class:`nextcord.ui.View`
-            The view to register for dispatching.
+            The view to remove from dispatching.
 
         Raises
         -------
--- a/nextcord/client.py
+++ b/nextcord/client.py
@@ -1817,6 +1817,38 @@ class Client:
             )
 
         self._connection.store_view(view, message_id)
+    
+    def remove_view(self, view: View, *, message_id: Optional[int] = None) -> None:
+        """Remove a :class:`nextcord.ui.View` for persistant listening.
+
+        .. versionadded:: 2.0
+
+        Parameters
+        ------------
+        view: :class:`nextcord.ui.View`
+            The view to register for dispatching.
+        message_id: Optional[:class:`int`]
+            The message ID that the view is attached to. This is currently used to
+            refresh the view's state during message update events. If not given
+            then message update events are not propagated for the view.
+        
+        Raises
+        -------
+        TypeError
+            A view was not passed.
+        ValueError
+            The view is not persistent. A persistent view has no timeout
+            and all their components have an explicitly provided custom_id.
+        """
+        if not isinstance(view, View):
+            raise TypeError(f"Expected an instance of View not {view.__class__!r}")
+
+        if not view.is_persistent():
+            raise ValueError(
+                "View is not persistent. Items need to have a custom_id set and View must have no timeout"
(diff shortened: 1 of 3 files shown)


Skelmis:
Fix your description.

nextcord/client.py
```diff
@@ -1818,6 +1818,34 @@ def add_view(self, view: View, *, message_id: Optional[int] = None) -> None:
 
         self._connection.store_view(view, message_id)
 
+    def remove_view(self, view: View, *, message_id: Optional[int] = None) -> None:
```
Skelmis, line 1821: `message_id` isn't used? Is it meant to be or was this a typo

MaskDuck:
```suggestion
    def remove_view(self, view: View) -> None:
```

nextcord/client.py
```diff
@@ -1818,6 +1818,34 @@ def add_view(self, view: View, *, message_id: Optional[int] = None) -> None:
 
         self._connection.store_view(view, message_id)
 
+    def remove_view(self, view: View) -> None:
+        """Remove a :class:`nextcord.ui.View` for persistant listening.
```
DenverCoder1, line 1822:
```suggestion
        """Remove a :class:`nextcord.ui.View` for persistent listening.
```

nextcord/client.py
```diff
@@ -1818,6 +1818,34 @@ def add_view(self, view: View, *, message_id: Optional[int] = None) -> None:
...
+
+        .. versionadded:: 2.0
+
+        Parameters
+        ------------
+        view: :class:`nextcord.ui.View`
+            The view to register for dispatching.
```
DenverCoder1, line 1829:
```suggestion
            The view to remove from dispatching.
```

nextcord/client.py
```diff
@@ -1818,6 +1818,34 @@ def add_view(self, view: View, *, message_id: Optional[int] = None) -> None:
 
         self._connection.store_view(view, message_id)
 
+    def remove_view(self, view: View) -> None:
+        """Remove a :class:`nextcord.ui.View` from persistent listening.
+
+        .. versionadded:: 2.0
+
+        Parameters
+        ------------
```
ooliver1, line 1827:
```suggestion
        Parameters
        ----------
```

nextcord/client.py
```diff
@@ -1818,6 +1818,34 @@ def add_view(self, view: View, *, message_id: Optional[int] = None) -> None:
...
+        view: :class:`nextcord.ui.View`
+            The view to remove from dispatching.
+
+        Raises
+        -------
```
ooliver1, line 1832:
```suggestion
        Raises
        ------
```

ooliver1 (changes requested):

ooliver1 (changes requested):
has this been tested? please test PRs before submitting for review
```py
@bot.event
async def on_ready():
    bot.add_view(MyView(timeout=None))
    print(bot._connection._view_store._views)
    bot.remove_view(MyView(timeout=None))
    print(bot._connection._view_store._views)
```
prints the same value both times

`nextcord.state.ConnectionState.remove_view` does not add a `message_id` to the tuple, so it will never pop anything as a tuple of 2 items will never match tuples of 3

ooliver1:
Do you not think users should be able to customise that message_id variable to remove views made using that?

Skelmis (changes requested):
To add to Oliver's comment with a bit more of a code touch, I would expect the tests found [here](https://nexttest.koldfusion.xyz/nextcord/706) to pass before approving this pr

ooliver1:
Because of https://github.com/nextcord/nextcord/pull/682#issuecomment-1183046000, you were no longer able to make contributions to nextcord for the time being, so this PR is being closed as it would go stale from no updates.

PR #708, opened 2022-06-26
feat: slash default

MaskDuck:
## Summary

add slash default

for example:
```py
from nextcord import InteractionAuthor, SlashOption

bot = commands.Bot(...)

# suppose user didnt provide args
# before:
@bot.slash_command()
async def balance(inter: Interaction, user: SlashOption(name='user', description='h', required=False)):
  if user is None:
    user = inter.user
  ...
# after
@bot.slash_command()
async def rob(inter: Interaction, user: SlashOption(name='user', description='h', required=False, default=InteractionAuthor)):
  ...
```

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -93,6 +93,10 @@ __all__ = (
     "message_command",
     "user_command",
     "Mentionable",
+    "InteractionDefault",
+    "InteractionAuthor",
+    "InteractionGuild",
+    "InteractionChannel"
 )
 
 _log = logging.getLogger(__name__)
@@ -110,7 +114,39 @@ def _cog_special_method(func: FuncT) -> FuncT:
     func.__cog_special_method__ = None
     return func
 
+class InteractionDefault:
+    # TODO: fix docstring lol
+    """A class that makes all thing default.
+    
+    To make a default from the :class:`Interaction`, you can do like this:
+    .. code-block:: python3
+        class InteractionUser(InteractionDefault):
+            def default(self, interaction: Interaction):
+                return interaction.user
+        
+        @bot.slash_command()
+        async def balance(interaction, user = SlashOption(name="user", description="The user to check the balance", default=InteractionUser):
+            ...
+    """
+    def default(self, interaction: Interaction) -> None:
+        raise NotImplementedError
 
+class InteractionAuthor(InteractionDefault):
+    """Return the author of the interaction."""
+    def default(self, interaction: Interaction) -> User:
+        return interaction.author
+    
+class InteractionGuild(InteractionDefault):
+    """Return the guild of this interaction."""
+    def default(self, interaction: Interaction) -> Guild:
+        return interaction.guild
+
+class InteractionChannel(InteractionDefault):
+    """Return the channel that the interaction was executed in."""
+    def default(self, interaction: Interaction) -> GuildChannel:
+        return interaction.channel
+
+    
 class CallbackWrapper:
     """A class used to wrap a callback in a sane way to modify aspects of application commands.
 
@@ -1191,6 +1227,8 @@ class SlashOption(ApplicationCommandOption, _CustomTypingMetaBase):
         )
 
         self.autocomplete_callback: Optional[Callable] = autocomplete_callback
+        if issubclass(self.default, InteractionDefault):
+            self.default: Any = default.default
         self.default: Any = default
         self._verify: bool = verify
         if self._verify:


nextcord/application_command.py
```diff
@@ -111,6 +117,46 @@ def _cog_special_method(func: FuncT) -> FuncT:
...
+
+    def default(self, interaction: Interaction) -> None:
+        raise NotImplementedError
+
+
+class InteractionAuthor(InteractionDefault):
+    """Return the author of the interaction."""
+
+    def default(self, interaction: Interaction) -> User:
+        return interaction.author
```
EmreTech, line 143:
```suggestion
        return interaction.user
```

nextcord/application_command.py
```diff
@@ -111,6 +117,46 @@ def _cog_special_method(func: FuncT) -> FuncT:
...
+        @bot.slash_command()
+        async def balance(interaction, user = SlashOption(name="user", description="The user to check the balance", default=InteractionUser):
+            ...
+    """
+
+    def default(self, interaction: Interaction) -> None:
+        raise NotImplementedError
+
+
+class InteractionAuthor(InteractionDefault):
```
EmreTech, line 139:
```suggestion
class InteractionUser(InteractionDefault):
```

nextcord/application_command.py
```diff
@@ -111,6 +117,46 @@ def _cog_special_method(func: FuncT) -> FuncT:
...
+    """Return the author of the interaction."""
```
EmreTech, line 140:
```suggestion
    """Return the user of the interaction."""
```

nextcord/application_command.py
```diff
@@ -67,6 +67,8 @@
 from .utils import MISSING, find, maybe_coroutine, parse_docstring
 
 if TYPE_CHECKING:
+    from abc import MessageableChannel
```
EmreTech, line 70: You forgot to use relative imports here

EmreTech (changes requested):
Looks good, here are some suggestions

nextcord/application_command.py
MaskDuck:
```suggestion
    from .abc import MessageableChannel
```

nextcord/application_command.py
```diff
@@ -93,6 +94,10 @@
     "message_command",
     "user_command",
     "Mentionable",
+    "InteractionDefault",
+    "InteractionAuthor",
```
EmreTech, line 98:
```suggestion
    "InteractionUser",
```

EmreTech (approved):
Haven't tested, but the code should work

nextcord/application_command.py
```diff
@@ -1191,6 +1239,8 @@ def __init__(
         )
 
         self.autocomplete_callback: Optional[Callable] = autocomplete_callback
+        if issubclass(self.default, InteractionDefault):
```
ooliver1, line 1242: How does this even work? It's checking the current default on if it's a subclass but not checking the actual provided default

Skelmis: Note, it also appears to override it on the line after regardless..?

Artemis21:
Do these dynamic defaults need to be classes? If they only have one method, they could just be functions. That way people could even just use lambdas for simple things. (`default=lambda inter: inter.user`)

ooliver1:
I feel like type checkers wouldn't be happy with this either

nextcord/application_command.py
```diff
@@ -111,6 +117,48 @@ def _cog_special_method(func: FuncT) -> FuncT:
     return func
 
 
+class InteractionDefault:
+    # TODO: fix docstring lol
+    """A class that makes all thing default.
```
ooliver1, line 122: reword this, what is "thing"

nextcord/application_command.py
```diff
@@ -111,6 +117,48 @@ def _cog_special_method(func: FuncT) -> FuncT:
...
+
+    To make a default from the :class:`Interaction`, you can do like this:
```
ooliver1, line 124: this should use an `Example` header like `Parameters` and `Attributes`

nextcord/application_command.py
```diff
@@ -111,6 +117,48 @@ def _cog_special_method(func: FuncT) -> FuncT:
...
+    .. code-block:: python3
+        class InteractionUser(InteractionDefault):
+            def default(self, interaction: Interaction):
+                return interaction.user
+
+        @bot.slash_command()
+        async def balance(interaction, user = SlashOption(name="user", description="The user to check the balance", default=InteractionUser):
```
ooliver1, line 131: missing a closing bracket on the whole definition

nextcord/application_command.py
```diff
@@ -111,6 +117,48 @@ def _cog_special_method(func: FuncT) -> FuncT:
...
+            ...
+    """
+
+    def default(self, interaction: Interaction) -> None:
```
ooliver1, line 135: could this be a `runtime_checkable` `typing.Protocol` or `abc.ABC`?

ooliver1 (changes requested):

nextcord/application_command.py
MaskDuck: i mean, i don't have any idea to write

ooliver1: `thing` still isnt an acceptable description, use plain english - this is a default value setter for application command arguments, so describe it that way

EmreTech:
> Do these dynamic defaults need to be classes? If they only have one method, they could just be functions. That way people could even just use lambdas for simple things. (`default=lambda inter: inter.user`)

I like this implementation a lot more. It's very simple and clean to write unlike making an entire class just to grab for example the `app_permissions` attribute from the interaction as a default.

nextcord/application_command.py
ooliver1: This was never actually resolved with any communication or commits @MaskDuck.

nextcord/application_command.py
```diff
@@ -111,6 +117,48 @@ def _cog_special_method(func: FuncT) -> FuncT:
...
+    def default(self, interaction: Interaction) -> Optional[Guild]:
+        return interaction.guild
+
+
+class InteractionChannel(InteractionDefault):
+    """Return the channel that the interaction was executed in."""
+
+    def default(
+        self, interaction: Interaction
+    ) -> Optional[Union[GuildChannel, PartialMessageable, Thread]]:
```
ooliver1, line 158: this does not include all possibilities - such as `DMChannel`, might be best to use `abc.Messageable`

nextcord/application_command.py
ooliver1, line 131: This should probably also be typehinted to `User`/`Member` as that's what the default uses.

ooliver1 (changes requested):
As shown by https://github.com/nextcord/nextcord/pull/708/files#diff-f70bec2fa1cfe36ba24ace327b56d9736cdebc2c64b197c2d5d216888ce5dd97R1242-R1244, I cannot believe that you would have actually tested this.

ooliver1:
Your example test code also uses a typehint instead of a default, which is not how `SlashOption` works.

EmreTech (changes requested):

EmreTech:
As said by Denver [here](https://discord.com/channels/881118111967883295/881124672056467456/990550169311399966) and [here](https://github.com/nextcord/nextcord/pull/744#issuecomment-1186732691), this solution isn't needed at all. He has provided enough examples in the links I provided in this comment.

PR #714, opened 2022-06-27
feat(message): add Message.interaction

TrapDrap:
nextcord.Message.interaction - Represents a command interaction in a message

If Message contains a command interaction, nexcord.Message.interaction will return:
The command name as ".name" (str)
The command ID (message ID?) as ".id" (int)
The user as ".user" (nextcord.User)
The interaction type as ".type" (nextcord.enums.InteractionType)

This is useful to detect when an application command by another bot is used in a message, and know who used the command

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1518,28 +1518,6 @@ of :class:`enum.Enum`.
 
         See also :meth:`InteractionResponse.edit_message`
 
-.. class:: MessageInteraction
-
-    Represents the interaction data of a message if applicable
-
-    .. versionadded:: 2.0
-
-        Represents the raw data of the message interaction
-    .. attribute:: data
-
-        Represent the id of the message where the interaction occured
-    .. attribute:: id
-
-        Represents the type of message interaction
-    .. attribute:: type
-
-        Represents the name of the message interaction
-    .. attribute:: name
-
-        Represents the user that used the interaction
-        Or the member that used the interaction if guild applicable
-    .. attribute:: user
-
 .. class:: ComponentType
 
     Represents the component type of a component.
@@ -3726,6 +3704,14 @@ PartialInteractionMessage
 .. autoclass:: PartialInteractionMessage()
     :members:
     :inherited-members:
+
+MessageInteraction
+~~~~~~~~~~~~
+
+.. attributetable:: MessageInteraction
+
+.. autoclass:: MessageInteraction()
+    :members:
 
 Member
 ~~~~~~
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -3706,7 +3706,7 @@ PartialInteractionMessage
     :inherited-members:
 
 MessageInteraction
-~~~~~~~~~~~~
+~~~~~~~~~~~~~~~~~~
 
 .. attributetable:: MessageInteraction
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1528,14 +1528,14 @@ of :class:`enum.Enum`.
     .. attribute:: data
 
         Represent the id of the message where the interaction occured
-    .. attribute:: id 
+    .. attribute:: id
 
         Represents the type of message interaction
     .. attribute:: type
-    
+
         Represents the name of the message interaction
     .. attribute:: name
-    
+
         Represents the user that used the interaction
         Or the member that used the interaction if guild applicable
     .. attribute:: user
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -1518,6 +1518,28 @@ of :class:`enum.Enum`.
 
         See also :meth:`InteractionResponse.edit_message`
 
+.. class:: MessageInteraction
+
+    Represents the interaction data of a message if applicable
+
+    .. versionadded:: 2.0
+
+        Represents the raw data of the message interaction
+    .. attribute:: data
+
+        Represent the id of the message where the interaction occured
+    .. attribute:: id 
+
+        Represents the type of message interaction
+    .. attribute:: type
+    
+        Represents the name of the message interaction
+    .. attribute:: name
+    
+        Represents the user that used the interaction
+        Or the member that used the interaction if guild applicable
+    .. attribute:: user
+
 .. class:: ComponentType
 
     Represents the component type of a component.
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -1280,7 +1280,7 @@ class MessageInteraction:
     user: :class:`User`
         The user that sent the interaction.
     member: Optional[:class:`Member`]
-        The member that sent the interaction, 
+        The member that sent the interaction,
         if the interaction occured in a guild.
     """
 
@@ -1294,7 +1294,7 @@ class MessageInteraction:
         "member",
     )
 
-    def __init__(self, *, data: MessageInteractionPayload, guild:Guild, state: ConnectionState):
+    def __init__(self, *, data: MessageInteractionPayload, guild: Guild, state: ConnectionState):
         self._state: ConnectionState = state
 
         self.data: MessageInteractionPayload = data
@@ -1303,7 +1303,5 @@ class MessageInteraction:
         self.name: str = data["name"]
         self.user: User = User(data=data["user"], state=state)
         self.member: Optional[Member] = (
-            Member(data=data["member"], guild=guild, state=state)
-            if guild is not None
-            else None
+            Member(data=data["member"], guild=guild, state=state) if guild is not None else None
         )
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -1295,6 +1295,6 @@ class MessageInteraction:
 
         self.data: MessageInteractionPayload = data
         self.id: int = int(data["id"])
-        self.type: InteractionType = data["type"]
+        self.type: int = data["type"]
         self.name: str = data["name"]
         self.user: User = User(data=data["user"], state=state)
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -1264,6 +1264,7 @@ class MessageInteraction:
 
     A message's interaction data is a property of a message when the message
     is a response to an interaction from any bot.
+
     .. versionadded:: 2.0
 
     Attributes
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -50,7 +50,6 @@ __all__ = (
     "InteractionMessage",
     "InteractionResponse",
     "PartialInteractionMessage",
-    "MessageInteraction",
 )
 
 if TYPE_CHECKING:
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -1277,11 +1277,10 @@ class MessageInteraction:
         The interaction type.
     name: :class:`str`
         The application ID that the interaction was for.
-    user: :class:`User`
-        The user that sent the interaction.
-    member: Optional[:class:`Member`]
-        The member that sent the interaction,
-        if the interaction occured in a guild.
+    user: Union[:class:`Member`, :class:`User`]
+        The user or member that sent the interaction;
+        Member if interaction occured in a guild.
+        User if guild is None.
     """
 
     __slots__ = (
@@ -1291,7 +1290,6 @@ class MessageInteraction:
         "type",
         "name",
         "user",
-        "member",
     )
 
     def __init__(self, *, data: MessageInteractionPayload, guild: Guild, state: ConnectionState):
@@ -1301,7 +1299,8 @@ class MessageInteraction:
         self.id: int = int(data["id"])
         self.type: int = data["type"]
         self.name: str = data["name"]
-        self.user: User = User(data=data["user"], state=state)
-        self.member: Optional[Member] = (
-            Member(data=data["member"], guild=guild, state=state) if guild is not None else None
+        self.user: Union[Member, User] = (
+            Member(data=data["member"], guild=guild, state=state)
+            if guild is not None
+            else User(data=data["user"], state=state)
         )
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -1277,10 +1277,8 @@ class MessageInteraction:
         The interaction type.
     name: :class:`str`
         The application ID that the interaction was for.
-    user: Union[:class:`Member`, :class:`User`]
-        The user or member that sent the interaction;
-        Member if interaction occured in a guild.
-        User if guild is None.
+    user: :class:`User`
+        The user or that sent the interaction;
     """
 
     __slots__ = (
@@ -1293,7 +1291,7 @@ class MessageInteraction:
     )
 
     def __init__(
-        self, *, data: MessageInteractionPayload, guild: Optional[Guild], state: ConnectionState
+        self, *, data: MessageInteractionPayload, state: ConnectionState
     ):
         self._state: ConnectionState = state
 
@@ -1301,10 +1299,4 @@ class MessageInteraction:
         self.id: int = int(data["id"])
         self.type: int = data["type"]
         self.name: str = data["name"]
-
-        self.user: Union[Member, User]
-        if "member" in data and guild is not None:
-            data["member"]["user"] = data["user"]
-            self.user = Member(data=data["member"], guild=guild, state=state)
-        else:
-            self.user = User(data=data["user"], state=state)
+        self.user: User = User(data=data["user"], state=state)
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -1280,7 +1280,7 @@ class MessageInteraction:
         The raw data from the interaction.
     """
 
-    __slots__: = (
+    __slots__ = (
         "_state",
         "data",
         "id",
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -1269,6 +1269,8 @@ class MessageInteraction:
 
     Attributes
     -----------
+    data: :class:`dict`
+        The raw data from the interaction.
     id: :class:`int`
         The interaction's ID.
     type: :class:`InteractionType`
@@ -1277,8 +1279,9 @@ class MessageInteraction:
         The application ID that the interaction was for.
     user: :class:`User`
         The user that sent the interaction.
-    data: :class:`dict`
-        The raw data from the interaction.
+    member: Optional[:class:`Member`]
+        The member that sent the interaction, 
+        if the interaction occured in a guild.
     """
 
     __slots__ = (
@@ -1288,9 +1291,10 @@ class MessageInteraction:
         "type",
         "name",
         "user",
+        "member",
     )
 
-    def __init__(self, *, data: MessageInteractionPayload, state: ConnectionState):
+    def __init__(self, *, data: MessageInteractionPayload, guild:Guild, state: ConnectionState):
         self._state: ConnectionState = state
 
         self.data: MessageInteractionPayload = data
@@ -1298,3 +1302,8 @@ class MessageInteraction:
         self.type: int = data["type"]
         self.name: str = data["name"]
         self.user: User = User(data=data["user"], state=state)
+        self.member: Optional[Member] = (
+            Member(data=data["member"], guild=guild, state=state)
+            if guild is not None
+            else None
+        )
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -1303,6 +1303,6 @@ class MessageInteraction:
         self.name: str = data["name"]
         self.user: Union[Member, User] = (
             Member(data=data["member"], guild=guild, state=state)
-            if guild is not None
+            if "member" in data and guild is not None
             else User(data=data["user"], state=state)
         )
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -1305,6 +1305,6 @@ class MessageInteraction:
         self.user: Union[Member, User]
         if "member" in data and guild is not None:
             data["member"]["user"] = data["user"]
-            self.user =  Member(data=data["member"], guild=guild, state=state)
+            self.user = Member(data=data["member"], guild=guild, state=state)
         else:
-            self.user =  User(data=data["user"], state=state)
+            self.user = User(data=data["user"], state=state)
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -1297,4 +1297,8 @@ class MessageInteraction:
         self.id: int = int(data["id"])
         self.type: int = data["type"]
         self.name: str = data["name"]
-        self.user: User = User(data=data["user"], state=state)
+        if "member" in data and guild is not None:
+            member_user = {**data["member"], "user": data["user"]}
+            self.user = Member(state=self._state, guild=guild, data=member_user)  # type: ignore
+        else:
+            self.user = User(state=self._state, data=data["user"])
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -72,7 +72,6 @@ if TYPE_CHECKING:
     from .types.interactions import (
         Interaction as InteractionPayload,
         InteractionData,
-        MessageInteraction as MessageInteractionPayload,
     )
     from .ui.modal import Modal
     from .ui.view import View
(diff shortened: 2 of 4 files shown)


Skelmis (changes requested):

nextcord/interactions.py
```diff
@@ -1252,3 +1257,44 @@ class InteractionMessage(_InteractionMessageMixin, Message):
...
+    Attributes
+    -----------
+    id: :class:`int`
+        The interaction's ID.
+    type: :class:`InteractionType`
+        The interaction type.
+    name: :class:`str`
+        The application ID that the interaction was for.
+    user: :class:`User`
+        The user or member that sent the interaction.
```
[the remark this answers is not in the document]
TrapDrap: Oh yea that's actually a comment error, it cannot be "member", only user, I will fix that

TrapDrap (commented):
Fixed suggested typos and changed to suggested phrasing

nextcord/message.py
```diff
@@ -658,6 +658,9 @@ class Message(Hashable):
         .. versionadded:: 2.0
     guild: Optional[:class:`Guild`]
         The guild that the message belongs to, if applicable.
+
```
MaskDuck, line 661:
```suggestion
```

nextcord/types/interactions.py
```diff
@@ -228,6 +228,13 @@ class Interaction(_InteractionOptional):
     version: int
 
 
+class MessageInteraction(TypedDict):
```
MaskDuck, line 231: looks like this is already defined?

nextcord/interactions.py
```diff
@@ -1252,3 +1257,43 @@ class InteractionMessage(_InteractionMessageMixin, Message):
...
+        The interaction type.
+    name: :class:`str`
+        The application ID that the interaction was for.
+    user: :class:`User`
+        The user that sent the interaction.
+    data: :class:`dict`
+        The raw data from the interaction.
+    """
+
+    __slots__: Tuple[str, ...] = (
```
MaskDuck, line 1283:
```suggestion
    __slots__: = (
```
no need to typehint this

nextcord/message.py
```diff
@@ -781,6 +785,14 @@ def __init__(
...
                 continue
 
+        from .interactions import MessageInteraction
+
+        self.interaction: Optional[MessageInteraction]
+        try:
+            self.interaction = MessageInteraction(data=data["interaction"], state=state)
+        except KeyError:
+            self.interaction = None
+
```
MaskDuck, line 795:
```suggestion
        self.interaction: Optional[MessageInteraction] = MessageInteraction(data.get("interaction")) if data.get("interaction") is not None else None

```

nextcord/interactions.py
```diff
@@ -1252,3 +1257,43 @@ class InteractionMessage(_InteractionMessageMixin, Message):
...
+    )
+
+    def __init__(self, *, data: MessageInteractionPayload, state: ConnectionState):
+        self._state: ConnectionState = state
+
+        self.data: MessageInteractionPayload = data
+        self.id: int = int(data["id"])
+        self.type: InteractionType = data["type"]
+        self.name: str = data["name"]
+        self.user: User = self._state.store_user(data["user"])
```
MaskDuck, line 1299:
```suggestion
        self.user: User = User(data["user"])
```
it's already stored, no need to re-store it

MaskDuck (changes requested):
also don't forget to update the message payload

nextcord/message.py
```diff
@@ -781,6 +785,14 @@ def __init__(
             except KeyError:
                 continue
 
+        from .interactions import MessageInteraction
```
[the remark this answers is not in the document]
TrapDrap: it could be, but its a type so put it in there, does it improve performance if its in the top file?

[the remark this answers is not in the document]
MaskDuck: @TrapDrap consistency with other part of the lib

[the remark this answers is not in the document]
MaskDuck: also it's weird to have an import in the code, most of python codes usually have import on the top of the file.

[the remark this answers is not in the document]
ooliver1: Performance makes no difference, no other code will run until all imports are resolved - which end up loading all files, imports are cached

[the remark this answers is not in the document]
Skelmis: There are also valid reasons for importing items outside of the top of files

nextcord/message.py
ooliver1: Could you not use `if "interaction" in data` and ["item"]

nextcord/interactions.py
```diff
@@ -1252,3 +1257,43 @@ class InteractionMessage(_InteractionMessageMixin, Message):
...
+    )
+
+    def __init__(self, *, data: MessageInteractionPayload, state: ConnectionState):
+        self._state: ConnectionState = state
+
+        self.data: MessageInteractionPayload = data
+        self.id: int = int(data["id"])
+        self.type: InteractionType = data["type"]
+        self.name: str = data["name"]
+        self.user: User = User(data["user"])
```
MaskDuck, line 1299:
```suggestion
        self.user: User = User(user=data["user"], state=state)
```

nextcord/interactions.py
```diff
@@ -1252,3 +1257,43 @@ class InteractionMessage(_InteractionMessageMixin, Message):
...
+        The interaction type.
+    name: :class:`str`
+        The application ID that the interaction was for.
+    user: :class:`User`
+        The user that sent the interaction.
+    data: :class:`dict`
+        The raw data from the interaction.
+    """
+
+    __slots__: = (
```
MaskDuck, line 1283:
```suggestion
    __slots__ = (
```

nextcord/message.py
```diff
@@ -658,6 +659,8 @@ class Message(Hashable):
         .. versionadded:: 2.0
     guild: Optional[:class:`Guild`]
         The guild that the message belongs to, if applicable.
+    interaction: Optional[:class:`Interaction`]
```
DenverCoder1, line 662:
```suggestion
    interaction: Optional[:class:`MessageInteraction`]
```

nextcord/interactions.py
```diff
@@ -1252,3 +1257,43 @@ class InteractionMessage(_InteractionMessageMixin, Message):
...
+
+    Attributes
+    -----------
+    id: :class:`int`
+        The interaction's ID.
+    type: :class:`InteractionType`
+        The interaction type.
+    name: :class:`str`
+        The application ID that the interaction was for.
+    user: :class:`User`
```
DenverCoder1, line 1277: The Discord docs call this `member` and it is optional

https://discord.com/developers/docs/interactions/receiving-and-responding#message-interaction-object-message-interaction-structure

nextcord/types/interactions.py
```diff
@@ -228,6 +228,13 @@ class Interaction(_InteractionOptional):
...
+    id: Snowflake
+    type: InteractionType
+    name: str
+    user: User
```
ooliver1, line 235: why was this just moved?

ooliver1 (changes requested):
this evidently has not been tested.

please use the pr template when submitting PRs.

nextcord/interactions.py
```diff
@@ -1252,3 +1257,43 @@ class InteractionMessage(_InteractionMessageMixin, Message):
...
 
     pass
+
+
+class MessageInteraction:
+    """Represents a message's interaction data, regardless of application.
+
+    A message's interaction data is a property of a message when the message
+    is a response to an interaction from any bot.
+    .. versionadded:: 2.0
```
DenverCoder1, line 1267:
```suggestion

    .. versionadded:: 2.0
```

nextcord/message.py
```diff
@@ -781,6 +785,12 @@ def __init__(
             except KeyError:
                 continue
 
+        self.interaction: Optional[MessageInteraction] = (
+            MessageInteraction(data.get("interaction"))
```
DenverCoder1, line 789: You made data a keyword argument and state also need state to call the constructor. Possibly something like this?
```suggestion
            MessageInteraction(data=data.get("interaction"), state=self._state)
```

nextcord/types/interactions.py
DenverCoder1, line 231: `member` can be included as a non-required key

https://discord.com/developers/docs/interactions/receiving-and-responding#message-interaction-object-message-interaction-structure

DenverCoder1 (changes requested):

nextcord/interactions.py
```diff
@@ -1252,3 +1257,43 @@ class InteractionMessage(_InteractionMessageMixin, Message):
...
+    )
+
+    def __init__(self, *, data: MessageInteractionPayload, state: ConnectionState):
+        self._state: ConnectionState = state
+
+        self.data: MessageInteractionPayload = data
+        self.id: int = int(data["id"])
+        self.type: InteractionType = data["type"]
+        self.name: str = data["name"]
+        self.user: User = User(user=data["user"], state=state)
```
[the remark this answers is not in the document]
TrapDrap: I have tested the submission beforehand, this part was suggested code that I have applied without overlooking. Thank you for correcting that.

nextcord/message.py
```diff
@@ -781,6 +785,12 @@ def __init__(
...
+            if data.get("interaction") is not None
```
ooliver1, line 790:
```suggestion
            if "interaction" in data
```

ooliver1 (changes requested):

nextcord/types/interactions.py
TrapDrap: it was not, the one below was a duplicate (very stupid mistake ik)

nextcord/interactions.py
TrapDrap: I'm not sure how to combine the message's guild with the member from the interaction

nextcord/interactions.py
```diff
@@ -1252,3 +1257,44 @@ class InteractionMessage(_InteractionMessageMixin, Message):
     """
 
     pass
+
+
+class MessageInteraction:
```
DenverCoder1, line 1262: This class should probably be added to api.rst so it shows up in the docs

nextcord/interactions.py
```diff
@@ -1252,3 +1257,51 @@ class InteractionMessage(_InteractionMessageMixin, Message):
...
+        The interaction's ID.
+    type: :class:`InteractionType`
+        The interaction type.
+    name: :class:`str`
+        The application ID that the interaction was for.
+    user: :class:`User`
+        The user that sent the interaction.
+    member: Optional[:class:`Member`]
+        The member that sent the interaction,
+        if the interaction occured in a guild.
```
DenverCoder1, line 1284: For consistency with Interaction, this should probably be just under `MessageInteraction.user` and just make it sometimes a `User` and sometimes a `Member`

nextcord/types/interactions.py
```diff
@@ -232,7 +232,7 @@ class MessageInteraction(TypedDict):
     id: Snowflake
     type: InteractionType
     name: str
-    user: User
+    user: Union[Member, User]
```
DenverCoder1, line 235: Types is supposed to represent Discord's data objects, so this should be just User
```suggestion
    user: User
```
Member can be added in an optional class for consistency, eg.
```py
class _MessageInteractionOptional(TypedDict, total=False):
	member: Member
```
And have MessageInteraction subclass it

nextcord/interactions.py
```diff
@@ -1252,3 +1257,50 @@ class InteractionMessage(_InteractionMessageMixin, Message):
...
+    __slots__ = (
+        "_state",
+        "data",
+        "id",
+        "type",
+        "name",
+        "user",
+    )
+
+    def __init__(self, *, data: MessageInteractionPayload, guild: Guild, state: ConnectionState):
```
DenverCoder1, line 1295:
```suggestion
    def __init__(self, *, data: MessageInteractionPayload, guild: Optional[Guild], state: ConnectionState):
```

nextcord/types/interactions.py
DenverCoder1: `member` still needs to be added to the TypedDict (see above)

TrapDrap: I was doing that, couldnt find my glasses lol💀
I saved changes

nextcord/interactions.py
TrapDrap: Should be added, please check if anything is wrong when you can 🤍

docs/api.rst
```diff
@@ -1518,6 +1518,28 @@ of :class:`enum.Enum`.
...
+        Represents the type of message interaction
+    .. attribute:: type
+
+        Represents the name of the message interaction
+    .. attribute:: name
+
+        Represents the user that used the interaction
+        Or the member that used the interaction if guild applicable
+    .. attribute:: user
+
```
MaskDuck, line 1542:
```suggestion
    :members:
```

MaskDuck (changes requested):
did you add attrs to the message payload yet

docs/api.rst
TrapDrap: What is the suggested change?

TrapDrap: It seems you just removed most of the attrs, but thanks for reminding me to update that last part

nextcord/interactions.py
```diff
@@ -1252,3 +1257,44 @@ class InteractionMessage(_InteractionMessageMixin, Message):
...
+    )
+
+    def __init__(self, *, data: MessageInteractionPayload, state: ConnectionState):
+        self._state: ConnectionState = state
+
+        self.data: MessageInteractionPayload = data
+        self.id: int = int(data["id"])
+        self.type: int = data["type"]
+        self.name: str = data["name"]
+        self.user: User = User(data=data["user"], state=state)
```
DenverCoder1, line 1300: Here's one way to get the member (tested).
You can put back all the good changes you reverted.
```suggestion
        if "member" in data and guild is not None:
            member_user = {**data["member"], "user": data["user"]}
            self.user = Member(state=self._state, guild=guild, data=member_user)  # type: ignore
        else:
            self.user = User(state=self._state, data=data["user"])
```

TrapDrap: Gotcha

nextcord/message.py
```diff
@@ -53,6 +53,7 @@
 from .file import File
 from .flags import MessageFlags
 from .guild import Guild
+from .interactions import MessageInteraction
```
DenverCoder1, line 56: This is causing a circular import since `interactions.py` imports from `message.py` already.

There are few ways to solve this that I can think of:
1. Move `MessageInteraction` into `message.py`
2. Make a separate file for the `MessageInteraction` class to import it from
3. Create a helper method on ConnectionState to create message interaction objects (not ideal)
4. Move `from .interactions import MessageInteraction` into the Message `__init__` (hacky solution, not ideal)

Possibly someone can come up with a better idea, but something has to be done to make this run at all.

nextcord/interactions.py
```diff
@@ -1252,3 +1257,48 @@ class InteractionMessage(_InteractionMessageMixin, Message):
...
+
+        self.data: MessageInteractionPayload = data
+        self.id: int = int(data["id"])
+        self.type: int = data["type"]
+        self.name: str = data["name"]
+        if "member" in data and guild is not None:
+            member_user = {**data["member"], "user": data["user"]}
+            self.user = Member(state=self._state, guild=guild, data=member_user)  # type: ignore
+        else:
+            self.user = User(state=self._state, data=data["user"])
```
DenverCoder1, line 1304: A `__repr__` method would help with printing the data more neatly
```suggestion
            self.user = User(state=self._state, data=data["user"])
            
            def __repr__(self):
                return f"<{self.__class__.__name__} id={self.id} type={self.type} name={self.name} user={self.user!r}>"
```

nextcord/message.py
```diff
@@ -543,6 +544,62 @@ def flatten_handlers(cls):
...
 
+class _MessageInteractionOptional(TypedDict, total=False):
+    member: Member
+
+
+class MessageInteractionPayload(_MessageInteractionOptional):
+    id: Snowflake
+    type: InteractionType
+    name: str
+    user: User
```
DenverCoder1, line 555: Don't put any typed dicts here. They should be in `nextcord.types` where they were before.

nextcord/interactions.py
```diff
@@ -50,6 +50,7 @@
     "InteractionMessage",
     "InteractionResponse",
     "PartialInteractionMessage",
+    "MessageInteraction",
```
DenverCoder1, line 53: This should be in the `__all__` of message.py since you moved it.

nextcord/message.py
```diff
@@ -543,6 +544,62 @@ def flatten_handlers(cls):
...
+    __slots__ = (
+        "_state",
+        "data",
+        "id",
+        "type",
+        "name",
+        "user",
+    )
+
+    def __init__(self, *, data: MessageInteractionPayload, guild: Guild, state: ConnectionState):
```
DenverCoder1, line 590: Still needs to be optional
```suggestion
    def __init__(self, *, data: MessageInteractionPayload, guild: Optional[Guild], state: ConnectionState):
```

nextcord/interactions.py
TrapDrap: give me a couple of minutes to recoup and fix everything, getting a bit overwhelmed lol

TrapDrap: yes I am aware, thanks for the notice tho

nextcord/interactions.py
TrapDrap: alright

nextcord/types/interactions.py
```diff
@@ -248,19 +248,23 @@ class InteractionResponse(_InteractionResponseOptional):
...
 
+class _MessageInteractionOptional(TypedDict, total=False):
+    member: Member
+
+
+class MessageInteraction(_MessageInteractionOptional):
+    id: Snowflake
+    type: InteractionType
+    name: str
+    user: User
```
DenverCoder1, line 265: These classes should be moved above `_EditApplicationCommandOptional`, so that it's together with `EditApplicationCommand` and also to reduce the PR diff.

DenverCoder1 (commented):
Looks much better now :+1:

nextcord/types/interactions.py
TrapDrap: interesting, I thought they already were

TrapDrap: looks like I did an oops on reading the original doc

nextcord/message.py
```diff
@@ -543,6 +545,56 @@ def flatten_handlers(cls):
...
+    Attributes
+    -----------
+    data: :class:`dict`
+        The raw data from the interaction.
+    id: :class:`int`
+        The interaction's ID.
+    type: :class:`InteractionType`
+        The interaction type.
+    name: :class:`str`
+        The application ID that the interaction was for.
```
DenverCoder1, line 565:
```suggestion
        The name of the application command.
```

nextcord/message.py
```diff
@@ -543,6 +545,56 @@ def flatten_handlers(cls):
...
+    user: Union[:class:`User`, :class:`Member`]
+        The user or that sent the interaction;
+        Or member if the interaction occured in a guild.
```
DenverCoder1, line 568:
```suggestion
        The :class:`User` who invoked the interaction or :class:`Member` if the interaction
        occurred in a guild.
```

nextcord/message.py
```diff
@@ -543,6 +545,56 @@ def flatten_handlers(cls):
     return cls
 
 
+class MessageInteraction:
+    """Represents a message's interaction data, regardless of application.
```
DenverCoder1, line 549: I feel it's a bit redundant given the extended description
```suggestion
    """Represents a message's interaction data.
```

DenverCoder1 (approved):
Tested :+1:

docs/api.rst
```diff
@@ -3705,6 +3705,14 @@ PartialInteractionMessage
     :members:
     :inherited-members:
 
+MessageInteraction
+~~~~~~~~~~~~
```
DenverCoder1, line 3709: Should match the length
```suggestion
~~~~~~~~~~~~~~~~~~
```

TrapDrap: Should I re-request review in charge requests? Or just wait

nextcord/message.py
```diff
@@ -543,6 +545,56 @@ def flatten_handlers(cls):
...
+        self._state: ConnectionState = state
+
+        self.data: MessageInteractionPayload = data
+        self.id: int = int(data["id"])
+        self.type: int = data["type"]
+        self.name: str = data["name"]
+        if "member" in data and guild is not None:
+            self.user = Member(state=self._state, guild=guild, data={**data["member"], "user": data["user"]})  # type: ignore
+        else:
+            self.user = User(state=self._state, data=data["user"])
```
DenverCoder1, line 592: This actually isn't working since User is only imported under TYPE_CHECKING. Another way to do this is
```suggestion
            self.user = self._state.create_user(data=data["user"])
```

TrapDrap: I originally had it as '_state.store_user', which does the same but also caches, both should work for just the data tho, 🤔

nextcord/message.py
```diff
@@ -543,6 +545,56 @@ def flatten_handlers(cls):
     return cls
 
 
+class MessageInteraction:
+    """Represents a message's interaction data.
+
+    A message's interaction data is a property of a message when the message
+    is a response to an interaction from any bot.
+
+    .. versionadded:: 2.0
```
Skelmis, line 554: I believe this will be `2.1` now? Right @ooliver1

ooliver1: Correct, sorry @TrapDrap but could you edit these versionadded please

nextcord/message.py
```diff
@@ -543,6 +545,56 @@ def flatten_handlers(cls):
...
+        self.type: int = data["type"]
+        self.name: str = data["name"]
+        if "member" in data and guild is not None:
+            self.user = Member(state=self._state, guild=guild, data={**data["member"], "user": data["user"]})  # type: ignore
+        else:
+            self.user = self._state.create_user(data=data["user"])
+
+    def __repr__(self):
+        return f"<{self.__class__.__name__} id={self.id} type={self.type} name={self.name} user={self.user!r}>"
+
```
DenverCoder1, line 596: As this object has a unique snowflake ID, we should be able to turn this into a created_at timetamp like most other objects. This time will be different from the message since the interaction occurs before the message is sent.

```suggestion

    @property
    def created_at(self) -> datetime.datetime:
        """:class:`datetime.datetime`: The interaction's creation time in UTC."""
        return snowflake_time(self.id)
```

nextcord/message.py
```diff
@@ -543,6 +545,56 @@ def flatten_handlers(cls):
     return cls
 
 
+class MessageInteraction:
```
DenverCoder1, line 548: Similarly, as this has a Discord id, we can have it implement Hashable.
```suggestion
class MessageInteraction(Hashable):
```

nextcord/message.py
TrapDrap: Interesting, never knew that

nextcord/message.py
TrapDrap: Also gotta add the .created_at to the docs

DenverCoder1: The docs should keep track of properties like that automatically

nextcord/message.py
TrapDrap: changed

nextcord/message.py
```diff
@@ -543,6 +545,61 @@ def flatten_handlers(cls):
     return cls
 
 
+class MessageInteraction(Hashable):
+    """Represents a message's interaction data.
+
+    A message's interaction data is a property of a message when the message
+    is a response to an interaction from any bot.
+
+    .. versionadded:: 2.1
```
DenverCoder1, line 554: Probably the last suggestion I'll have. These docs can be added due to it being Hashable.

```suggestion
    .. versionadded:: 2.1
    
    .. container:: operations

        .. describe:: x == y

            Checks if two message interactions are equal.

        .. describe:: x != y

            Checks if two interaction messages are not equal.

        .. describe:: hash(x)

            Returns the message interaction's hash.
```

DenverCoder1 (approved):

ooliver1 (changes requested):

nextcord/message.py
```diff
@@ -543,6 +545,75 @@ def flatten_handlers(cls):
...
+
+            Checks if two interaction messages are not equal.
+
+        .. describe:: hash(x)
+
+            Returns the message interaction's hash.
+
+    Attributes
+    -----------
+    data: :class:`dict`
```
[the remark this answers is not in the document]
TrapDrap: I'm not so sure about that, the dict can contain str (name), int (id and type), and dicts for the partial member & user, that's it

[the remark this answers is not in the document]
DenverCoder1: The key is a string ("name", "id", "type", etc.), and the values can be may things like you said. So to keep it simple, Any can be put there.

``Dict[:class:`str`, Any]`` would just mean that the keys of the dictionary are strings.

Skelmis (approved):

ooliver1 (approved):

PR #719, opened 2022-06-28
refactor(bot)!: remove the use of **options for options

ooliver1:
## Summary

Continuation of #695
Uses actual kwargs over pass-down, ignored `**kwargs`.

### Breaking Changes

This is breaking for the same reason as #695.

### Reasons for seemingly unrelated edits

`_DefaultRepr` inherits from `HelpCommand` so it can continue to be used as the default for `help_command` without type issues.
`Client.__init__` is called manually so `GroupMixin` does not need pass-down `**kwargs` to call `super().__init__` another time unnecessarily. `bot.py` *may* need to have the idea of `BotBase` removed to eliminate similar issues such as this (different PR).
`get_prefix` is edited to use an else clause over overwriting the value so the type checker is aware of the `TypeGuard` which `callable` provides.
`maybe_coroutine` is included to have types so the type checker is at least aware that this will not spit out the original callable, `Any` unfortunately has to be used as thats the `TypeGuard` that `_isawaitable` provides.

### Test Code

https://paste.nextcord.dev/?id=1656448764816195

## Checklist

- [ ] This PR fixes an issue.

## Priority

High as the last PR broke all bot-specific kwargs :confused:

### DIFF
--- a/nextcord/ext/commands/bot.py
+++ b/nextcord/ext/commands/bot.py
@@ -52,9 +52,6 @@ from typing import (
 )
 
 import nextcord
-from nextcord.activity import BaseActivity
-from nextcord.enums import Status
-from nextcord.flags import Intents
 
 from . import errors
 from .cog import Cog
@@ -69,6 +66,8 @@ if TYPE_CHECKING:
     import aiohttp
     from typing_extensions import Self
 
+    from nextcord.activity import BaseActivity
+    from nextcord.enums import Status
     from nextcord.flags import MemberCacheFlags
     from nextcord.mentions import AllowedMentions
     from nextcord.message import Message
@@ -181,7 +180,7 @@ class BotBase(GroupMixin):
         shard_id: Optional[int] = None,
         shard_count: Optional[int] = None,
         application_id: Optional[int] = None,
-        intents: Intents = Intents.default(),
+        intents: nextcord.Intents = nextcord.Intents.default(),
         member_cache_flags: MemberCacheFlags = MISSING,
         chunk_guilds_at_startup: bool = MISSING,
         status: Optional[Status] = None,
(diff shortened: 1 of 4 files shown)


nextcord/ext/commands/core.py
```diff
@@ -1202,13 +1202,11 @@ class GroupMixin(Generic[CogT]):
...
     """
 
-    def __init__(self, *args: Any, **kwargs: Any) -> None:
-        case_insensitive = kwargs.get("case_insensitive", False)
+    def __init__(self, *, case_insensitive: bool = False) -> None:
         self.all_commands: Dict[str, Command[CogT, Any, Any]] = (
             _CaseInsensitiveDict() if case_insensitive else {}
         )
         self.case_insensitive: bool = case_insensitive
-        super().__init__(*args, **kwargs)
```
DenverCoder1, line 1211: #695 broke `case_insensitive` since it is getting passed to Client as part of kwargs here.

Removing this call to `Client.__init__` fixes that bug, so no suggestion here, just saying priority might be a bit higher for this change.

nextcord/ext/commands/bot.py
```diff
@@ -35,9 +35,26 @@
...
+    Sequence,
+    Type,
+    TypeVar,
+    Union,
+)
 
 import nextcord
+from nextcord.activity import BaseActivity
+from nextcord.enums import Status
+from nextcord.flags import Intents
```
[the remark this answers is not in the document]
ooliver1: intents cannot as the default value is used, but the rest can yes

DenverCoder1 (approved):

Skelmis (approved):
Untested, however looks good to me.

nextcord/ext/commands/bot.py
```diff
@@ -149,9 +162,80 @@ class MissingMessageContentIntentWarning(UserWarning):
...
+        rollout_delete_unknown: bool = True,
+        rollout_register_new: bool = True,
+        rollout_update_known: bool = True,
+        rollout_all_guilds: bool = False,
+        owner_id: Optional[int] = None,
+        owner_ids: Iterable[int] = set(),
+        strip_after_prefix: bool = False,
+        case_insensitive: bool = False,
+    ):
+        nextcord.Client.__init__(
```
DenverCoder1, line 207: This could be `nextcord.Client` or `nextcord.AutoShardedClient`.

I think this is what broke `AutoShardedBot`s

PR #724, opened 2022-07-01
fix(interactions): get info from cache first if available

### LINKED ISSUE #605: Status not displaying properly
### Summary

nextcord not responding with the correct Status

### Reproduction Steps

1. Enable presence/member intent
2. make a slash cmd that replies with `interaction.user.raw_status`

### Minimal Reproducible Code

```python
@client.slash_command(guild_ids=TESTING_GUILD_IDS)
async def presence(interaction: Interaction):
    await interaction.send(interaction.user.raw_status)
```

### Expected Results

show the correct status like dnd, online etc.

### Actual Results

It always displays offline whereas the user has a different status.
doing this `interaction.guild.get_member(interaction.user.id).raw_status` displays the right status but `interaction.user.raw_status` doesn't

### Intents

all

### System Information

- Python v3.9.1-final
- nextcord v2.0.0-alpha
    - nextcord pkg_resources: v2.0.0a9
- aiohttp v3.8.1
- system info: Windows 10 10.0.19041

### Checklist

- [x] I have removed my token from display, if visible.

### Additional Context

_No response_

Soheab:
## Summary

Fixes #605

This PR changes the assignment of the following attributes for `Interaction` to try cache before constructing the object: `.message, .user`

This is needed because fields like `status` are not sent in the Interaction member payload but are usually cached.

Preview: image
I don't know how to test `.message` and User for `.user` but the correct object is returned.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -211,13 +211,12 @@ class Interaction:
 
         # TODO: there's a potential data loss here
         if self.guild_id:
+            guild = self.guild or Object(int(self.guild_id))
             try:
                 member = data["member"]
             except KeyError:
                 pass
             else:
-                guild = self.guild or Object(int(self.guild_id))
-
                 user_id = int(member["user"]["id"])  # type: ignore
                 cached_member = self.guild and self.guild.get_member(user_id)
                 self.user = cached_member or Member(state=self._state, guild=guild, data=member)  # type: ignore
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -217,8 +217,8 @@ class Interaction:
             except KeyError:
                 pass
             else:
-                cached_member = self.guild and self.guild.get_member(int(member["user"]["id"]))  # type: ignore
-                self.user = cached_member or Member(state=self._state, guild=guild, data=member)  # type: ignore
+                cached_member = self.guild and self.guild.get_member(int(member["user"]["id"]))  # type: ignore # user key should be present here
+                self.user = cached_member or Member(state=self._state, guild=guild, data=member)  # type: ignore # user key should be present here
                 self._permissions = int(member.get("permissions", 0))
         else:
             try:
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -217,8 +217,7 @@ class Interaction:
             except KeyError:
                 pass
             else:
-                user_id = int(member["user"]["id"])  # type: ignore
-                cached_member = self.guild and self.guild.get_member(user_id)
+                cached_member = self.guild and self.guild.get_member(int(member["user"]["id"]))  # type: ignore
                 self.user = cached_member or Member(state=self._state, guild=guild, data=member)  # type: ignore
                 self._permissions = int(member.get("permissions", 0))
         else:
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -211,7 +211,7 @@ class Interaction:
 
         # TODO: there's a potential data loss here
         if self.guild_id:
-            guild = self.guild or Object(int(self.guild_id))
+            guild = self.guild or Object(id=self.guild_id)
             try:
                 member = data["member"]
             except KeyError:
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -199,7 +199,10 @@ class Interaction:
 
         self.message: Optional[Message]
         try:
-            self.message = Message(state=self._state, channel=self.channel, data=data["message"])  # type: ignore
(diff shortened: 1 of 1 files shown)


DenverCoder1:
If not all intents are enabled, there may be some events that don't update the cached message I think.

Is it possible the interaction message will contain data that is newer/more accurate than the cache?

nextcord/interactions.py
```diff
@@ -214,11 +217,15 @@ def _from_data(self, data: InteractionPayload):
             except KeyError:
                 pass
             else:
-                self.user = Member(state=self._state, guild=guild, data=member)  # type: ignore
+                cached_member = self.guild and self.guild.get_member(int(member["user"]["id"]))  # type: ignore # user key should be present here
+                self.user = cached_member or Member(state=self._state, guild=guild, data=member)  # type: ignore # user key should be present here
```
EmreTech, line 221: I don't think it's necessary to duplicate this comment twice

EmreTech (approved):
Looks good

nextcord/interactions.py
```diff
@@ -199,7 +199,10 @@ def _from_data(self, data: InteractionPayload):
 
         self.message: Optional[Message]
         try:
-            self.message = Message(state=self._state, channel=self.channel, data=data["message"])  # type: ignore
+            message = data["message"]
+            self.message = self._state._get_message(int(message["id"])) or Message(
+                state=self._state, channel=self.channel, data=message  # type: ignore
+            )
```
DenverCoder1, line 205:
```suggestion
            message = data["message"]
            cached_message = self._state._get_message(int(message["id"]))
            if cached_message is not None:
                cached_message._update(message)
                self.message = cached_message
            else:
                self.message = Message(state=self._state, channel=self.channel, data=message)  # type: ignore
```

nextcord/interactions.py
DenverCoder1, line 221:
```suggestion
                cached_member = self.guild and self.guild.get_member(int(member["user"]["id"]))  # type: ignore # user key should be present here
                if cached_member is not None:
                    cached_member._update(member)
                    self.user = cached_member
                else:
                    self.user = Member(state=self._state, guild=guild, data=member)  # type: ignore
```

nextcord/interactions.py
```diff
@@ -214,11 +217,15 @@ def _from_data(self, data: InteractionPayload):
...
+                cached_member = self.guild and self.guild.get_member(int(member["user"]["id"]))  # type: ignore # user key should be present here
+                self.user = cached_member or Member(state=self._state, guild=guild, data=member)  # type: ignore # user key should be present here
                 self._permissions = int(member.get("permissions", 0))
         else:
             try:
-                self.user = User(state=self._state, data=data["user"])
+                user = data["user"]
+                self.user = self._state.get_user(int(user["id"])) or User(
+                    state=self._state, data=user
+                )
```
DenverCoder1, line 228:
```suggestion
                user = data["user"]
                cached_user = self._state.get_user(int(user["id"]))
                if cached_user is not None:
                    cached_user._update(user)
                    self.user = cached_user
                else:
                    self.user = User(state=self._state, data=user)
```

DenverCoder1 (commented):
Let me know what you think of these suggestions. This should use the cache, but also update any additional fields provided by the interaction data.

Soheab:
> If not all intents are enabled, there may be some events that don't update the cached message I think.
> 
> Is it possible the interaction message will contain data that is newer/more accurate than the cache?
> 
> Possibly they may need to be merged together?



> If not all intents are enabled, there may be some events that don't update the cached message I think.
> 
> Is it possible the interaction message will contain data that is newer/more accurate than the cache?
> 
> Possibly they may need to be merged together?


~~keep forgetting to respond, sorry~~
But no, I haven't experienced a situation where the Interaction Message was missing information or was out of date, mainly because it's not really used since Interaction has most of the stuff 

I just did it for the sake of "why not".


> If not all intents are enabled, there may be some events that don't update the cached message I think.

Same as above, I don't know but if that's true but doesn't sound good and a potential dataloss.

----

**Response to the suggestions:**
They look good to me, will of course require testing. Will do that when I can, thank you :)

DenverCoder1:
> But no, I haven't experienced a situation where the Interaction Message was missing information or was out of date, mainly because it's not really used since Interaction has most of the stuff

I believe you are correct since if the member/message is cached at all, it should always update in the corresponding update/edit event, so just using the item from the cache should be enough.

DenverCoder1 (approved):

ooliver1 (approved):

PR #732, opened 2022-07-04
feat(app-cmds): support min/max length for string options

EmreTech:
## Summary

This adds support for the newly added `min_length` and `max_length` fields in Slash Command options. 
More information about this: https://github.com/discord/discord-api-docs/pull/5143

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -235,8 +235,12 @@ class ApplicationCommandOption:
         Minimum value the user can input. Should only be set if this is an integer or number option.
     min_length: :class:`int`
         Minimum length of a string the user can input. Should only be set if this is a string option.
+
+        .. versionadded:: 2.1
     max_length: :class:`int`
         Maximum length of a string the user can input. Should only be set if this is a string option.
+
+        .. versionadded:: 2.1
     autocomplete: :class:`bool`
         If the command option should have autocomplete enabled.
     """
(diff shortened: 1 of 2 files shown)


nextcord/application_command.py
```diff
@@ -1378,6 +1404,30 @@ def verify(self) -> bool:
...
+        # see here: https://github.com/discord/discord-api-docs/issues/5149
+        if (
+            self.min_length is not None and self.max_length is not None
+        ) and self.min_length > self.max_length:
+            raise ValueError("min_length must be less than or equal to max_length")
+
+        if (self.min_length is not None or self.max_length is not None) and self.type is not (
+            ApplicationCommandOptionType.string
+        ):
+            raise ValueError("min_length or max_length can only be set if the type is a string.")
```
DenverCoder1, line 1429: I believe for the most part Nextcord is avoiding adding manual constraint checks and letting Discord HTTP exceptions handle most issues.

DenverCoder1 (approved):
Besides the comment (should probably be discussed with Epic and/or ooliver), this looks good to me. I have not tested yet.

nextcord/application_command.py
EmreTech: I know, but the if statement with the comments is necessary because of that issue with the Discord API
When the issue is ever resolved then I could remove that check, but for now, that won't be the case

DenverCoder1: If Discord allows it, does it need to be checked? While it might not function in a friendly way, the developer should be able to notice and correct the mistake while testing.

I think as a API wrapper, it should allow passing to Discord any values you want even if it doesn't make sense.

I'm open to hearing additional opinions, though.

DenverCoder1 (approved):
Looks great, tested.

DenverCoder1:
`_ApplicationCommandOptionOptional` in `nextcord.types` might need to be updated

Skelmis (approved):
untested, lgtm

ooliver1 (changes requested):
could some `.. versionadded:: 2.1`s be added to the added parameters/attributes

ooliver1 (approved):

PR #733, opened 2022-07-06
feat(ClientCog): remove ClientCog.to_register

EmreTech:
## Summary

This PR removes the deprecated `to_register` property from ClientCog.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -531,16 +531,6 @@ class ClientCog:
         """
         pass
 
-    @property
-    def to_register(self) -> List[BaseApplicationCommand]:
-        warnings.warn(
-            ".to_register is deprecated, please use .application_commands instead.",
-            stacklevel=2,
-            category=FutureWarning,
-        )
-        # TODO: Remove at later date.
-        return self.__cog_application_commands__
-
 
 class CallbackMixin:
     name: Optional[str]


TAG-Epic:
This cant be done rn due to it being breaking

Skelmis:
> This cant be done rn due to it being breaking

In my eyes this wouldn't be breaking? It was deprecated in the slash refactor, and would be slated for 2.1. isn't that enough of a deprecation period?

TAG-Epic:
> > This cant be done rn due to it being breaking
> 
> In my eyes this wouldn't be breaking? It was deprecated in the slash refactor, and would be slated for 2.1. isn't that enough of a deprecation period?

Major changes should be done when bumping the major version

Skelmis:
> > > This cant be done rn due to it being breaking
> > 
> > In my eyes this wouldn't be breaking? It was deprecated in the slash refactor, and would be slated for 2.1. isn't that enough of a deprecation period?
> 
> Major changes should be done when bumping the major version

A) so we can never remove anything before 3.0?
B) As seen [here](https://docs.nextcord.dev/en/stable/version_guarantees.html#version-guarantees), undocumented features are not considered a part of the public api and subject to change. The attribute also did not appear when searched, which further indicates it is not part of the public api to me

TAG-Epic:
A) Yeah no major changes before 3.0
B) Ah should be fine then I guess

TAG-Epic (approved):

Skelmis (approved):

ooliver1:
> > This cant be done rn due to it being breaking
> 
> In my eyes this wouldn't be breaking? It was deprecated in the slash refactor, and would be slated for 2.1. isn't that enough of a deprecation period?

i personally think the version guarantees docs should be edited before the 2.1 release to state we use a python-like semver tweak where minor can have breaking changes

PR #735, opened 2022-07-08
fix(commands): use AutoShardedClient initialiser

ooliver1:
## Summary

if `AutoShardedBot` was used, `Client` would still be initialised
this explicitly uses the correct superclass client in the relevant bot classes

## Checklist

- [ ] This PR fixes an issue.

## Issues

`shard_ids` is not passed to (`AutoSharded`)`Client` as it is not present in both, not sure what to do about that


nextcord/ext/commands/bot.py
```diff
@@ -40,6 +40,7 @@
     Any,
     Awaitable,
     Callable,
+    ClassVar,
```
DenverCoder1, line 43: No longer used?
```suggestion
```

nextcord/ext/commands/bot.py
```diff
@@ -1435,12 +1384,164 @@ class Bot(BotBase, nextcord.Client):
...
+        enable_debug_events: bool = False,
+        loop: Optional[asyncio.AbstractEventLoop] = None,
+        lazy_load_commands: bool = True,
+        rollout_associate_known: bool = True,
+        rollout_delete_unknown: bool = True,
+        rollout_register_new: bool = True,
+        rollout_update_known: bool = True,
+        rollout_all_guilds: bool = False,
+        owner_id: Optional[int] = None,
+        owner_ids: Iterable[int] = set(),
```
DenverCoder1, line 1505: Using `set()` in a default is discouraged since it would be shared across multiple calls if modified. Not a huge problem in this context, but using `None` and setting it to `set()` within the function would be recommended.

https://rules.sonarsource.com/python/RSPEC-5717

nextcord/ext/commands/bot.py
```diff
@@ -1435,12 +1384,164 @@ class Bot(BotBase, nextcord.Client):
...
+            rollout_associate_known=rollout_associate_known,
+            rollout_delete_unknown=rollout_delete_unknown,
+            rollout_register_new=rollout_register_new,
+            rollout_update_known=rollout_update_known,
+            rollout_all_guilds=rollout_all_guilds,
+        )
+
+        BotBase.__init__(
+            self,
+            command_prefix=command_prefix,  # type: ignore
```
DenverCoder1, line 1540: Maybe to avoid needing `# type: ignore`, it could use `Union[Bot, AutoShardedBot]` instead of `Self` in the type for `command_prefix`?

(similar to `when_mentioned`/`when_mentioned_or`)

DenverCoder1 (approved):
LGTM. A lot of testing is needed before 2.1

ooliver1:
This also manages to fix another broken issue where `Client.__init__` is never called properly and instead gets unexpected options.

Skelmis (approved):
untested, lgtm

PR #736, opened 2022-07-08
feat(app-cmds): give descriptive error message when params are missing

### LINKED ISSUE #420: Non-descriptive error in Slash when forgetting self
### Summary

Making a slash command in a cog and forgetting to put the "self" argument in the command description results in a ~~strange error.~~ better error since #530, but still isn't the greatest.

The core library

### The Problem

Making a slash command in a cog and forgetting to put the "self" argument in the command description results in a strange error. While erroring out is correct behavior, the error that it provides should be a bit better.

### The Ideal Solution

Find a way, preferably before runtime, to figure out if `self` is possibly missing and raise an error if it is missing.

### The Current Solution

_No response_

EmreTech:
## Summary

This PR gives a descriptive error message when you forget parameters for a Slash Command. Fixes #420.

Current Issues:
* ~~This assumes that the self parameter will not be type hinted~~ added in 96b68b6
* This assumes that if there is no self or interaction parameter then the next two options (if there) are type hinted

Notes:
* This implementation could be moved to the `__init__` of CallbackMixin but there's also a chance that the parent cog won't be set

## Checklist

- [x] This PR fixes an issue.


nextcord/application_command.py
```diff
@@ -694,12 +694,38 @@ def from_callback(
...
+                    set(map(lambda param: param.name, callback_params.values())) - possible_options
+                )
+
+                if self.parent_cog is not None and len(intersection) < 2:
+                    raise ValueError(
+                        f"Callback {self.error_name} is missing the self and/or interaction parameters. Please double check your function define."
+                    )
+                elif len(intersection) < 1:
+                    raise ValueError(
+                        f"Callback {self.error_name} is missing the interaction parameter. Please double check your function define."
```
Skelmis, line 717: definition

nextcord/application_command.py
```diff
@@ -694,12 +694,38 @@ def from_callback(
...
+
+                    possible_options.add(param.name)
+
+                intersection = (
+                    set(map(lambda param: param.name, callback_params.values())) - possible_options
+                )
+
+                if self.parent_cog is not None and len(intersection) < 2:
+                    raise ValueError(
+                        f"Callback {self.error_name} is missing the self and/or interaction parameters. Please double check your function define."
```
Skelmis, line 713: definition

Skelmis (changes requested):

nextcord/application_command.py
EmreTech: Elaborate.

nextcord/application_command.py
EmreTech: Elaborate.

ooliver1: > "... define."
definition

nextcord/application_command.py
ooliver1: > "... define."

definition

EmreTech: Oh I see now. Apologies.

nextcord/application_command.py
```diff
@@ -694,12 +694,38 @@ def from_callback(
             if option_class:
                 first_arg = True
                 typehints = typing.get_type_hints(self.callback)
+                callback_params = signature(self.callback).parameters
+                possible_options: Set[str] = set()
+
+                for _, param in callback_params.items():
+                    if param.annotation is param.empty or param.annotation is Interaction:
```
ooliver1, line 701: surely this also skips options that have no type - this is supported as it defaults to `str`

nextcord/application_command.py
```diff
@@ -694,12 +694,38 @@ def from_callback(
             if option_class:
                 first_arg = True
                 typehints = typing.get_type_hints(self.callback)
+                callback_params = signature(self.callback).parameters
+                possible_options: Set[str] = set()
+
+                for _, param in callback_params.items():
```
ooliver1, line 700:
```suggestion
                for param in callback_params.values():
```

ooliver1 (changes requested):

nextcord/application_command.py
EmreTech: I was thinking about that, but no it doesn't support those
There isn't a good way to distinguish between the two without doing name checking (which is something I want to avoid)

ooliver1: could it not at least check if its the first parameter when its empty

EmreTech: Yeah good point, I'll look into that

nextcord/application_command.py
```diff
@@ -694,12 +694,40 @@ def from_callback(
             if option_class:
                 first_arg = True
                 typehints = typing.get_type_hints(self.callback)
+                callback_params = signature(self.callback).parameters
+                possible_options: Set[str] = set()
+
+                for i, param in enumerate(callback_params.values()):
+                    if (
+                        param.annotation is param.empty or param.annotation is Interaction
```
Skelmis, line 702: This should use `issubclass` as well as `is` for people who roll with a custom `Interaction` object. (It happens, we should support it)

```py
                    if (
                        param.annotation is param.empty or param.annotation is Interaction or issubclass(param.annotation, Interaction)
                    ) and i in (0, 1):
                        # self or interaction parameter, ignore
                        continue

                    possible_options.add(param.name)
```

Resolves the tests for me locally

Skelmis (changes requested):
Tests relevant to this pr can be found [here](https://nexttest.koldfusion.xyz/nextcord/736), my review comment should resolve the

nextcord/application_command.py
EmreTech: Advice for next time: you could put a suggestion in that review comment so I could apply the changes immediately

Skelmis (approved):
[Tested](https://nexttest.koldfusion.xyz/nextcord/736), looks all good to me.

DenverCoder1:
I know it's not common, but what would happen if someone were to type-annotate self with a `TypeVar`, a future annotation of the current class, or [`typing.Self`](https://peps.python.org/pep-0673/)?

All of those should be legal.

EmreTech:
> I know it's not common, but what would happen if someone were to type-annotate self with a `TypeVar`, a future annotation of the current class, or [`typing.Self`](https://peps.python.org/pep-0673/)?
> 
> All of those should be legal.

Yes I know
I'll add support for those soon
That's why it's in the "current issues" section in the description of this PR

nextcord/application_command.py
```diff
@@ -694,12 +694,42 @@ def from_callback(
...
+                        or issubclass(param.annotation, Interaction)
+                    ) and i in (0, 1):
+                        # self or interaction parameter, ignore
+                        continue
+
+                    possible_options.add(param.name)
+
+                intersection = (
+                    set(map(lambda param: param.name, callback_params.values())) - possible_options
+                )
```
teaishealthy, line 713:
```suggestion
                intersection = {param.name for param in callback_params.values()} - possible_options
```

nextcord/application_command.py
```diff
@@ -694,12 +694,42 @@ def from_callback(
...
                 first_arg = True
                 typehints = typing.get_type_hints(self.callback)
+                callback_params = signature(self.callback).parameters
+                possible_options: Set[str] = set()
+
+                for i, param in enumerate(callback_params.values()):
+                    if (
+                        param.annotation is param.empty
+                        or param.annotation is Interaction
+                        or issubclass(param.annotation, Interaction)
```
teaishealthy, line 704:
```suggestion
                        param.annotation is param.empty
                        or issubclass(param.annotation, Interaction)
```

```py
>>> class A: pass
... 
>>> issubclass(A, A)
True
```

nextcord/application_command.py
```diff
@@ -694,12 +694,42 @@ def from_callback(
...
+                    ) and i in (0, 1):
+                        # self or interaction parameter, ignore
+                        continue
```
teaishealthy, line 707: Why the `continue`, negate the condition and move the actual loop body into the condition body?

teaishealthy (changes requested):

nextcord/application_command.py
EmreTech: I would not suggest changes that go against pre-commit fixes
That change wasn't me, it was pre-commit

teaishealthy (approved):

nextcord/application_command.py
```diff
@@ -745,12 +745,50 @@ def from_callback(
...
+                            param.annotation, Interaction
+                        )  # will always be an interaction parameter
+                        # TODO: use typing.Self when 3.11 is standard
+                        or param.annotation
+                        is typing_extensions.Self  # will always be a self parameter
+                        or isinstance(
+                            param.annotation, typing.TypeVar
+                        )  # will always be a self parameter
+                    ) and i in (0, 1):
+                        possible_options.add(param.name)
```
DenverCoder1, line 764: It seems an error came about through the addition of a `not` in [`6fb2131`](https://github.com/nextcord/nextcord/pull/736/commits/6fb213165f981cfc8524f21cda3e1ad07f9db0b1).

The negation was added only on the *first* part of the condition which would make it so only the first two parameters could be added as possible options.

Here are **two** suggestions that could fix this issue. **Note:** the second suggestion is a simplified version and should be preferred over the first.

Here is what it would look like with the `not` added to outside *both* conditions instead of the first, however this can be simplified (see below)

```py
# Only add params that are not self or Interaction
if not (
    (
        # could be a self or interaction parameter
        param.annotation is param.empty
        # will always be an interaction parameter
        or issubclass(param.annotation, Interaction)
        # will always be a self parameter
        # TODO: use typing.Self when 3.11 is standard
        or param.annotation is typing_extensions.Self
        # will always be a self parameter
        or isinstance(param.annotation, typing.TypeVar)
    )
    and i in (0, 1)
):
    possible_options.add(param.name)
```

Using de-morgan simplification, we can distribute the negation:

```py
# Only add params that are not self or Interaction
if (
    # could be a self or interaction parameter
    param.annotation is not param.empty
    # will always be an interaction parameter
    and not issubclass(param.annotation, Interaction)
    # will always be a self parameter
    # TODO: use typing.Self when 3.11 is standard
    and param.annotation is not typing_extensions.Self
    # will always be a self parameter
    and not isinstance(param.annotation, typing.TypeVar)
) or i not in (0, 1):
    possible_options.add(param.name)
```

```suggestion
                    # Only add params that are not self or Interaction
                    if (
                        # could be a self or interaction parameter
                        param.annotation is not param.empty
                        # will always be an interaction parameter
                        and not issubclass(param.annotation, Interaction)
                        # will always be a self parameter
                        # TODO: use typing.Self when 3.11 is standard
                        and param.annotation is not typing_extensions.Self
                        # will always be a self parameter
                        and not isinstance(param.annotation, typing.TypeVar)
                    ) or i not in (0, 1):
                        possible_options.add(param.name)
```

nextcord/application_command.py
```diff
@@ -745,12 +745,50 @@ def from_callback(
...
+
+                intersection = {param.name for param in callback_params.values()} - possible_options
```
DenverCoder1, line 766: This variable should probably be renamed since the term "intersection" means the things in common, but this is more like the opposite since it is finding the set difference (in this case, the items that are *not* in common between the two sets). Therefore, I think `difference` would be a more appropriate variable name. Or perhaps something more descriptive.

nextcord/application_command.py
```diff
@@ -745,12 +745,50 @@ def from_callback(
...
+                    raise ValueError(
+                        f"Callback {self.error_name} is missing the self and/or interaction parameters. Please double check your function definition."
+                    )
+                elif len(intersection) < 1:
+                    raise ValueError(
+                        f"Callback {self.error_name} is missing the interaction parameter. Please double check your function definition."
+                    )
+
+                # not needed later on
+                del possible_options, intersection
```
DenverCoder1, line 778: I don't really see the point of deleting variables here, I think Python handles freeing local variables automatically when the function finishes and these aren't taking up a lot of memory.

nextcord/application_command.py
```diff
@@ -745,12 +745,50 @@ def from_callback(
...
-                if self.parent_cog:
-                    skip_counter += 1
+                self_skip = bool(self.parent_cog)
+                first_arg = True
+                for name, param in callback_params.items():
+                    if first_arg:
+                        if not self_skip:
+                            first_arg = False
+                        else:
+                            self_skip = False
```
DenverCoder1, line 791: I'm not sure I see what this is doing. It appears `first_arg` and `self_skip` are not referenced anywhere else in the rest of this function, so it seems this code is unused.

DenverCoder1 (changes requested):

nextcord/application_command.py
EmreTech: I don't really like the "distributing the negative" solution, so I'll go with the first one

nextcord/application_command.py
EmreTech: I thought this was old code? Not sure
I'll remove it anyways

nextcord/application_command.py
EmreTech: Nevermind

nextcord/application_command.py
```diff
@@ -745,12 +745,35 @@ def from_callback(
...
+                            param.annotation, Interaction
+                        )  # will always be an interaction parameter
+                        # TODO: use typing.Self when 3.11 is standard
+                        or param.annotation
+                        is not typing_extensions.Self  # will always be a self parameter
+                        or not isinstance(
+                            param.annotation, typing.TypeVar
+                        )  # will always be a self parameter
+                    ) and i not in (0, 1):
+                        possible_options.add(param.name)
```
DenverCoder1, line 765: You have an `and` instead of `or` on all the conditions which is supposed to change when distributing a negation (De Morgan's Law).

Also, I don't really like the way Black breaks the lines here when the comments go at the end.

It would look nicer in my opinion for the comments to be moved to make the lines shorter (as done in the previous suggestion). Debatable, I suppose.

```suggestion
                    # Only add params that are not self or Interaction
                    if (
                        # could be a self or interaction parameter
                        param.annotation is not param.empty
                        # will always be an interaction parameter
                        and not issubclass(param.annotation, Interaction)
                        # will always be a self parameter
                        # TODO: use typing.Self when 3.11 is standard
                        and param.annotation is not typing_extensions.Self
                        # will always be a self parameter
                        and not isinstance(param.annotation, typing.TypeVar)
                    ) or i not in (0, 1):
                        possible_options.add(param.name)
```

nextcord/application_command.py
```diff
@@ -745,12 +745,35 @@ def from_callback(
                 skip_counter = 1
                 # TODO: use typing.get_type_hints when 3.9 is standard
                 typehints = typing_extensions.get_type_hints(self.callback, include_extras=True)
-                # Getting the callback with `self_skip = inspect.ismethod(self.callback)` was problematic due to the
-                #  decorator going into effect before the class is instantiated, thus being a function at the time.
-                #  Try to look into fixing that in the future?
-                #  If self.parent_cog isn't reliable enough, we can possibly check if the first parameter name is `self`
-                if self.parent_cog:
-                    skip_counter += 1
```
DenverCoder1, line 753: I think skip_counter should still be incremented if self is used, doesn't seem like this part of the code was affected.

DenverCoder1 (changes requested):

nextcord/application_command.py
```diff
@@ -743,15 +743,43 @@ def from_callback(
...
+                difference = {param.name for param in callback_params.values()} - possible_options
+
+                if self.parent_cog is not None and len(difference) < 2:
+                    raise ValueError(
+                        f"Callback {self.error_name} is missing the self and/or interaction parameters. Please double check your function definition."
+                    )
+                elif len(difference) < 1:
+                    raise ValueError(
+                        f"Callback {self.error_name} is missing the interaction parameter. Please double check your function definition."
+                    )
```
DenverCoder1, line 781: Since `difference` is only used for its length, it may be more readable to instead calculate it as a number.

If `difference` is not calculated using set difference, `possible_options` also does not need to be a set.

By using this information, it should be possible to instead calculate the number of parameters that aren't necessarily options and just use a single number variable.

I'm thinking it could be simplified like this. This should be tested, but I believe this is equivalent and eliminates the need for the for-loop.

```py
callback_params = signature(self.callback).parameters

# Count the number of parameters that are not necessarily command options
# Note, this count may include parameters that are used as command options
num_non_option_params = sum(
    # could be a self or interaction parameter
    param.annotation is param.empty
    # will always be an interaction parameter
    or issubclass(param.annotation, Interaction)
    # will always be a self parameter
    # TODO: use typing.Self when 3.11 is standard
    or param.annotation is typing_extensions.Self
    # will always be a self parameter
    or isinstance(param.annotation, typing.TypeVar)
    for param in list(signature(self.callback).parameters.values())[:2]
)

if self.parent_cog is not None and num_non_option_params < 2:
    raise ValueError(
        f"Callback {self.error_name} is missing the self and/or interaction parameters. Please double check your function definition."
    )
elif num_non_option_params < 1:
    raise ValueError(
        f"Callback {self.error_name} is missing the interaction parameter. Please double check your function definition."
    )
```

Note: it may even be possible to use `[:skip_counter]` instead of `[:2]` since you probably don't need to check the first *two* parameters when it is not in a cog.

DenverCoder1 (approved):
It looks pretty good. It looks like it should work, I'll probably want to test it to be sure though.

nextcord/application_command.py
```diff
@@ -743,16 +743,39 @@ def from_callback(
...
+                    param.annotation is not param.empty
+                    # will always be an interaction parameter
+                    and not issubclass(param.annotation, Interaction)
+                    # will always be a self parameter
+                    # TODO: use typing.Self when 3.11 is standard
+                    and param.annotation is not typing_extensions.Self
+                    # will always be a self parameter
+                    and not isinstance(param.annotation, typing.TypeVar)
+                    for param in list(callback_params.values())[:skip_counter]
+                )
```
DenverCoder1, line 767: Sorry for changing it so much, but when testing it I noticed it failing on some cases and it seems like it actually should be negated again (remove the `not`s and change `and` back to `or`).

That way it will count how many parameters *could* be self or interaction.

```suggestion
                non_option_params = sum(
                    # could be a self or interaction parameter
                    param.annotation is param.empty
                    # will always be an interaction parameter
                    or issubclass(param.annotation, Interaction)
                    # will always be a self parameter
                    # TODO: use typing.Self when 3.11 is standard
                    or param.annotation is typing_extensions.Self
                    # will always be a self parameter
                    or isinstance(param.annotation, typing.TypeVar)
                    for param in list(callback_params.values())[:skip_counter]
                )
```

DenverCoder1 (approved):

ooliver1 (approved):

PR #738, opened 2022-07-09
docs(version_guarantees): add minor may break

ooliver1:
## Summary

Minor version bumps can actually contain breaking changes, this lists in the guarantees that this may be the case.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/version_guarantees.rst
+++ b/docs/version_guarantees.rst
@@ -3,9 +3,25 @@
 Version Guarantees
 ==================
 
-The library follows a `semantic versioning principle <https://semver.org/>`_ which means that the major version is updated every time there is an incompatible API change. However due to the lack of guarantees on the Discord side when it comes to breaking changes along with the fairly dynamic nature of Python it can be hard to discern what can be considered a breaking change and what isn't.
+Nextcord does not follow semantic versioning exactly, The main difference is that breaking changes may also be made on minor version bumps (similar to Python's scheme).
+The reason for this is the Discord API lacks guarantees on breaking changes per API version, so library breaking changes unfortunately have to be made to match.
 
-The first thing to keep in mind is that breaking changes only apply to **publicly documented functions and classes**. If it's not listed in the documentation here then it is not part of the public API and is thus bound to change. This includes attributes that start with an underscore or functions without underscores that are not documented.
+However, any breaking changes will always be marked as such in the release notes/changelogs.
+
+Summary of The Versioning Scheme
+--------------------------------
+
+The version scheme ``major.minor.micro`` used aims to follow these rules:
+
+- ``major`` bumps usually contain large refactors/changes (most likely containing breaking changes).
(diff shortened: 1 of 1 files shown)


docs/version_guarantees.rst
```diff
@@ -3,9 +3,25 @@
 Version Guarantees
 ==================
 
-The library follows a `semantic versioning principle <https://semver.org/>`_ which means that the major version is updated every time there is an incompatible API change. However due to the lack of guarantees on the Discord side when it comes to breaking changes along with the fairly dynamic nature of Python it can be hard to discern what can be considered a breaking change and what isn't.
+Nextcord does not follow semantic versioning exactly, The main difference is that breaking changes may also be made on minor version bumps (similar to Python's scheme).
```
Skelmis, line 6: Lowercase the `t`

Skelmis (changes requested):

Skelmis (approved):

EmreTech (approved):

TAG-Epic:
Semver should be followed.

EmreTech:
> Semver should be followed.

Okay then. Get Discord to somewhat follow Semver with their API, then we'll talk.

ooliver1:
> Semver should be followed.

This change is so we are not limited to changing the major version every time there is a breaking change on Discord's API - which is pretty common. This reflects how discord.py actually handled such as [discord.py v1.5.0](https://discordpy.readthedocs.io/en/latest/whats_new.html#v1-5-0) when gateway intents were released.

DenverCoder1 (approved):
Sounds good to me.

(Using actual semver, the main repo would be on v4 before nc v3 is released 😃)

ooliver1:
> > Sounds good to me.
> > (Using actual semver, the main repo would be on v4 before nc v3 is released 😃)
> 
> Yeah this has been a internal topic for a while now, I dont want to block NCv2 from bumping the major version due to bad naming.

I don’t know what internal conversations this is about, because I don't think it's involved anyone.

ooliver1:
> Sounds good to me.
> 
> (Using actual semver, the main repo would be on v4 before nc v3 is released smiley)
> 
> **Updated with discussion from Discord**
> 
> If it's a breaking change that will affect very few users I don't think it needs a major release.
> 
> For something that many people would need to fix, a major would make sense.
> 
> One thing that's common for dependencies is a locked major version (eg. `nextcord>=2.0.0,<3`). That way the code is less likely to break from automatic dependency updates. So that's one reason to make all breaking changes major releases.
> 
> Some "breaking changes" are a bit of a stretch though and require making up an unlikely scenario, and for those, a major version change is not needed in my opinion.
> 
> Tbh it's not critical to go one way or the other as long as changes are clearly documented in the release notes.

Good words on the updates, we will go with that thank you. Time will see on what changes will require a major and what is a stretch.

PR #744, opened 2022-07-18
feat: support for lambdas in slash options

EmreTech:
## Summary

This PR allows for the user to set the default in `SlashOption` to a lambda. This lambda can be used to set the default of an option to something from the interaction. Supersedes #708. 

For example: 
```python
import nextcord

client = nextcord.Client()

@client.slash_command()
async def greet(inter: nextcord.Interaction, target: nextcord.User = nextcord.SlashOption(default=lambda inter: inter.user)):
    await inter.response.send_message(f"Yo {target.display_name}!")

client.run("token")
```
### Current Issues

- Pyright complains that the type of the parameter in the Slash Option lambda (if provided) is unknown. This could possibly be fixed by changing how we type hint `default`.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -1542,7 +1542,7 @@ class SlashCommandMixin(CallbackMixin):
                     raise ValueError(f"default lambda has to take in one parameter, not {len_params}!")
                 default = default(interaction)
 
-            kwargs[uncalled_arg.functional_name] = uncalled_arg.default
+            kwargs[uncalled_arg.functional_name] = default
 
         return kwargs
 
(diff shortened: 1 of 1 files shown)


DenverCoder1:
I don't really see the need for this kind of thing. Can't you just let it default to None and handle the conversion within the command itself?

EmreTech:
> I don't really see the need for this kind of thing. Can't you just let it default to None and handle the conversion within the command itself?
> 
> The most pythonic way of doing this would look like this. Easier to understand in my opinion.
> 
> ```python
> @client.slash_command()
> async def greet(inter: nextcord.Interaction, target: Optional[nextcord.Member]):
>     member = target or interaction.user
>     await inter.response.send_message(f"Yo {member.display_name}!")
> ```

Alright fine. Closing this PR.
But also this same argument could be made for #708

DenverCoder1:
I think it was discussed on Discord before, but I guess I didn't comment on that PR

PR #752, opened 2022-07-24
feat: add File.force_close to close bytesio files

TrapDrap:
## Summary

Added force_close to `File` class that will prevent memory leak to occur when a message containing a file is sent if enabled.
`nextcord.abc` calls `File.close()` for each file after the message is sent.
Files bytes could not be deleted after `File.close()` is called.
Causing uploaded files to linger in memory until the garbage collector removes them, usually after a few hours.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/asset.py
+++ b/nextcord/asset.py
@@ -171,7 +171,11 @@ class AssetMixin:
         data = await self.read()
         file_filename = filename if filename is not MISSING else yarl.URL(self.url).name
         return File(
-            io.BytesIO(data), filename=file_filename, description=description, spoiler=spoiler, force_close=force_close
+            io.BytesIO(data),
+            filename=file_filename,
+            description=description,
+            spoiler=spoiler,
+            force_close=force_close,
         )
 
 
--- a/nextcord/asset.py
+++ b/nextcord/asset.py
@@ -149,6 +149,8 @@ class AssetMixin:
             memory after it is sent or used once.
             Keep this enabled if you don't wish to reuse the same bytes.
 
+            .. versionadded:: 2.2
+
         Raises
         ------
         DiscordException
--- a/nextcord/asset.py
+++ b/nextcord/asset.py
@@ -124,6 +124,7 @@ class AssetMixin:
         filename: Optional[str] = MISSING,
         description: Optional[str] = None,
         spoiler: bool = False,
+        force_close: bool = True,
     ) -> File:
         """|coro|
 
@@ -141,6 +142,12 @@ class AssetMixin:
             The description for the file.
         spoiler: :class:`bool`
             Whether the file is a spoiler.
+        force_close: :class:`bool`
+            Whether to forcibly close the bytes used to create the file
+            when ``.close()`` is called.
+            This will also make the file bytes unusable by flusing it from
+            memory after it is sent or used once.
+            Keep this enabled if you don't wish to reuse the same bytes.
 
         Raises
         ------
@@ -164,7 +171,7 @@ class AssetMixin:
         data = await self.read()
         file_filename = filename if filename is not MISSING else yarl.URL(self.url).name
         return File(
-            io.BytesIO(data), filename=file_filename, description=description, spoiler=spoiler
+            io.BytesIO(data), filename=file_filename, description=description, spoiler=spoiler, force_close=force_close
         )
 
 
--- a/nextcord/asset.py
+++ b/nextcord/asset.py
@@ -145,7 +145,7 @@ class AssetMixin:
         force_close: :class:`bool`
             Whether to forcibly close the bytes used to create the file
             when ``.close()`` is called.
-            This will also make the file bytes unusable by flusing it from
+            This will also make the file bytes unusable by flushing it from
             memory after it is sent or used once.
             Keep this enabled if you don't wish to reuse the same bytes.
 
--- a/nextcord/file.py
+++ b/nextcord/file.py
@@ -167,7 +167,6 @@ class File:
             self.fp.seek(self._original_pos)
 
     def close(self) -> None:
+        self.fp.close = self._closer
         if self._owner or self.force_close:
             self._closer()
-        else:
-            self.fp.close = self._closer
--- a/nextcord/file.py
+++ b/nextcord/file.py
@@ -68,7 +68,7 @@ class File:
         when ``.close()`` is called.
         This will also make the file bytes unusable by flushing it from
         memory after it is sent once.
-        Keep this enabled if you don't wish to reuse the same bytes.
+        Enable this if you don't wish to reuse the same bytes.
 
     Attributes
     -----------
--- a/nextcord/file.py
+++ b/nextcord/file.py
@@ -92,7 +92,16 @@ class File:
         Keep this enabled if you don't wish to reuse the same bytes.
     """
 
-    __slots__ = ("fp", "filename", "spoiler", "force_close", "_original_pos", "_owner", "_closer", "description")
+    __slots__ = (
+        "fp",
+        "filename",
+        "spoiler",
+        "force_close",
+        "_original_pos",
+        "_owner",
+        "_closer",
+        "description",
+    )
 
     if TYPE_CHECKING:
         fp: Union[io.BufferedReader, io.BufferedIOBase]
@@ -108,7 +117,7 @@ class File:
         *,
         description: Optional[str] = None,
         spoiler: bool = False,
-        force_close: bool = True, # Accessed in .close()
+        force_close: bool = True,  # Accessed in .close()
     ):
         if isinstance(fp, io.IOBase):
             if not (fp.seekable() and fp.readable()):
--- a/nextcord/file.py
+++ b/nextcord/file.py
@@ -129,7 +129,7 @@ class File:
             self.fp = open(fp, "rb")
             self._original_pos = 0
             self._owner = True
-        
+
         self.force_close = force_close
 
         # aiohttp only uses two methods from IOBase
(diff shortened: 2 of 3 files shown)


Skelmis (approved):
Untested, however looks good to me

EmreTech (approved):

DenverCoder1:
I feel like the `_owner` check must have been there for some reason.

Are you sure this is not reverting work done in https://github.com/yagomichalak/discord.py/commit/f55cd18b494a6044666357e4f1daa3ebcb5e4afe to fix https://github.com/Rapptz/discord.py/issues/1809?

Maybe a 2-step close on your end would avoid potentially causing issues with it being closed when it is not supposed to?
```py
file.close()
file.fp.close()
```

(I haven't tested it yet, so I'm still not sure if the check is required or not.)

TrapDrap:
> 

It shouldn't.
aiohttp shouldn't call the `.close()` method from of the File class, it calls `.fp.close()` from the BufferIO/BytesIO object, which was stubbed in line 114 & 115.
I did not remove the stub that revoked the ability for aiohttp to close the data.

DenverCoder1:
After calling file.close, the fp.close method is restored to its original method on the first line

TrapDrap:
> After calling file.close, the fp.close method is restored to its original method on the first line

But question is why make the user have to manually close the file, it should be handled on file closing
You also wouldn't be able to do that if it was sent something like `await ch.send(file=File(bytes, name))`

DenverCoder1:
Here's what discord.py says about it;

> if the File object doesn't own the io object (it owns it e.g. when you supply a filename), then it assumes you'll be the one to handle closing it

\- Zephyrkul 

> yup
> (also, closing a BytesIO doesn't really do anything tangible)

\- Danny

TrapDrap:
> > if the File object doesn't own the io object (it owns it e.g. when you supply a filename), then it assumes you'll be the one to handle closing it

This also applies to converting an asset to file btw. (Using `.to_file()`)

> > Here's what discord.py says about it

What is the point of keeping it the same? We don't have to be the same as dpy 

What if we gave `File` a parameter for automatic closing, defaulting to True?
In most cases, you won't need to reuse bytes that have been sent.

DenverCoder1:
I'm not saying it has to be the same as discord.py, I'm just adding it here for additional context so it can be discussed.

I do think there is a case for keeping it as is, for example if you are using BytesIO for a running buffer and want to keep appending to it even after a snapshot has been sent as a file.

I like your idea of making it togglable, though, especially for the case you mentioned with to_file where the user may be unaware the BytesIO is not closed automatically.

Maybe to avoid breaking changes, it could default to the current behavior, but override the _owner check if the new attribute (force_close?) is enabled?

TrapDrap:
> Maybe to avoid breaking changes, it could default to the current behavior, but override the _owner check if the new attribute (force_close?) is enabled?

That sounds like a good idea, I will work on it tmr

nextcord/file.py
```diff
@@ -144,6 +167,7 @@ def reset(self, *, seek: Union[int, bool] = True) -> None:
             self.fp.seek(self._original_pos)
 
     def close(self) -> None:
-        self.fp.close = self._closer
-        if self._owner:
+        if self._owner or self.force_close:
             self._closer()
+        else:
+            self.fp.close = self._closer
```
DenverCoder1, line 173: `self.fp.close = self._closer` should probably stay where it is. Keeping the order the same and just adding `or self.force_close` would be the best to avoid unnecessary changes.

nextcord/file.py
```diff
@@ -95,6 +117,7 @@ def __init__(
         *,
         description: Optional[str] = None,
         spoiler: bool = False,
+        force_close: bool = True,  # Accessed in .close()
```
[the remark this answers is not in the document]
TrapDrap: I believe it should default to `True` since there are rarely any cases where you would need persistent bytes after using the file.
But I'm open to getting more opinions from others too.

nextcord/file.py
TrapDrap: done

nextcord/file.py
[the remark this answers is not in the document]
ooliver1: As the discussion you guys have had above says, keeping it to the original behaviour avoids breaking changes (user closing it manually may throw an error)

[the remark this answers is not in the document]
TrapDrap: > As the discussion you guys have had above says, keeping it to the original behaviour avoids breaking changes (user closing it manually may throw an error)

The concern isn't that. The user should be able to call `.close()` multiple times without a problem. That includes if the file is automatically closed. The concern lies only if the user plans to reuse the file bytes for whatever reason.
But with due respect, since you both think it's a better idea to keep `force_close` False for user-created files.

DenverCoder1 (approved):

nextcord/file.py
```diff
@@ -63,6 +63,12 @@ class File:
         in the Discord client.
     spoiler: :class:`bool`
         Whether the attachment is a spoiler.
+    force_close: :class:`bool`
+        Whether to forcibly close the bytes used to create the file
+        when ``.close()`` is called.
+        This will also make the file bytes unusable by flusing it from
```
DenverCoder1, line 69:
```suggestion
        This will also make the file bytes unusable by flushing it from
```

nextcord/file.py
```diff
@@ -63,6 +63,12 @@ class File:
...
+        memory after it is sent once.
+        Keep this enabled if you don't wish to reuse the same bytes.
```
DenverCoder1, line 71:
```suggestion
        Enable this if you don't wish to reuse the same bytes.
```

nextcord/file.py
```diff
@@ -78,15 +84,31 @@ class File:
         in the Discord client.
     spoiler: :class:`bool`
         Whether the attachment is a spoiler.
+    force_close: :class:`bool`
+        Whether to forcibly close the bytes used to create the file
+        when ``.close()`` is called.
+        This will also make the file bytes unusable by flusing it from
+        memory after it is sent or used once.
+        Keep this enabled if you don't wish to reuse the same bytes.
```
DenverCoder1, line 92:
```suggestion
        Whether to forcibly close the bytes used to create the file
        when ``.close()`` is called.
        This will also make the file bytes unusable by flushing it from
        memory after it is sent or used once.
        Enable this if you don't wish to reuse the same bytes.
```

DenverCoder1 (changes requested):

DenverCoder1:
If you want to set it to True in the `to_file` functions, I don't think there would be any harm as far as I can tell

TrapDrap:
> If you want to set it to True in the `to_file` functions, I don't think there would be any harm as far as I can tell

Yes I am working on that currently

nextcord/asset.py
```diff
@@ -141,6 +142,12 @@ async def to_file(
             The description for the file.
         spoiler: :class:`bool`
             Whether the file is a spoiler.
+        force_close: :class:`bool`
+            Whether to forcibly close the bytes used to create the file
+            when ``.close()`` is called.
+            This will also make the file bytes unusable by flusing it from
```
DenverCoder1, line 148:
```suggestion
            This will also make the file bytes unusable by flushing it from
```

nextcord/file.py
DenverCoder1: Was marked as resolved, but it doesn't seem like these were fixed yet

TrapDrap: Weird I committed the latest

DenverCoder1 (approved):

Skelmis (approved):

nextcord/asset.py
```diff
@@ -141,6 +142,12 @@ async def to_file(
             The description for the file.
         spoiler: :class:`bool`
             Whether the file is a spoiler.
+        force_close: :class:`bool`
+            Whether to forcibly close the bytes used to create the file
+            when ``.close()`` is called.
+            This will also make the file bytes unusable by flushing it from
+            memory after it is sent or used once.
+            Keep this enabled if you don't wish to reuse the same bytes.
```
ooliver1, line 150:
```suggestion
            Keep this enabled if you don't wish to reuse the same bytes.

            .. versionadded:: 2.2
```

nextcord/file.py
```diff
@@ -63,6 +63,12 @@ class File:
         in the Discord client.
     spoiler: :class:`bool`
         Whether the attachment is a spoiler.
+    force_close: :class:`bool`
+        Whether to forcibly close the bytes used to create the file
+        when ``.close()`` is called.
+        This will also make the file bytes unusable by flushing it from
+        memory after it is sent once.
+        Enable this if you don't wish to reuse the same bytes.
```
ooliver1, line 71:
```suggestion
        Enable this if you don't wish to reuse the same bytes.

        .. versionadded:: 2.2
```

nextcord/file.py
```diff
@@ -78,15 +84,31 @@ class File:
         in the Discord client.
     spoiler: :class:`bool`
         Whether the attachment is a spoiler.
+    force_close: :class:`bool`
+        Whether to forcibly close the bytes used to create the file
+        when ``.close()`` is called.
+        This will also make the file bytes unusable by flushing it from
+        memory after it is sent or used once.
+        Enable this if you don't wish to reuse the same bytes.
```
ooliver1, line 92:
```suggestion
        Enable this if you don't wish to reuse the same bytes.

        .. versionadded:: 2.2
```

ooliver1 (changes requested):

ooliver1 (approved):

teaishealthy (approved):

PR #753, opened 2022-07-25
refactor(embeds): deprecate Embed.Empty in favour of None

Soheab:
## Summary

This PR removes `EmptyEmbed` and everything related to it.
`EmptyEmbed` and `Embed.Empty` are kept but are assigned to `None` to make this a non-breaking for users, comparing and using it still works.
One thing that does make it a breaking change is `len()` on an empty field which is not possible on None.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/embeds.py
+++ b/nextcord/embeds.py
@@ -25,12 +25,13 @@ DEALINGS IN THE SOFTWARE.
 from __future__ import annotations
 
 import datetime
+
 from typing import (
     TYPE_CHECKING,
     Any,
     Dict,
-    Final,
     List,
+    Literal,
     Mapping,
     Optional,
     Protocol,
@@ -44,19 +45,8 @@ from .colour import Colour
 
 __all__ = ("Embed",)
 
-
-class _EmptyEmbed:
-    def __bool__(self) -> bool:
-        return False
-
-    def __repr__(self) -> str:
-        return "Embed.Empty"
-
-    def __len__(self) -> int:
-        return 0
-
-
-EmptyEmbed: Final = _EmptyEmbed()
+# Backwards compatibility
+EmptyEmbed: Literal[None] = None
 
 
 class EmbedProxy:
@@ -70,8 +60,8 @@ class EmbedProxy:
         inner = ", ".join((f"{k}={v!r}" for k, v in self.__dict__.items() if not k.startswith("_")))
         return f"EmbedProxy({inner})"
 
-    def __getattr__(self, attr: str) -> _EmptyEmbed:
-        return EmptyEmbed
+    def __getattr__(self, _: str) -> None:
+        return None
 
 
 E = TypeVar("E", bound="Embed")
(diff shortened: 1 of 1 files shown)


DenverCoder1:
`len(EmptyEmbed)` now throws an error which is a breaking change for users, but likely not for many.

nextcord/embeds.py
```diff
@@ -316,37 +303,37 @@ def __bool__(self) -> bool:
...
+    def colour(self) -> Optional[Colour]:
+        return getattr(self, "_colour", None)
 
     @colour.setter
-    def colour(self, value: Union[int, Colour, _EmptyEmbed]):  # type: ignore
-        if isinstance(value, (Colour, _EmptyEmbed)):
-            self._colour = value
+    def colour(self, value: Optional[Union[int, Colour]]):  # type: ignore
+        if value is None:
+            self._colour = None
```
DenverCoder1, line 312: This seems to no longer set the attribute if it is an instance of `Colour`?

teaishealthy (changes requested):

nextcord/embeds.py
```diff
@@ -183,33 +170,34 @@ class Embed:
         "description",
     )
 
-    Empty: Final = EmptyEmbed
+    # backwards compatibility
+    Empty = EmptyEmbed
```
ooliver1, line 174: Could this perhaps be some form of property with a `FutureWarning` (as default filters ignore `DeprecationWarning`), or `__getattr__` so users know it should not be used?

nextcord/embeds.py
```diff
@@ -236,18 +224,18 @@ def from_dict(cls: Type[E], data: Mapping[str, Any]) -> E:
 
         # fill in the basic fields
 
-        self.title = data.get("title", EmptyEmbed)
-        self.type = data.get("type", EmptyEmbed)
-        self.description = data.get("description", EmptyEmbed)
-        self.url = data.get("url", EmptyEmbed)
+        self.title = data.get("title", None)
```
ooliver1, line 227: `dict.get` defaults to `None` already

nextcord/embeds.py
```diff
@@ -568,11 +556,11 @@ def remove_author(self: E) -> E:
 
     @property
     def fields(self) -> List[_EmbedFieldProxy]:
-        """List[Union[``EmbedProxy``, :attr:`Empty`]]: Returns a :class:`list` of ``EmbedProxy`` denoting the field contents.
+        """List[Union[``EmbedProxy``, ``None``]]: Returns a :class:`list` of ``EmbedProxy`` denoting the field contents.
```
ooliver1, line 559: This should use `Optional[T]` instead of `Union[T, None]`

ooliver1 (changes requested):

nextcord/embeds.py
Soheab: Done!

nextcord/embeds.py
```diff
@@ -236,18 +220,18 @@ def from_dict(cls: Type[E], data: Mapping[str, Any]) -> E:
...
-        self.type = data.get("type", EmptyEmbed)
-        self.description = data.get("description", EmptyEmbed)
-        self.url = data.get("url", EmptyEmbed)
+        self.title = data.get("title")
+        self.type = data.get("type")
+        self.description = data.get("description")
+        self.url = data.get("url")
 
-        if self.title is not EmptyEmbed:
+        if self.title is not None:
```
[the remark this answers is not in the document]
ooliver1: It is more explicit and means that the condition is `True` even if somehow an object is falsy but defines a `__str__`

ooliver1 (approved):

DenverCoder1 (approved):

PR #755, opened 2022-07-26
fix: bound interaction.client to either client/autoshardedclient instance/subclass instead of explictly lock return type to client

MaskDuck:
## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -28,7 +28,7 @@ from __future__ import annotations
 
 import asyncio
 from datetime import datetime, timedelta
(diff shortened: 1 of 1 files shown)


ooliver1 (changes requested):
I do not see any point in this way of doing it. Please use the summary for longer descriptions instead of a huge title. 

AutoShardedClient is a subclass of Client, if you're gonna use a type var, at least use it twice like they are meant to be used (a generic).

TAG-Epic:
Please see https://github.com/nextcord/nextcord/pull/682#issuecomment-1183046000

PR #756, opened 2022-07-31
feat(app-cmds): get_application_command_from_signature improvements

Soheab:
## Summary

This PR improves the `get_application_command_from_signature method(s)` to support getting a group or subcommand and optionally searches through the locale values for the (sub)command. 

This is a breaking change for 3.0

Test code: https://paste.nextcord.dev/?id=1696858771972802

## Checklist

- [ ] This PR fixes an issue.


searchify7 (approved):

Skelmis (commented):
Is there a reason to make this a breaking change? Functionally I'm not seeing it

Soheab:
> Is there a reason to make this a breaking change? Functionally I'm not seeing it

It was discussed in a [thread](https://discord.com/channels/881118111967883295/1000478739416748173) in the [nextcord server]https://discord.com/invite/nextcord) [here](https://discord.com/channels/881118111967883295/1000478739416748173/1000816150189445302) by @alentoghostflame and me.

ooliver1 (commented):
I personally think these should be all keyword forced (`*`) as there is no semantic meaning without the parameter names.

What about deprecating the current method and creating a new method with the new way of doing so?

Soheab:
> I personally think these should be all keyword forced (`*`) as there is no semantic meaning without the parameter names.
> 
> What about deprecating the current method and creating a new method with the new way of doing so?

I have done the changes, all params are now kwargs only. Also renamed `cmd_type` to `type` and re-ordered the params to make more sense.

What do you mean with deprecating and creating a new one? this method is meant for internal use so I don't think that's needed...

ooliver1:
> What do you mean with deprecating and creating a new one? this method is meant for internal use so I don't think that's needed...

It is documented, so any breaking changes to the public API are breaking changes in terms of semver/version guarantees.

Skelmis (commented):
Can you please bump the version changed to 2.3? 

I haven't tested it but I see no issues with it

Soheab:
> Can you please bump the version changed to 2.3?
> 
> I haven't tested it but I see no issues with it

I had to make some changes to make this not a breaking changed... discussed those in the following thread in the nextcord server: [#refactor get app command from signature](https://discord.com/channels/881118111967883295/1017784454795759616)

nextcord/client.py
```diff
@@ -2023,33 +2039,171 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+            if index in (0, 1):
+                if isinstance(arg, str):
+                    kwargs["qualified_name"] = kwargs.get("qualified_name", arg)
+                elif isinstance(arg, (int, ApplicationCommandType)):
+                    kwargs["type"] = kwargs.get("type", arg)
+            elif index == 2:
+                kwargs["guild_id"] = kwargs.get("guild_id", arg)
+
+        # mimicking the python error
+        def ARGS_ERROR_MESSAGE(*required_arguments: str) -> str:
```
EmreTech, line 2166: Why is this function name in all caps?

nextcord/client.py
```diff
@@ -2023,33 +2039,171 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+                without_last_argument = required_arguments[:-1]
+                last_argument = required_arguments[-1]
+                base += (
+                    ", ".join(f"'{key}'" for key in without_last_argument)
+                    + f", and '{last_argument}'"
+                )
+
+            return base
+
+        MISSING_ARGS = []
```
EmreTech, line 2182: Why is this variable name in all caps, it's not constant

nextcord/client.py
```diff
@@ -2023,33 +2039,171 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
-        cmd_type: Union[:class:`int`, :class:`ApplicationCommandType`]
+        type: Union[:class:`int`, :class:`ApplicationCommandType`]
             Type of application command.
+        qualified_name: :class:`str`
+            Full name of the application command. Capital sensitive.
+            Subcommand commands must be separated by a space, E.g, ``parent group subcommand``.
         guild_id: Optional[:class:`int`]
             Guild ID of the signature. If set to ``None``, it will attempt to get the global signature.
+        search_locales: :class:`bool`
+            Whether to also search through the command's name locals. Defaults to ``False``.
```
EmreTech, line 2083:
```suggestion
            Whether to also search through the command's name locales. Defaults to ``False``.
```

nextcord/client.py
```diff
@@ -2023,33 +2039,171 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+
         Parameters
         ----------
-        name: :class:`str`
-            Name of the application command. Capital sensitive.
-        cmd_type: Union[:class:`int`, :class:`ApplicationCommandType`]
+        type: Union[:class:`int`, :class:`ApplicationCommandType`]
             Type of application command.
+        qualified_name: :class:`str`
+            Full name of the application command. Capital sensitive.
```
EmreTech, line 2078:
```suggestion
            Full name of the application command. Case sensitive.
```

nextcord/client.py
```diff
@@ -2023,33 +2039,171 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+            Renamed the ``cmd_type`` parameter to ``type``.
+
+        .. versionchanged:: 2.3
+            Added the ``search_locales`` keyword argument.
+
+        .. versionchanged:: 2.3
+            Group/subcommand commands can now be retrieved with this method.
+
+        .. versionchanged:: 2.3
+            Changed the signature from ``(qualified_name, type, guild_id, *, search_locales)`` to ``(*, type, qualified_name, guild_id, search_locales)``.
```
EmreTech, line 2070: This part is redundant

nextcord/client.py
```diff
@@ -1950,18 +1951,33 @@ async def process_application_commands(self, interaction: Interaction) -> None:
                     f"nextcord.Client: Interaction command not found, attempting to lazy load."
                 )
                 # _log.debug(f"nextcord.Client: %s", interaction.data)
-                response_signature = (
+                debug_response_signature = (
```
EmreTech, line 1954: What's the point of this when there's `response_signature`

nextcord/client.py
```diff
@@ -2023,33 +2039,171 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
         Parameters
         ----------
-        name: :class:`str`
-            Name of the application command. Capital sensitive.
-        cmd_type: Union[:class:`int`, :class:`ApplicationCommandType`]
+        type: Union[:class:`int`, :class:`ApplicationCommandType`]
             Type of application command.
+        qualified_name: :class:`str`
+            Full name of the application command. Capital sensitive.
+            Subcommand commands must be separated by a space, E.g, ``parent group subcommand``.
```
EmreTech, line 2079:
```suggestion
            Subcommands must be separated by a space, E.g, ``parent group subcommand``.
```

nextcord/client.py
```diff
@@ -2023,33 +2039,171 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+        search_locales: :class:`bool`
+            Whether to also search through the command's name locals. Defaults to ``False``.
 
         Returns
         -------
-        command: Optional[:class:`BaseApplicationCommand`]
-            Application Command with the given signature. If no command with that signature is
-            found, returns ``None`` instead.
+        command: Optional[:class:`BaseApplicationCommand`, :class:`SlashApplicationSubcommand`]
+            Application Command with the given signature. If found else ``None`` is returned instead.
```
EmreTech, line 2088:
```suggestion
            Application Command with the given signature. If no command with that signature is found, ``None`` is returned instead.
```
The phrasing here didn't make sense and the older version didn't need changing anyways

nextcord/client.py
```diff
@@ -2023,33 +2039,171 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
         """
         return self._connection.get_application_command(command_id)
 
-    def get_application_command_from_signature(
-        self, name: str, cmd_type: Union[int, ApplicationCommandType], guild_id: Optional[int]
-    ) -> Optional[BaseApplicationCommand]:
+    def __actual_get_application_command_from_signature(
```
EmreTech, line 2042: Why was this split from `get_application_command_from_signature`? Seems like an unnecessary change since this is only called once

nextcord/client.py
```diff
@@ -2023,33 +2039,171 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+            Renamed the ``name`` parameter to ``qualified_name``.
+
+        .. versionchanged:: 2.3
+            Renamed the ``cmd_type`` parameter to ``type``.
+
+        .. versionchanged:: 2.3
+            Added the ``search_locales`` keyword argument.
+
+        .. versionchanged:: 2.3
+            Group/subcommand commands can now be retrieved with this method.
```
EmreTech, line 2067:
```suggestion
            Subcommands/Subcommand groups can now be retrieved with this method.
```

nextcord/client.py
```diff
@@ -2023,33 +2039,171 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+        qualified_name: str,
+        guild_id: Optional[int],
+        search_locales: bool = False,
+    ) -> Optional[Union[BaseApplicationCommand, SlashApplicationSubcommand]]:
         """Gets a locally stored application command object that matches the given signature.
 
+        .. versionadded:: 2.0
+
+        .. versionchanged:: 2.3
+            it's no longer possible to pass ``None`` to the ``name`` parameter.
```
EmreTech, line 2055: Shouldn't this part be merged in with the part below? Because the `name` parameter was renamed to `qualified_name`

EmreTech (changes requested):

nextcord/client.py
Soheab: Not sure but `response_signature` is defined below as a dict as is passed to `get_application_command_from_signature` when this is only for the debug log.

The `guild_id` seems to be different 🤷🏼

nextcord/client.py
Soheab: Yes but Ooliver wanted this to be a non-breaking change and since the signature was changed, I needed to use `*args, **kwargs` and do all the parsing.

nextcord/client.py
Soheab: Not really related imo

nextcord/client.py
Soheab: Thanks.

nextcord/client.py
Soheab: How so? I think it's necessary to note the kwargs only change.

nextcord/client.py
Soheab: Oh, i was going with a dpy like style 🤷🏼

nextcord/client.py
Soheab: No reason to, will change.

Soheab:
Why are those separate reviews 😞

nextcord/client.py
EmreTech: Well the specific line I left this comment on seems unnecessary because the next line pretty much sums it up.

nextcord/client.py
```diff
@@ -2023,33 +2039,171 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+                    ),
+                    stacklevel=2,
+                    category=FutureWarning,
+                )
+                kwargs[kwargs_convert_table[kw]] = kwargs.pop(kw)
+            elif kw not in VALID_KWARGS:
+                raise TypeError(f"{FUNC} got an unexpected keyword argument '{kw}'")
+
+        if len(args) > 3:
+            raise TypeError(f"{FUNC} takes 3 positional argument but {len(args)} were given")
```
EmreTech, line 2155:
```suggestion
            raise TypeError(f"{FUNC} takes 3 positional arguments but {len(args)} were given")
```

nextcord/client.py
```diff
@@ -2023,33 +2039,171 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+        for index, arg in enumerate(args):
+            if index in (0, 1):
+                if isinstance(arg, str):
+                    kwargs["qualified_name"] = kwargs.get("qualified_name", arg)
+                elif isinstance(arg, (int, ApplicationCommandType)):
+                    kwargs["type"] = kwargs.get("type", arg)
```
EmreTech, line 2161:
```suggestion
            if isinstance(arg, str) and index == 0:
                kwargs["qualified_name"] = kwargs.get("qualified_name", arg)
            elif isinstance(arg, (int, ApplicationCommandType)) and index == 1:
                kwargs["type"] = kwargs.get("type", arg)
```

nextcord/client.py
```diff
@@ -2023,33 +2039,171 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+            elif index == 2:
+                kwargs["guild_id"] = kwargs.get("guild_id", arg)
```
EmreTech, line 2163: Shouldn't the type of the argument also be checked here to be in line with the other if statements?

nextcord/client.py
```diff
@@ -2023,33 +2039,171 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+            return base
+
+        MISSING_ARGS = []
+        for arg in REQUIRED_ARGS:
+            if arg not in kwargs:
+                MISSING_ARGS.append(arg)
+
+        if MISSING_ARGS:
+            message = ARGS_ERROR_MESSAGE(*MISSING_ARGS)
+            raise TypeError(message)
```
EmreTech, line 2189: Couldn't you just raise with direct result of `ARGS_ERROR_MESSAGE` instead of putting it in a variable?

EmreTech (changes requested):
Few more things I noticed

teaishealthy (approved):
Except for Emre's comments, LGTM

nextcord/client.py
Soheab: No, the others are changed because of the original order, this should always be int or None.

nextcord/client.py
```diff
@@ -2102,33 +2117,170 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
         """
         return self._connection.get_application_command(command_id)
 
-    def get_application_command_from_signature(
-        self, name: str, cmd_type: Union[int, ApplicationCommandType], guild_id: Optional[int]
-    ) -> Optional[BaseApplicationCommand]:
+    def __actual_get_application_command_from_signature(
+        self,
+        *,
+        type: Union[int, ApplicationCommandType],
```
alentoghostflame, line 2123: I'm a little iffy about `type` being used, as that shadows a built-in Python function.

alentoghostflame (commented):
I haven't tested at all and am entirely relying on you to fully test it, but other than that one bit it looks good to me.

nextcord/client.py
EmreTech: I don't see a problem with using `type` here. The builtin `type` isn't used in this function and I don't see why it would be used since all this does is call `ConnectionState.get_application_command_from_signature`.

nextcord/client.py
```diff
@@ -2102,33 +2117,170 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+    def get_application_command_from_signature(
+        self, *args: Any, search_locales: bool = False, **kwargs: Any
+    ) -> Optional[Union[BaseApplicationCommand, SlashApplicationSubcommand]]:
+        REQUIRED_ARGS = ("qualified_name", "type", "guild_id")
+        FUNC = "get_application_command_from_signature()"
+        ARGS_DEPRECATED_MESSAGE = "The signature passed to {name} is deprecated, please use keyword arguments only instead. Converted signature: {converted_signature}."
+        KWARG_DEPRECATION_MESSAGE = 'The keyword argument "{old}=" used in {name} is deprecated, please use "{new}=" instead.'
+
+        kwargs_convert_table: Dict[str, Any] = {"name": "qualified_name", "cmd_type": "type"}
+        VALID_KWARGS = ("type", "qualified_name", "guild_id", "search_locales")
```
EmreTech, line 2218: Wouldn't `search_locales` get consumed by the explicitly defined keyword only argument?

nextcord/client.py
```diff
@@ -2102,33 +2117,170 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+        qualified_name: str,
+        type: Union[int, ApplicationCommandType],
+        guild_id: Optional[int],
+        *,
+        search_locales: bool = ...,
+    ) -> Optional[Union[BaseApplicationCommand, SlashApplicationSubcommand]]:
+        ...
+
+    @overload
+    def get_application_command_from_signature(
```
EmreTech, line 2198: Why is this overload here? The original `get_application_command_from_signature` never supported this function signature, so I don't see a point in adding it if positional arguments are deprecated for this method.

nextcord/state.py
```diff
@@ -881,8 +921,21 @@ async def discover_application_commands(
...
-            response_signature = (raw_response["name"], int(payload_type), fixed_guild_id)
-            if app_cmd := self.get_application_command_from_signature(*response_signature):
+            response_signature = {
+                "type": int(payload_type),
+                "qualified_name": raw_response["name"],
+                "guild_id": fixed_guild_id,
+            }
+            app_cmd = self.get_application_command_from_signature(**response_signature)
+            if app_cmd:
+                if not isinstance(app_cmd, BaseApplicationCommand):
```
EmreTech, line 931: When would `get_application_command_from_signature` not return an instance of `BaseApplicationCommand`?
Remember that instances of a class that are a subclass of the second argument for `isinstance` still returns `True`:
```python
>>> class A: pass
... 
>>> class B(A): pass
... 
>>> b = B()
>>> isinstance(b, A)
True
>>>
```

EmreTech (changes requested):

ooliver1:
@Soheab any updates on this?

nextcord/state.py
Soheab: That was added because pyright/lance was complaining about some attr. But that was in Nextcord v2.1/2 and I haven't checked if something changed regarding to that.

nextcord/client.py
Soheab: Good catch, thanks.

nextcord/client.py
Soheab: Unsure about that since the overloads are making this a non-breaking change and the overload you mention is the new signature but minus the kwargs only. Also, the code supports it 🤷🏼

EmreTech: Well why do we need a new posargs signature if the old posargs signature is supposed to be deprecated?

nextcord/state.py
EmreTech: From my own testing, pyright is complaining that `SlashApplicationSubcommand` is incompatible with `BaseApplicationCommand` and `SlashApplicationSubcommand` not having certain attributes and all of that stuff. If the received command is supposed to be a `BaseApplicationCommand` and not a `SlashApplicationSubcommand`, then this is fine to keep in.

EmreTech: I also looked around the code, and found out that `SlashApplicationSubcommand` is not a subclass of `BaseApplicationCommand`. I thought it was for some reason.

nextcord/client.py
```diff
@@ -2029,18 +2029,33 @@ async def process_application_commands(self, interaction: Interaction) -> None:
...
                 )
                 # _log.debug(f"nextcord.Client: %s", interaction.data)
-                response_signature = (
+                debug_response_signature = (
                     interaction.data["name"],
                     int(interaction.data["type"]),
                     interaction.guild_id,
                 )
-                _log.debug("nextcord.Client: %s", response_signature)
+                _log.debug(f"nextcord.Client: %s", debug_response_signature)
```
ooliver1, line 2037:
```suggestion
                _log.debug("nextcord.Client: %s", debug_response_signature)
```

nextcord/client.py
ooliver1: It isn't a breaking change to remove type annotations, as people using type checkers should be able to know their signature is deprecated.

ooliver1 (changes requested):

nextcord/client.py
Soheab: Hmm yeah that's true

nextcord/client.py
```diff
@@ -2096,33 +2111,170 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+        *,
+        type: Union[int, ApplicationCommandType],
+        qualified_name: str,
+        guild_id: Optional[int],
+        search_locales: bool = ...,
+    ) -> Optional[Union[BaseApplicationCommand, SlashApplicationSubcommand]]:
+        ...
+
+    @overload
+    def get_application_command_from_signature(
```
EmreTech, line 2181: Shouldn't there also be overloads where the parameters have their old names? The old signature isn't positional only, after all.

EmreTech (commented):
With 3.0 development on the horizon (most likely), I think it would be better if the signature-related changes were in a separate PR.

Soheab:
Updated for 3.0! Needs testing.

nextcord/client.py
```diff
@@ -2181,29 +2195,48 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
...
+    ) -> Optional[Union[BaseApplicationCommand, SlashApplicationSubcommand]]:
         """Gets a locally stored application command object that matches the given signature.
 
+        .. versionadded:: 2.0
+
+        .. versionchanged:: 3.0
+            - Subcommands/Subcommand groups can now be retrieved with this method.
+            - ``name`` parameter was refactored to ``qualified_name`` and now accepts subcommands/subcommand groups separated by a space.
+            - ``cmd_type`` parameter was refactored to ``type``, defaults to :attr:`.ApplicationCommandType.chat_input` and is now a keyword-only parameter.
+            - ``guild_id`` parameter was refactored to ``guild`` with type :class:`~nextcord.abc.Snowflake`, defaults to ``None`` and is now a keyword-only parameter.
```
EmreTech, line 2213:
```suggestion
            - ``name`` parameter was renamed to ``qualified_name`` and now accepts subcommands/subcommand groups separated by a space.
            - ``cmd_type`` parameter was renamed to ``type``, defaults to :attr:`.ApplicationCommandType.chat_input` and is now a keyword-only parameter.
            - ``guild_id`` parameter was renamed to ``guild`` with type :class:`~nextcord.abc.Snowflake`, defaults to ``None`` and is now a keyword-only parameter.
```
I don't think "refactored" is the right term for parameter renames.

nextcord/client.py
```diff
@@ -2181,29 +2195,48 @@ def get_application_command(self, command_id: int) -> Optional[BaseApplicationCo
         return self._connection.get_application_command(command_id)
 
     def get_application_command_from_signature(
-        self, name: str, cmd_type: Union[int, ApplicationCommandType], guild_id: Optional[int]
-    ) -> Optional[BaseApplicationCommand]:
+        self,
+        qualified_name: str,
+        *,
+        type: Union[int, ApplicationCommandType] = ApplicationCommandType.chat_input,
+        guild: Optional[Snowflake] = None,
```
EmreTech, line 2202: Why was this changed to accept guild objects? I don't see a good reason for it, considering the underlying method in the state still accepts the guild id anyways.

nextcord/client.py
Soheab: I was thinking since more than one thing changed about it but ig.

nextcord/client.py
EmreTech: I think changing it to be a union of `int` and `Snowflake` would be a good idea.

Skelmis (approved):
Tested using provided test code, seems fine to me

EmreTech (approved):
Tested

ooliver1 (approved):

PR #784, opened 2022-08-20
Localization examples

i80287:
## Summary

Added file with two examples for the localization: localization for the name and description of the slash command and for the SlashOption in French, German and Spanish

## Checklist

- [x ] If code changes were made then they have been tested.
- [ ] This PR fixes an issue.
- [x ] This PR is **not** a code change (e.g. documentation, README, ...)

### DIFF
--- a/examples/application_commands/localization.py
+++ b/examples/application_commands/localization.py
@@ -1,0 +1,101 @@
+
+from nextcord import Interaction, Locale, SlashOption
+from nextcord.ext import commands
+
+
+TESTING_GUILD_ID = 123456789  # Replace with your testing guild id
+BOT_TOKEN = ""                # Replace with your bot token
+
+
+bot = commands.Bot()
+
+
+@bot.slash_command(
+    name="hello",              # name of the command by default in English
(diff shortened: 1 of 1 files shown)


examples/application_commands/localization.py
```diff
@@ -0,0 +1,99 @@
...
+BOT_TOKEN = ""  # Replace with your bot token
+
+
+bot = commands.Bot()
+
+
+@bot.slash_command(
+    name="hello",  # name of the command by default in English
+    name_localizations={
+        Locale.de: "hallo",  # localization for the name of the command in German (Deutsch)
```
ooliver1, line 14: Do we really need an inline comment for almost every line?

ooliver1 (changes requested):

examples/application_commands/localization.py
i80287: Is this question for other library developers or for me?

i80287: In which lines should I remove the comments?

PR #785, opened 2022-08-20
fix(app-cmds): stop syncing on no options/channel type sorting

alentoghostflame:
## Summary

Does the following:
* Stops the slash command payload from always setting "options", even when they have no options or children.
  * This fixes possible needless resyncing of option-less commands.
  * This fixes lazy loading of option-less commands.
* Sorts the command option payload "channel_types" list due to Discord sending them back sorted.
  * This fixes needless resyncing of commands with an option that has channel_types set out of order.
  * This fixes lazy loading of ^

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -2597,8 +2597,9 @@ class SlashApplicationCommand(SlashCommandMixin, BaseApplicationCommand, Autocom
         ret = super().get_payload(guild_id)
         if self.children:
             ret["options"] = [child.payload for child in self.children.values()]
-        else:
+        elif self.options:
             ret["options"] = [parameter.payload for parameter in self.options.values()]
+
         return ret
 
     async def call(self, state: ConnectionState, interaction: Interaction) -> None:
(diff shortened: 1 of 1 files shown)


Skelmis (approved):
I have not tested this, however, the reasoning sounds clear to me

nextcord/application_command.py
```diff
@@ -367,6 +367,10 @@ def payload(self) -> dict:
         if self.channel_types:
             # noinspection PyUnresolvedReferences
             ret["channel_types"] = [channel_type.value for channel_type in self.channel_types]
+            ret[
+                "channel_types"
+            ].sort()  # When reading application commands from Discord, they return the channel types
+            #  sorted. To prevent needless syncing and allow lazy loading, we need to sort them as well.
```
DenverCoder1, line 373: Better to put comments above for readability, can use `sorted` as a more "functional" and concise approach
```suggestion
            # When reading application commands from Discord, they return the channel types sorted.
            # To prevent needless syncing and allow lazy loading, we need to sort them as well.
            ret["channel_types"] = sorted([channel_type.value for channel_type in self.channel_types])
```

ooliver1 (approved):
Apart from denver's comment.

DenverCoder1 (approved):

PR #787, opened 2022-08-21
docs(examples): add a localization example

i80287:
## Summary

This is a pull request with two examples for localization: for slash command and for SlashOption in German, French and Spanish
## Checklist

- [ ] This PR fixes an issue.


examples/application_commands/localization.py
```diff
@@ -0,0 +1,97 @@
...
+async def hello_command(interaction: Interaction):
+
+    if interaction.locale == "de":
+        await interaction.response.send_message("Hallo Welt!")
+    elif interaction.locale == "fr":
+        await interaction.response.send_message("Bonjour le monde!")
+    elif interaction.locale == "es-ES":
+        await interaction.response.send_message("¡Hola Mundo!")
+    else:
+        await interaction.response.send_message("Hello, world!")
```
DenverCoder1, line 35: These can use the Locale enum also for consistency?

```suggestion
    if interaction.locale == Locale.de:
        await interaction.response.send_message("Hallo Welt!")
    elif interaction.locale == Locale.fr:
        await interaction.response.send_message("Bonjour le monde!")
    elif interaction.locale == Locale.es_ES:
        await interaction.response.send_message("¡Hola Mundo!")
    else:
        await interaction.response.send_message("Hello, world!")
```

examples/application_commands/localization.py
```diff
@@ -0,0 +1,97 @@
...
+        await interaction.response.send_message(f"Sie haben **`{colour}`** Farbe gewählt")
+    elif interaction.locale == "fr":
+        await interaction.response.send_message(f"Vous avez choisi la couleur rouge **`{colour}`**")
+    elif interaction.locale == "es-ES":
+        await interaction.response.send_message(f"Elegiste el color **`{colour}`**")
+    else:
+        await interaction.response.send_message(f"You chose **`{colour}`** colour")
+
+
+bot.run(BOT_TOKEN)
```
DenverCoder1, line 97: In other examples, we just use token in quotes for the example token
```suggestion
bot.run("token")
```

examples/application_commands/localization.py
```diff
@@ -0,0 +1,97 @@
+from nextcord import Interaction, Locale, SlashOption
+from nextcord.ext import commands
+
+TESTING_GUILD_ID = 123456789  # Replace with your testing guild id
+BOT_TOKEN = ""  # Replace with your bot token
```
DenverCoder1, line 5: See previous comment
```suggestion
```

examples/application_commands/localization.py
DenverCoder1: This applies to both commands

DenverCoder1: It seems like I was mistaken about this. It appears the comparison with string and Locale does not currently work the way I expected it would. Maybe in the future this will be valid, but for now the previous way you had it was correct. Sorry about that.

DenverCoder1 (approved):

examples/application_commands/localization.py
```diff
@@ -0,0 +1,96 @@
...
+    description_localizations={
+        Locale.de: "Beschreibung des Befehls auf Deutsch",
+        Locale.fr: "Description de la commande en français",
+        Locale.es_ES: "Descripción del comando en español",
+    },
+    guild_ids=[TESTING_GUILD_ID],
+)
+async def hello_command(interaction: Interaction):
+
+    if interaction.locale == "de":
```
ooliver1, line 27: Slight nitpick but could this unnecessary blank line be removed?

```suggestion
async def hello_command(interaction: Interaction):
    if interaction.locale == "de":
```

ooliver1 (changes requested):

Skelmis (approved):

ooliver1 (approved):

PR #790, opened 2022-08-22
feat: Add Client.parse_mentions, Guild.parse_mentions, Guild.parse_role_mentions, Guild.parse_channel_mentions

DenverCoder1:
## Summary

Adds methods to parse mentions from strings. This is useful especially when using slash commands with string options where you want the users, roles, or channels mentioned in the slash option (for example when using a string option to emulate variadic arguments). See example code below.

Adds:

* `utils.parse_raw_mentions` (returns User IDs)
* `utils.parse_raw_role_mentions` (returns Role IDs)
* `utils.parse_raw_channel_mentions` (returns Channel IDs)
* `Client.parse_mentions` (returns Users)
* `Guild.parse_mentions` (returns Members and/or Users)
* `Guild.parse_role_mentions` (returns Roles)
* `Guild.parse_channel_mentions` (returns GuildChannels)

Example code:

```py
@bot.slash_command()
async def mentions(interaction: Interaction, mentions: str):
    if interaction.guild:
        users = interaction.guild.parse_mentions(mentions)
        roles = interaction.guild.parse_role_mentions(mentions)
        channels = interaction.guild.parse_channel_mentions(mentions)
        await interaction.send(f"Users: {users}\nRoles: {roles}\nChannels: {channels}")
    else:
        users = bot.parse_mentions(mentions)
        await interaction.send(f"Users: {users}")
```

## Checklist

- [ ] This PR fixes an issue.


nextcord/client.py
```diff
@@ -2621,3 +2621,28 @@ def decorator(func: Callable):
...
+        text: :class:`str`
+            String to parse mentions in.
+
+        Returns
+        -------
+        List[:class:`User`]
+            List of :class:`User` objects that were mentioned in the string.
+        """
+
+        it = filter(None, map(self.get_user, utils.parse_raw_mentions(text)))
```
EmreTech, line 2647: What if in the text someone mentions a user object that isn't in the cache?

nextcord/utils.py
```diff
@@ -851,6 +854,68 @@ def escape_mentions(text: str) -> str:
...
+    -----------
+    text: :class:`str`
+        The text to parse mentions from.
+
+    Returns
+    --------
+    List[:class:`int`]
+        A list of user IDs that were mentioned.
+    """
+    return [int(x) for x in re.findall(r"<@!?(\d{15,20})>", text)]
```
EmreTech, line 878: In an attempt to future-proof this, I think there shouldn't be a maximum character count

nextcord/utils.py
```diff
@@ -851,6 +854,68 @@ def escape_mentions(text: str) -> str:
     return re.sub(r"@(everyone|here|[!&]?[0-9]{17,20})", "@\u200b\\1", text)
 
 
+def parse_raw_mentions(text: str) -> List[int]:
+    """A helper function that parses mentions from a string as an array of user IDs
+    matched with the syntax of ``<@user_id>``.
```
EmreTech, line 859:
```suggestion
    matched with the syntax of ``<@user_id>`` or ``<@!user_id>``.
```

nextcord/client.py
DenverCoder1: In that event, the developer should use parse_raw_mentions and fetch users as needed.

I don't think the helper function for getting the mentions from text should be making API calls.

This is also the same implementation used already by Message.channel_mentions, etc.

nextcord/utils.py
DenverCoder1: The nickname mention syntax is deprecated, but it is still supported like normal mentions, so I agree

nextcord/client.py
EmreTech: Ah that makes sense

nextcord/utils.py
DenverCoder1: This is simply relocated from its original location in Message.raw_mentions without logic changes.

21-digit IDs won't be coming until the year 2770 (probably way beyond Discord's existence) so I think this is fine

https://snowsta.mp/?l=en-us&z=1t&f=atx10uuark-lfls

EmreTech: Ah yes

ooliver1 (approved):

Skelmis (approved):

PR #792, opened 2022-08-23
fix(app-cmds): delete_application_commands for guild commands

benedictvs:
## Summary

Fixed the issue regarding delete_application_commands where commands could not be deleted from one guild specifically.
This was caused by a hardcoded value of None for the guild_id being passed to delete_application_command. Additionally, 
logging messages and error messages were added for when a non-global command is attempted to be globally unregistered,
which simply caused a KeyError exception before.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/state.py
+++ b/nextcord/state.py
@@ -1086,17 +1086,26 @@ class ConnectionState:
         if self.application_id is None:
             raise NotImplementedError("Could not get the current application id")
 
-        if guild_id:
-            await self.http.delete_guild_command(
-                self.application_id, guild_id, command.command_ids[guild_id]
-            )
-        else:
-            await self.http.delete_global_command(
-                self.application_id, command.command_ids[guild_id]
-            )
+        try:
+            if guild_id:
+                await self.http.delete_guild_command(
+                    self.application_id, guild_id, command.command_ids[guild_id]
+                )
+            else:
+                await self.http.delete_global_command(
+                    self.application_id, command.command_ids[guild_id]
+                )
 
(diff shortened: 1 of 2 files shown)


EmreTech (approved):
LGTM, haven't tested

DenverCoder1 (approved):

nextcord/state.py
```diff
@@ -1086,17 +1086,31 @@ async def delete_application_command(
...
-                self.application_id, guild_id, command.command_ids[guild_id]
+        try:
+            if guild_id:
+                await self.http.delete_guild_command(
+                    self.application_id, guild_id, command.command_ids[guild_id]
+                )
+            else:
+                await self.http.delete_global_command(
+                    self.application_id, command.command_ids[guild_id]
+                )
```
DenverCoder1, line 1097: What specific other exceptions is this block catching?

Also, maybe the KeyError and other possible errors should be added to the docstrings (Raises section) if they can occur.

benedictvs: The last except block is catching any other possible exceptions, it is mirroring the catch-all except block that is present in register_application_commands.

And that is a good idea, although the KeyError is the only specific exception I have come across.

DenverCoder1: Looks good, it seems like it logs HTTP exceptions and makes the error more specific

teaishealthy (approved):
lgtm except for denvers comment

ooliver1 (approved):
Same review summary as above.

PR #801, opened 2022-08-31
feat: Make enums comparable and hashable as a str or int

DenverCoder1:
## Summary

* Refactors Enums using the built-in implementation for simplicity and maintainability, which may be a few nanoseconds slower*, but has many advantages over the current implementation including support for multiple levels of inheritance and data type mixins.

* The integer value Enums now subclass `IntEnum` which allows treating attributes as integers when comparing and when hashing. The string value Enums now subclass `StrEnum` which allows treating attributes as strings when comparing and when hashing. Calling `int` on any `IntEnum` subclass or `str` on any `StrEnum` subclass will return the value.

* Everything from the old implementation including name and value attributes, `try_enum`, as well ordering comparisons are still supported.

* When `try_enum` fails to find the specified attribute, an instance of the named tuple `UnknownEnumValue` is returned rather than one of the dynamically generated private classes for the enum (eg. `_EnumValue_Locale`) which were otherwise inaccessible.

### Why?

When comparing with string attributes such as `interaction.locale`, instead of using a hardcoded string, using `.value`, or using `str()`, you can now compare directly with the `Locale`:
```py
# OLD:
if interaction.locale == "en-US":
    await interaction.response.send_message(f"You chose **`{colour}`** color")
# or
if interaction.locale == Locale.en_US.value:
    await interaction.response.send_message(f"You chose **`{colour}`** color")
# or
if interaction.locale == str(Locale.en_US):
    await interaction.response.send_message(f"You chose **`{colour}`** color")

# NEW
if interaction.locale == Locale.en_US:
    await interaction.response.send_message(f"You chose **`{colour}`** color")
```

Strings and Locales (and other string Enums) can now be used interchangeably in dicts:
```py
# OLD:
locales = {
    Locale.en_US: "a",
}
locales["en-US"] = "b"
print(locales[Locale.en_US])  # OLD: prints "a", NEW: prints "b"
```

\* the performance difference is negligible.

## Checklist

- [ ] This PR fixes an issue.


teaishealthy (changes requested):
Instead of defining all comparison methods on `UnknownEnumValue` just define `__lt__` and `__eq__` and add the `functools.total_ordering` decorator on the class. It's less error prone and easier to maintain.

EmreTech:
> Instead of defining all comparison methods on `UnknownEnumValue` just define `__lt__` and `__eq__` and add the `functools.total_ordering` decorator on the class. It's less error prone and easier maintain.

`easier maintain` -> `easier to maintain`

DenverCoder1:
> Instead of defining all comparison methods on `UnknownEnumValue` just define `__lt__` and `__eq__` and add the `functools.total_ordering` decorator on the class. It's less error prone and easier to maintain.

Can't use it if I want it to compare with other types such as `int` or `str`.

I considered it, but it would break in some cases, for example:

```py
nextcord.enums.try_enum(ChannelType, 20) <= 20

  File "test_801.py", line 132, in <module>
    nextcord.enums.try_enum(ChannelType, 20) <= 20,
TypeError: '<=' not supported between instances of 'UnknownEnumValue' and 'int'
```

teaishealthy (approved):

Skelmis (approved):
Untested but looks good to me and would be a nice step forward for maintainability

ooliver1 (approved):
Definitely cleans up `enums.py` a huge amount, thank you!

PR #802, opened 2022-09-01
refactor(http): raise warning when downgrading API version

DenverCoder1:
## Summary

* Added an `UnsupportedAPIVersion` warning raised when setting the API version lower than the currently supported version*.
* Removed `_modify_api_version` message content intent workaround from docs.

\*"Supported version" in this context means a version that is expected to work with all methods in nextcord and users may ask for support with. Old API versions even if they are marked "Available" by Discord, may be incompatible with some methods, so are not supported.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/http.py
+++ b/nextcord/http.py
@@ -128,9 +128,10 @@ def _modify_api_version(version: Literal[9, 10]):
     """Modify the API version used by the HTTP client.
 
     Additional versions may be added around the time of a Discord API
-    version bump to allow temporarily downgrading to an older API version.
+    version bump to allow temporarily downgrading to an older API version
+    or upgrading to a newer version that is not yet supported by the library.
 
-    Downgrading to an older API version is not supported and may result in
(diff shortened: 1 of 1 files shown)


TAG-Epic:
Can this not just be removed? Imo a deprecation here is not needed

DenverCoder1:
> Can this not just be removed? Imo a deprecation here is not needed

I'm thinking it may be useful in the future around new API version changes and also for development purposes for testing newer versions.

It also would be less breaking since people may have not removed it from their code at the time they upgrade their nextcord version. This is not the biggest concern, though, as while it was suggested in the server and docs, is not technically part of the public API.

Skelmis:
I agree with Denver on warning rather then removing.

Also, let's be real here. No doubt discord will do something similar in future API versions so already having this capability wouldn't be a big issue to just have it left in.

ArjunSharda (approved):

Skelmis (approved):

PR #803, opened 2022-09-04
feat: add the ability to use custom interactions

### LINKED ISSUE #757: Custom Interactions
### Summary

a `get_context()` alternative for slash command Interaction objects

Application commands

### The Problem

Currently, custom functions cannot be applied to `Interaction` objects like they can to the `Context` object used in normal commands

### The Ideal Solution

A `get_application()` method gets added to the `Client` object.

### The Current Solution

_No response_

### Additional Context

_No response_

Skelmis:
## Summary

Adds the ability to provide a custom interaction class which the library will use.

Closes #757 

Note the return type of `get_interaction` varies from it's `get_context` counterpart as this was the only way I could find that would appease `task pyright`. If you are aware of an alternative where the return type can simply be `InterT` please let me know.

## Checklist

- [x] This PR fixes an issue.

## Test code

```python
import os

from nextcord import Interaction, Client

TESTING_GUILD_ID = 500525882226769931


class CustomInter(Interaction):
    @property
    def custom_prop(self) -> bool:
        return True


class CustomClient(Client):
    def get_interaction(self, data, *, cls=Interaction):
        return super().get_interaction(data, cls=CustomInter)


client = CustomClient()


@client.event
async def on_ready():
    print("...")


async def test(interaction: CustomInter):
    await interaction.send(f"{isinstance(interaction, CustomInter)=}, {interaction.custom_prop}")


client.run(os.environ["TOKEN"])
```

### DIFF
--- a/examples/custom_interaction.py
+++ b/examples/custom_interaction.py
@@ -20,9 +20,8 @@ class Client(nextcord.Client):
         return super().get_interaction(data, cls=MyInteraction)
 
 
+client = Client()
 TESTING_GUILD_ID = 123456789  # Replace with your testing guild id
-intents = nextcord.Intents.default()
-client = Client(intents=intents)
 
 
 @client.slash_command(guild_ids=[TESTING_GUILD_ID])
--- a/examples/custom_interaction.py
+++ b/examples/custom_interaction.py
@@ -1,0 +1,33 @@
+import nextcord
+
+
+class MyInteraction(nextcord.Interaction):
+    async def send_embed(self, *, title: str, description: str):
+        """Easily send a basic embed."""
+        embed: nextcord.Embed = nextcord.Embed(
+            title=title,
+            description=description,
+        )
+        embed.set_author(name=self.user.display_name, icon_url=self.user.display_avatar)
+        await self.send(embed=embed)
+
+
+class Client(nextcord.Client):
+    def get_interaction(self, data, *, cls=nextcord.Interaction):
+        # when you override this method, you pass your new Interaction
+        # subclass to the super() method, which tells the bot to
+        # use the new MyInteraction class
+        return super().get_interaction(data, cls=MyInteraction)
+
+
+TESTING_GUILD_ID = 123456789  # Replace with your testing guild id
+intents = nextcord.Intents.default()
+client = Client(intents=intents)
+
+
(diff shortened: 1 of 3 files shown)


nextcord/client.py
```diff
@@ -2652,3 +2655,40 @@ def parse_mentions(self, text: str) -> List[User]:
...
+            class be provided, it should be a subclass of :py:class:`Interaction`.
+
+        Returns
+        -------
+        Interaction
+            An instance :py:class:`Interaction` or the provided subclass.
+
+        Notes
+        -----
+        This is synchronous due to how slash commands are implemented.
```
ooliver1, line 2692: So it shows in a blue box like other notes
```suggestion
        .. note::

            This is synchronous due to how slash commands are implemented.
```

ooliver1 (changes requested):
It might make sense to add an example on how to use this. In the docs or here in `./examples`

DenverCoder1:
> If you are aware of an alternative where the return type can simply be `InterT` please let me know.

You can do it the dpy way, by adding `# type: ignore` on the return line.

Skelmis:
> > If you are aware of an alternative where the return type can simply be `InterT` please let me know.
> 
> You can do it the dpy way, by adding `# type: ignore` on the return line.

Ideally without just ignoring things.

Skelmis:
I couldn't think of a better example for a custom interaction, but it should serve it's purpose none the less

ooliver1 (approved):

teaishealthy (approved):

examples/custom_interaction.py
```diff
@@ -0,0 +1,32 @@
...
+
+class MyInteraction(nextcord.Interaction):
+    async def send_embed(self, *, title: str, description: str):
+        """Easily send a basic embed."""
+        embed: nextcord.Embed = nextcord.Embed(
+            title=title,
+            description=description,
+        )
+        embed.set_author(name=self.user.display_name, icon_url=self.user.display_avatar)
+        await self.send(embed=embed)
```
DenverCoder1, line 12: I'd probably make this return the partialinteractionmessage/webhookmessage if I were doing this, but it's just an example, so it's fine either way

DenverCoder1 (approved):
Looks good, tested

ArjunSharda (approved):

PR #813, opened 2022-09-09
feat: use delete_message_seconds for Guild/Member.ban

### LINKED ISSUE #800: Use `delete_message_seconds` for `Guild/Member.ban`
### Summary

discord/discord-api-docs#5219

The core library

### The Problem

The new way to set a duration is with seconds not days, this will likely be removed in API V11.

*`reason` being removed means nothing to us as since API V10, all routes use the `X-Audit-Log-Reason` header*

### The Ideal Solution

`seconds` refers to `delete_message_seconds`, same with `days`
- Expose `seconds` to the user
- deprecate `days` but translate the given value to `seconds` (so no breaking changes)
- raise a `ValueError` if both provided (like `embed` and `embeds` are both mutually exclusive)

### The Current Solution

_No response_

### Additional Context

_No response_

teaishealthy:
## Summary

Resolves #800

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/guild.py
+++ b/nextcord/guild.py
@@ -2887,6 +2887,8 @@ class Guild(Hashable):
         You must have the :attr:`~Permissions.ban_members` permission to
         do this.
 
+        For backwards compatibility reasons, by default one day worth of messages will be deleted.
+
         .. warning::
             delete_message_days is deprecated and will be removed in a future version.
             Use delete_message_seconds instead.
@@ -2895,14 +2897,14 @@ class Guild(Hashable):
         -----------
         user: :class:`abc.Snowflake`
             The user to ban from their guild.
-        reason: Optional[:class:`str`]
-            The reason the user got banned.
         delete_message_seconds: Optional[:class:`int`]
             The number of seconds worth of messages to delete from the user
             in the guild.
         delete_message_days: Optional[:class:`int`]
             The number of days worth of messages to delete from the user
             in the guild. The minimum is 0 and the maximum is 7.
+        reason: Optional[:class:`str`]
+            The reason the user got banned.
 
         Raises
         -------
--- a/nextcord/guild.py
+++ b/nextcord/guild.py
@@ -43,7 +43,6 @@ from typing import (
     Set,
     Tuple,
     Union,
-    cast,
     overload,
 )
 
@@ -2899,7 +2898,7 @@ class Guild(Hashable):
             The user to ban from their guild.
         delete_message_seconds: Optional[:class:`int`]
             The number of seconds worth of messages to delete from the user
-            in the guild.
+            in the guild. The minimum is 0 and the maximum is 604800 (7 days).
         delete_message_days: Optional[:class:`int`]
             The number of days worth of messages to delete from the user
             in the guild. The minimum is 0 and the maximum is 7.
@@ -2917,21 +2916,18 @@ class Guild(Hashable):
             raise InvalidArgument(
                 "Cannot pass both delete_message_days and delete_message_seconds."
             )
-        elif delete_message_days is None and delete_message_seconds is None:
-            # Default to one day
-            delete_message_seconds = 24 * 60 * 60
-
-        if delete_message_days is not None:
+        elif delete_message_days is not None:
             warnings.warn(
                 DeprecationWarning(
                     "delete_message_days is deprecated, use delete_message_seconds instead."
                 )
             )
             delete_message_seconds = delete_message_days * 24 * 60 * 60
+        elif delete_message_seconds is None:
+            # Default to one day
+            delete_message_seconds = 24 * 60 * 60
 
-        await self._state.http.ban(
-            user.id, self.id, cast(int, delete_message_seconds), reason=reason
-        )
+        await self._state.http.ban(user.id, self.id, delete_message_seconds, reason=reason)
 
     async def unban(self, user: Snowflake, *, reason: Optional[str] = None) -> None:
         """|coro|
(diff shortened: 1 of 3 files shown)


nextcord/guild.py
```diff
@@ -2885,15 +2887,22 @@ async def ban(
...
 
+        .. warning::
+            delete_message_days is deprecated and will be removed in a future version.
+            Use delete_message_seconds instead.
+
         Parameters
         -----------
         user: :class:`abc.Snowflake`
             The user to ban from their guild.
-        delete_message_days: :class:`int`
```
EmreTech, line 2892: Why was this moved down below? Seems unnecessary TBH

nextcord/guild.py
```diff
@@ -2874,7 +2875,8 @@ async def ban(
         user: Snowflake,
         *,
         reason: Optional[str] = None,
-        delete_message_days: Literal[0, 1, 2, 3, 4, 5, 6, 7] = 1,
+        delete_message_seconds: Optional[int] = None,
```
EmreTech, line 2878: This type hint should be set to `int` and its default should be set to `0` to match the HTTP route implementation

nextcord/guild.py
```diff
@@ -2902,7 +2911,25 @@ async def ban(
...
+        if delete_message_days is not None:
+            warnings.warn(
+                DeprecationWarning(
+                    "delete_message_days is deprecated, use delete_message_seconds instead."
+                )
+            )
+            delete_message_seconds = delete_message_days * 24 * 60 * 60
+
+        await self._state.http.ban(
+            user.id, self.id, cast(int, delete_message_seconds), reason=reason
```
EmreTech, line 2931: Since casting an object doesn't actually do anything to the object itself at runtime, this is pretty unsafe if `delete_message_seconds` isn't passed into this method
First of all, it violates the type hints of the `ban` method and second of all I don't think Discord can handle `null` for that attribute

nextcord/member.py
```diff
@@ -648,14 +648,20 @@ def communication_disabled_until(self) -> Optional[datetime.datetime]:
     async def ban(
         self,
         *,
-        delete_message_days: Literal[0, 1, 2, 3, 4, 5, 6, 7] = 1,
+        delete_message_seconds: Optional[int] = None,
```
EmreTech, line 651: The type hint and defaults should also be changed here

EmreTech (changes requested):

nextcord/guild.py
EmreTech: Resolved via a Discord conversation starting at this message: https://canary.discord.com/channels/881118111967883295/881124672056467456/1018006566097666098

nextcord/guild.py
EmreTech: https://github.com/nextcord/nextcord/pull/813#discussion_r967581183

nextcord/member.py
EmreTech: https://github.com/nextcord/nextcord/pull/813#discussion_r967581183

nextcord/guild.py
```diff
@@ -2902,7 +2911,25 @@ async def ban(
         HTTPException
             Banning failed.
         """
-        await self._state.http.ban(user.id, self.id, delete_message_days, reason=reason)
+        if delete_message_days is not None and delete_message_seconds is not None:
+            raise InvalidArgument(
+                "Cannot pass both delete_message_days and delete_message_seconds."
+            )
+        elif delete_message_days is None and delete_message_seconds is None:
+            # Default to one day
```
Skelmis, line 2919: Can we document this default in the docstring?

EmreTech (approved):
LGTM, haven't tested

Skelmis (approved):
Untested, lgtm

nextcord/http.py
```diff
@@ -879,12 +879,12 @@ def ban(
         self,
         user_id: Snowflake,
         guild_id: Snowflake,
-        delete_message_days: int = 1,
+        delete_message_seconds: int = 0,
```
DenverCoder1, line 882: Maybe this should be the number of seconds in a day in case it is called directly. If it isn't ever called without it, there wouldn't be a need for a default at all.
```suggestion
        delete_message_seconds: int = 86400,
```

nextcord/guild.py
```diff
@@ -2902,7 +2913,25 @@ async def ban(
...
+            warnings.warn(
+                DeprecationWarning(
+                    "delete_message_days is deprecated, use delete_message_seconds instead."
+                )
+            )
+            delete_message_seconds = delete_message_days * 24 * 60 * 60
+
+        await self._state.http.ban(
+            user.id, self.id, cast(int, delete_message_seconds), reason=reason
+        )
```
DenverCoder1, line 2934: You don't need to use `cast` if you have a condition that checks if `delete_message_seconds` is `None`.

```suggestion
        elif delete_message_days is not None:
            warnings.warn(
                DeprecationWarning(
                    "delete_message_days is deprecated, use delete_message_seconds instead."
                )
            )
            delete_message_seconds = delete_message_days * 24 * 60 * 60
        elif delete_message_seconds is None:
            # Default to one day
            delete_message_seconds = 24 * 60 * 60

        await self._state.http.ban(user.id, self.id, delete_message_seconds, reason=reason)
```

nextcord/guild.py
```diff
@@ -43,6 +43,7 @@
     Set,
     Tuple,
     Union,
+    cast,
```
DenverCoder1, line 46: See above (cast is not needed)

```suggestion
```

DenverCoder1 (approved):

ooliver1 (approved):

PR #814, opened 2022-09-11
fix(app-cmds): Slash Commands cannot have options with the name "state"

EmreTech:
## Summary

This PR fixes an issue found in the Nextcord Discord server where a Slash Command cannot have an option with the name "state" without raising an exception.

Traceback (from original help thread): 
```
Ignoring exception in on_interaction
  File "/root/env/lib/python3.8/site-packages/nextcord/client.py", line 499, in _run_event
    await coro(*args, **kwargs)
  File "/root/env/lib/python3.8/site-packages/nextcord/client.py", line 1924, in on_interaction
    await self.process_application_commands(interaction)
  File "/root/env/lib/python3.8/site-packages/nextcord/client.py", line 1944, in process_application_commands
    await app_cmd.call_from_interaction(interaction)
  File "/root/env/lib/python3.8/site-packages/nextcord/application_command.py", line 2255, in call_from_interaction
    await self.call(self._state, interaction)  # type: ignore
  File "/root/env/lib/python3.8/site-packages/nextcord/application_command.py", line 2619, in call
    await self.call_slash(state, interaction, option_data)
  File "/root/env/lib/python3.8/site-packages/nextcord/application_command.py", line 1716, in call_slash
    await self.invoke_callback_with_hooks(state, interaction, **kwargs)
TypeError: invoke_callback_with_hooks() got multiple values for argument 'state'
```

## Checklist

- [x] This PR fixes an issue.
- [x] This PR is a breaking change (e.g. methods or parameters removed/renamed) (technically more of an internal breaking change)

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -835,7 +835,7 @@ class CallbackMixin:
         return True
 
     async def invoke_callback_with_hooks(
-        self, state: ConnectionState, interaction: Interaction, *args, **kwargs
+        self, state: ConnectionState, interaction: Interaction, args: List[Any] = [], kwargs: Dict[str, Any] = {}
     ) -> None:
         """|coro|
         Invokes the callback with all hooks and checks.
@@ -1730,7 +1730,7 @@ class SlashCommandMixin(CallbackMixin):
             )
         else:
             kwargs = await self.get_slash_kwargs(state, interaction, option_data)
-            await self.invoke_callback_with_hooks(state, interaction, **kwargs)
+            await self.invoke_callback_with_hooks(state, interaction, kwargs=kwargs)
 
     def get_mention(self, guild: Optional[Snowflake] = None) -> str:
         """Returns a string that allows you to mention the slash command.
@@ -2809,7 +2809,7 @@ class UserApplicationCommand(BaseApplicationCommand):
 
     async def call(self, state: ConnectionState, interaction: Interaction):
         await self.invoke_callback_with_hooks(
-            state, interaction, get_users_from_interaction(state, interaction)[0]
+            state, interaction, args=[get_users_from_interaction(state, interaction)[0]]
         )
 
     def from_callback(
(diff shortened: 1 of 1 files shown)


alentoghostflame (approved):
Haven't tested it myself, but looks good to me.

nextcord/application_command.py
```diff
@@ -835,7 +835,11 @@ async def can_run(self, interaction: Interaction) -> bool:
         return True
 
     async def invoke_callback_with_hooks(
-        self, state: ConnectionState, interaction: Interaction, *args, **kwargs
+        self,
+        state: ConnectionState,
+        interaction: Interaction,
+        args: List[Any] = [],
+        kwargs: Dict[str, Any] = {},
```
Skelmis, line 842: Is there a way we can avoid using mutable defaults here?

N.b. I'm aware they are not mutated within this method but it'd still be a good idea for the future if changes were to occur in this method which mutated them

Skelmis (changes requested):

nextcord/application_command.py
EmreTech: Wdym "mutable defaults"
Like setting the defaults to an empty list and an empty dict?

Skelmis: Correct

Reference [link](https://florimond.dev/en/posts/2018/08/python-mutable-defaults-are-the-source-of-all-evil/) explaining it a bit better then I can

EmreTech: Ohhhhhhh I see now

EmreTech: 1de830e

nextcord/application_command.py
```diff
@@ -835,11 +835,21 @@ async def can_run(self, interaction: Interaction) -> bool:
...
+        interaction: Interaction,
+        args: Optional[List[Any]] = None,
+        kwargs: Optional[Dict[str, Any]] = None,
     ) -> None:
         """|coro|
         Invokes the callback with all hooks and checks.
         """
+
+        if not args:
+            args = []
```
Skelmis, line 849:
```suggestion
            args = ()
```

Skelmis (approved):
Untested but looks good to me. Only other comment is possibly making `args` and `kwargs` keyword only

ooliver1 (approved):

PR #819, opened 2022-09-18
Update reaction.users docs

Miolus:
## Summary

Update docs to reflect changes in API:
I recently ran into errors due to a change in the Discord API: reaction.users will not return a User object for left members anymore but for all members that reacted.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/reaction.py
+++ b/nextcord/reaction.py
@@ -205,10 +205,8 @@ class Reaction:
         Yields
         --------
         Union[:class:`User`, :class:`Member`]
-            The member (if retrievable) or the user that has reacted
-            to this message. The case where it can be a :class:`Member` is
-            in a guild message context. Sometimes it can be a :class:`User`
-            if the member has left the guild.
+            Bot reactions will return a :class:`Member`, while user reactions will return a :class:`User`.
+            All reactions will be a :class:`User`, if the member has left.
         """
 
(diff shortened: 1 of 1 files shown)


Skelmis:
Can you link those docs for the sake of us who don't follow all the changes?

Miolus:
grafik
https://discord.com/developers/docs/resources/channel#get-reactions

They state they only return user objects which somehow is not reality. Should it be changed to only reflect the docs?

Skelmis:
Okay so I've spent some time looking into how this works and there is a couple factors at play.

Essentially the issue as I understand based on your comments is that your only ever getting `User` objects, and you are correct that this is all the API returns. 
Internally the way nextcord handles this is that if the reactions are within a guild we will first attempt to get a `Member` object from the internal cache and if that fails then nextcord will instead give you the `User` object as returned by the API.

My best guess as to the cause of your issue is that your bot does not contain `Member` objects for the users who have reacted, which is why your receiving only `User` objects from this.

I'm going to close this as invalid as this does appear to be documented correctly, however, I'd be happy to discuss both this and your own issues further in the nextcord discord. (Ping @Skelmis)

Hope this helps!

Note: The implementation for this can be found [here](https://github.com/nextcord/nextcord/blob/master/nextcord/iterators.py#L213-L239)

PR #820, opened 2022-09-19
feat(client): add Client.remove_view and Client.remove_modal

### LINKED ISSUE #425: Implement remove_view in Client

EmreTech:
## Summary

This implements `remove_view` and `remove_modal` in `nextcord.Client` and `nextcord.ConnectionState`. Closes #425.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/client.py
+++ b/nextcord/client.py
@@ -1872,7 +1872,7 @@ class Client:
 
         self._connection.store_view(view, message_id)
 
-    def remove_view(self, view: View) -> None:
+    def remove_view(self, view: View, message_id: Optional[int] = None) -> None:
         """Removes a :class:`~nextcord.ui.View` from persistent listening.
 
         This method should be used if a persistent view is set in a cog and
@@ -1901,7 +1901,7 @@ class Client:
                 "View is not persistent. Items need to have a custom_id set and View must have no timeout"
             )
 
-        self._connection.remove_view(view)
+        self._connection.remove_view(view, message_id)
 
     def add_modal(self, modal: Modal, *, user_id: Optional[int] = None) -> None:
         """Registers a :class:`~nextcord.ui.Modal` for persistent listening.
--- a/nextcord/client.py
+++ b/nextcord/client.py
@@ -1862,7 +1862,6 @@ class Client:
             The view is not persistent. A persistent view has no timeout
             and all their components have an explicitly provided custom_id.
         """
-
         if not isinstance(view, View):
             raise TypeError(f"Expected an instance of View not {view.__class__!r}")
 
@@ -1872,6 +1871,37 @@ class Client:
             )
 
         self._connection.store_view(view, message_id)
+
+    def remove_view(self, view: View) -> None:
+        """Removes a :class:`~nextcord.ui.View` from persistent listening.
+
+        This method should be used if a persistent view is set in a cog and
+        should be freed when the cog is unloaded to save memory.
+
+        .. versionadded:: 2.3
+
+        Parameters
+        ------------
+        view: :class:`nextcord.ui.View`
+            The view to remove from dispatching.
+        
+        Raises
+        -------
+        TypeError
+            A view was not passed.
+        ValueError
+            The view is not persistent. A persistent view has no timeout
+            and all their components have an explicitly provided custom_id.
+        """
+        if not isinstance(view, View):
+            raise TypeError(f"Expected an instance of View not {view.__class__!r}")
+
+        if not view.is_persistent():
+            raise ValueError(
+                "View is not persistent. Items need to have a custom_id set and View must have no timeout"
+            )
+
+        self._connection.remove_view(view)
 
     def add_modal(self, modal: Modal, *, user_id: Optional[int] = None) -> None:
         """Registers a :class:`~nextcord.ui.Modal` for persistent listening.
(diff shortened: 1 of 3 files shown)


nextcord/client.py
```diff
@@ -1873,6 +1872,37 @@ def add_view(self, view: View, *, message_id: Optional[int] = None) -> None:
...
+    def remove_view(self, view: View) -> None:
+        """Removes a :class:`~nextcord.ui.View` from persistent listening.
+
+        This method should be used if a persistent view is set in a cog and
+        should be freed when the cog is unloaded to save memory.
+
+        .. versionadded:: 2.3
+
+        Parameters
+        ------------
```
ooliver1, line 1884: Unnecessary heading underlining.
```suggestion
        Parameters
        ----------
```

nextcord/client.py
```diff
@@ -1873,6 +1872,37 @@ def add_view(self, view: View, *, message_id: Optional[int] = None) -> None:
...
+        view: :class:`nextcord.ui.View`
+            The view to remove from dispatching.
+
+        Raises
+        -------
```
ooliver1, line 1889:
```suggestion
        Raises
        ------
```

nextcord/client.py
```diff
@@ -1907,6 +1937,38 @@ def add_modal(self, modal: Modal, *, user_id: Optional[int] = None) -> None:
...
+    def remove_modal(self, modal: Modal) -> None:
+        """Removes a :class:`~nextcord.ui.Modal` from persistent listening.
+
+        This method should be used if a persistent modal is set in a cog and
+        should be freed when the cog is unloaded to save memory.
+
+        .. versionadded:: 2.3
+
+        Parameters
+        ------------
```
ooliver1, line 1949:
```suggestion
        Parameters
        ----------
```

nextcord/client.py
```diff
@@ -1907,6 +1937,38 @@ def add_modal(self, modal: Modal, *, user_id: Optional[int] = None) -> None:
...
+        modal: :class:`nextcord.ui.Modal`
+            The view to remove from dispatching.
+
+        Raises
+        -------
```
ooliver1, line 1954:
```suggestion
        Raises
        ------
```

ooliver1 (changes requested):

Skelmis (changes requested):
Sadly the view store implements `add_view` and `remove_view` in different ways

``add_view`` stores views under the key ``(item.type.value, message_id, item.custom_id)`` while ``remove_view`` attempts to pop the key ``(item.type.value, item.custom_id)``, which don't line up. This is not your fault, however, requires fixing for this PR to actually work based on my tests.

I do not have a suitable fix for this at time of review. This should likely become a thread in bikeshedding as it requires further investigation + debugging with live data I don't have on hand right now.
Note: I will open the thread myself so I can add debug data

Skelmis (approved):
[Tested](https://nexttest.koldfusion.xyz/nextcord/820)

nextcord/client.py
```diff
@@ -1873,6 +1872,37 @@ def add_view(self, view: View, *, message_id: Optional[int] = None) -> None:
...
+
+        This method should be used if a persistent view is set in a cog and
+        should be freed when the cog is unloaded to save memory.
+
+        .. versionadded:: 2.3
+
+        Parameters
+        ----------
+        view: :class:`nextcord.ui.View`
+            The view to remove from dispatching.
```
DenverCoder1, line 1886: `message_id` parameter is not yet documented

nextcord/client.py
```diff
@@ -1907,6 +1940,38 @@ def add_modal(self, modal: Modal, *, user_id: Optional[int] = None) -> None:
...
+
+        This method should be used if a persistent modal is set in a cog and
+        should be freed when the cog is unloaded to save memory.
+
+        .. versionadded:: 2.3
+
+        Parameters
+        ----------
+        modal: :class:`nextcord.ui.Modal`
+            The view to remove from dispatching.
```
DenverCoder1, line 1954:
```suggestion
            The modal to remove from dispatching.
```

DenverCoder1 (commented):
What happens when a message_id is not passed? If the view is sent in multiple messages, do you need to remove it for each one? What happens if the view wasn't sent in a message?

EmreTech:
> What happens when a message_id is not passed? If the view is sent in multiple messages, do you need to remove it for each one? What happens if the view wasn't sent in a message?

I don't get what the confusion is about? It should pretty much work the opposite of `add_view`, where by default `message_id` is None.
Can you elaborate why you asked these questions in the first place? I just don't get it.

DenverCoder1:
I didn't realize `add_view` took a message_id, but it makes a lot of sense now. I had just always used it without one like in the examples, so I wasn't sure where the `message_id` was coming from.

DenverCoder1 (approved):

nextcord/client.py
```diff
@@ -1873,6 +1872,40 @@ def add_view(self, view: View, *, message_id: Optional[int] = None) -> None:
...
+        Raises
+        ------
+        TypeError
+            A view was not passed.
+        ValueError
+            The view is not persistent. A persistent view has no timeout
+            and all their components have an explicitly provided custom_id.
+        """
+        if not isinstance(view, View):
+            raise TypeError(f"Expected an instance of View not {view.__class__!r}")
```
teaishealthy, line 1900:
```suggestion
            raise TypeError(f"Expected an instance of View not {view.__class__.__name__}")
```
This is inconsistent with the rest of the library

teaishealthy (changes requested):

ooliver1 (approved):

PR #833, opened 2022-10-03
feat(app-cmds): default_guild_ids

### LINKED ISSUE #395: Global scope of application command guild ids
### Summary

Something to make itso that you dont have to put guild_ids in every slash command

The core library

### The Problem

So on each slash command, if you don't want to wait 1 hour for the command to finally update on your testing server, you have to put guild_ids on every command
but it gets annoying something and I have a bad memory
I could forget

### The Ideal Solution

Something that you could put in nextcord.Client() or a function in wich u put ur guild id

### The Current Solution

guild_ids :/

### Additional Context

The ideal solution doesnt have to be what I proposed, but still, it would be those 2 solutions that I think are ideal

EmreTech:
## Summary

This PR adds a new parameter to `Client` and `AutoShardedClient` named `default_guild_ids` that acts as the default guild ids for application commands that don't have any guild ids set. Closes #395.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/interactions.rst
+++ b/docs/interactions.rst
@@ -266,3 +266,24 @@ Shown below is an example of a simple command running in a cog:
             await interaction.response.send_message("Hello I am a slash command in a cog!")
 
 Context menu commands can also be made in cogs using the :meth:`nextcord.user_command` or :meth:`nextcord.message_command` decorators.
+
+
+Default Guild Ids
+-----------------
+
+If you would like for all of your application commands to have the same guild ids unless explicitly stated, then you can set the ``default_guild_ids`` keyword argument in :class:`nextcord.ext.commands.Bot` and :class:`nextcord.ext.commands.AutoShardedBot`!
+
+Here's an example of this:
+
+.. code-block:: python3
+
+    bot = commands.Bot(..., default_guild_ids=[TESTING_GUILD_ID])
+
+    @bot.slash_command()
+    async def bye(interaction: nextcord.Interaction):
+        await interaction.response.send_message(f"Goodbye {bot.default_guild_ids}!")
+
+    @bot.slash_command(guild_ids=None)
+    async def bye_global(interaction: nextcord.Interaction):
+        await interaction.response.send_message("Goodbye everyone!")
+
--- a/docs/interactions.rst
+++ b/docs/interactions.rst
@@ -268,7 +268,7 @@ Shown below is an example of a simple command running in a cog:
 Context menu commands can also be made in cogs using the :meth:`nextcord.user_command` or :meth:`nextcord.message_command` decorators.
 
 
-Default Guild Ids
+Default Guild IDs
 -----------------
 
 If you would like for all of your application commands to have the same guild ids unless explicitly stated, then you can set the ``default_guild_ids`` keyword argument in :class:`nextcord.ext.commands.Bot` and :class:`nextcord.ext.commands.AutoShardedBot`!
--- a/docs/interactions.rst
+++ b/docs/interactions.rst
@@ -286,4 +286,3 @@ Here's an example of this:
     @bot.slash_command(guild_ids=None)
     async def bye_global(interaction: nextcord.Interaction):
         await interaction.response.send_message("Goodbye everyone!")
-
--- a/docs/interactions.rst
+++ b/docs/interactions.rst
@@ -271,7 +271,7 @@ Context menu commands can also be made in cogs using the :meth:`nextcord.user_co
 Default Guild IDs
 -----------------
 
-If you would like for all of your application commands to have the same guild ids unless explicitly stated, then you can set the ``default_guild_ids`` keyword argument in :class:`nextcord.ext.commands.Bot` and :class:`nextcord.ext.commands.AutoShardedBot`!
+If you would like for all of your application commands to have the same guild ids unless explicitly stated, then you can set the ``default_guild_ids`` keyword argument in :class:`~nextcord.Client`, :class:`~nextcord.ext.commands.Bot`, or :class:`~nextcord.ext.commands.AutoShardedBot`!
 
 Here's an example of this:
 
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -1815,7 +1815,7 @@ class BaseApplicationCommand(CallbackMixin, CallbackWrapperMixin):
             Callback to make the application command from, and to run when the application command is called.
         guild_ids: Iterable[:class:`int`]
             An iterable list/set/whatever of guild ID's that the application command should register to.
-            If not passed and :attr:`Client.default_guild_ids` is set, then those default guild ids will 
+            If not passed and :attr:`Client.default_guild_ids` is set, then those default guild ids will
             be used instead. If both of those are unset, then the command will be a global command. Defaults to ``MISSING``.
         dm_permission: :class:`bool`
             If the command should be usable in DMs or not. Setting to ``False`` will disable the command from being
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -2912,7 +2912,7 @@ def slash_command(
     *,
     name_localizations: Optional[Dict[Union[Locale, str], str]] = None,
     description_localizations: Optional[Dict[Union[Locale, str], str]] = None,
-    guild_ids: Optional[Iterable[int]] = None,
+    guild_ids: Optional[Iterable[int]] = MISSING,
     dm_permission: Optional[bool] = None,
     default_member_permissions: Optional[Union[Permissions, int]] = None,
     force_global: bool = False,
@@ -2933,8 +2933,10 @@ def slash_command(
     description_localizations: Dict[Union[:class:`Locale`, :class:`str`], :class:`str`]
         Description(s) of the subcommand for users of specific locales. The locale code should be the key, with the
         localized description as the value.
-    guild_ids: Iterable[:class:`int`]
-        IDs of :class:`Guild`'s to add this command to. If unset, this will be a global command.
+    guild_ids: Optional[Iterable[:class:`int`]]
+        IDs of :class:`Guild`'s to add this command to. If set to :attr:`utils.MISSING` and :attr:`Client.default_guild_ids` is
+        set, then those default guild ids will be used instead. If both of those are unset, then the command will
+        be a global command. Defaults to `MISSING`.
     dm_permission: :class:`bool`
         If the command should be usable in DMs or not. Setting to ``False`` will disable the command from being
         usable in DMs. Only for global commands, but will not error on guild.
@@ -2971,7 +2973,7 @@ def message_command(
     name: Optional[str] = None,
     *,
     name_localizations: Optional[Dict[Union[Locale, str], str]] = None,
-    guild_ids: Optional[Iterable[int]] = None,
+    guild_ids: Optional[Iterable[int]] = MISSING,
     dm_permission: Optional[bool] = None,
     default_member_permissions: Optional[Union[Permissions, int]] = None,
     force_global: bool = False,
@@ -2986,8 +2988,10 @@ def message_command(
     name_localizations: Dict[Union[:class:`Locale`, :class:`str`], :class:`str`]
         Name(s) of the command for users of specific locales. The locale code should be the key, with the localized
         name as the value
-    guild_ids: Iterable[:class:`int`]
-        IDs of :class:`Guild`'s to add this command to. If unset, this will be a global command.
+    guild_ids: Optional[Iterable[:class:`int`]]
+        IDs of :class:`Guild`'s to add this command to. If set to :attr:`utils.MISSING` and :attr:`Client.default_guild_ids` is
+        set, then those default guild ids will be used instead. If both of those are unset, then the command will
+        be a global command. Defaults to `MISSING`.
     dm_permission: :class:`bool`
         If the command should be usable in DMs or not. Setting to ``False`` will disable the command from being
         usable in DMs. Only for global commands, but will not error on guild.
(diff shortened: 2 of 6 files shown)


nextcord/client.py
```diff
@@ -301,6 +305,7 @@ def __init__(
         rollout_register_new: bool = True,
         rollout_update_known: bool = True,
         rollout_all_guilds: bool = False,
+        default_guild_ids: List[int] = [],
```
Skelmis, line 308: Can this be defined as a non-mutable default value please

nextcord/shard.py
```diff
@@ -357,6 +357,7 @@ def __init__(
         rollout_register_new: bool = True,
         rollout_update_known: bool = True,
         rollout_all_guilds: bool = False,
+        default_guild_ids: List[int] = [],
```
Skelmis, line 360: Same here with regard to mutable defaults

Skelmis (changes requested):

DenverCoder1:
Copying my comment from https://github.com/nextcord/nextcord/pull/567#issuecomment-1094166889

Something might need to be changed in the slash commands code to make it so that setting a command's guild_ids to [] will make it global (and not use the client default).

It may need to differentiate between guild_ids being not being set and guild_ids being set to None/empty list.

Maybe we could have a flag to check if guild_ids was MISSING or not in the constructor?

Expected behavior:

Command guild_ids not set (or set to MISSING) - client default guild ids should be used if any exist

Command guild_ids set to None or empty iterable - command should be global, client default should not be used

EmreTech:
> You could also maybe look at the implementation of https://github.com/nextcord/nextcord/pull/567, as I think it was pretty clean, didn't require modifying the state.

It wasn't the cleanest imo. The code that handled adding the default guild ids to the application commands was duplicated and added to two methods in `Client`. Sure, maybe modifying the state to accomplish this isn't the best idea, but it's better than duplicated code.

What I can do instead is to add my code that actually adds the default guild ids to an application command in `Client.add_application_command` instead of `ConnectionState.add_application_command`.

EmreTech:
> Maybe we could have a flag to check if guild_ids was MISSING or not in the constructor?

How would we check if `guild_ids` is not in the constructor? There are no keyword variable arguments we could exploit.

DenverCoder1:
> > Maybe we could have a flag to check if guild_ids was MISSING or not in the constructor?
> 
> How would we check if `guild_ids` is not in the constructor? There are no keyword variable arguments we could exploit.

BaseApplicationCommand, SlashApplicationCommand, the decorators, etc have `guild_ids` defaulting to `None`. This makes it create an empty set for guild ids to roll out.

If instead MISSING were the default, you'd know if the parameter was specified as None/iterable or not specified at all.

Does that make sense?

EmreTech:
> > > Maybe we could have a flag to check if guild_ids was MISSING or not in the constructor?
> > 
> > 
> > How would we check if `guild_ids` is not in the constructor? There are no keyword variable arguments we could exploit.
> 
> BaseApplicationCommand, SlashApplicationCommand, the decorators, etc have `guild_ids` defaulting to `None`. This makes it create an empty set for guild ids to roll out.
> 
> If instead MISSING were the default, you'd know if the parameter was specified as None/iterable or not specified at all.
> 
> Does that make sense?

That makes sense. Thank you!

DenverCoder1:
It may require an extra field or something, since just storing the empty doesn't keep the info of whether guild_ids was passed or not

nextcord/client.py
```diff
@@ -351,6 +356,10 @@ def __init__(
         self._rollout_all_guilds: bool = rollout_all_guilds
         self._application_commands_to_add: Set[BaseApplicationCommand] = set()
 
+        if default_guild_ids is None:
+            default_guild_ids = []
+        self._default_guild_ids = default_guild_ids
```
DenverCoder1, line 361: I believe this can be made more concise:

```suggestion
        self._default_guild_ids = default_guild_ids or []
```

Or, to remove duplicates:

```suggestion
        self._default_guild_ids = set(default_guild_ids) if default_guild_ids is not None else set()
```

nextcord/shard.py
```diff
@@ -357,6 +357,7 @@ def __init__(
         rollout_register_new: bool = True,
         rollout_update_known: bool = True,
         rollout_all_guilds: bool = False,
+        default_guild_ids: Optional[List[int]] = None,
```
DenverCoder1, line 360: Does this also need to be added to ext.commands.Bot? (I haven't tested)

EmreTech: Yup

EmreTech: 030af98

nextcord/application_command.py
```diff
@@ -2932,8 +2933,10 @@ def slash_command(
     description_localizations: Dict[Union[:class:`Locale`, :class:`str`], :class:`str`]
         Description(s) of the subcommand for users of specific locales. The locale code should be the key, with the
         localized description as the value.
-    guild_ids: Iterable[:class:`int`]
-        IDs of :class:`Guild`'s to add this command to. If unset, this will be a global command.
+    guild_ids: Optional[Iterable[:class:`int`]]
+        IDs of :class:`Guild`'s to add this command to. If set to :attr:`utils.MISSING` and :attr:`Client.default_guild_ids` is
+        set, then those default guild ids will be used instead. If both of those are unset, then the command will
+        be a global command. Defaults to `MISSING`.
```
ooliver1, line 2939:
```suggestion
        be a global command. Defaults to ``MISSING``.
```

nextcord/application_command.py
```diff
@@ -2985,8 +2988,10 @@ def message_command(
     name_localizations: Dict[Union[:class:`Locale`, :class:`str`], :class:`str`]
         Name(s) of the command for users of specific locales. The locale code should be the key, with the localized
         name as the value
-    guild_ids: Iterable[:class:`int`]
-        IDs of :class:`Guild`'s to add this command to. If unset, this will be a global command.
+    guild_ids: Optional[Iterable[:class:`int`]]
+        IDs of :class:`Guild`'s to add this command to. If set to :attr:`utils.MISSING` and :attr:`Client.default_guild_ids` is
+        set, then those default guild ids will be used instead. If both of those are unset, then the command will
+        be a global command. Defaults to `MISSING`.
```
ooliver1, line 2994:
```suggestion
        be a global command. Defaults to ``MISSING``.
```

nextcord/application_command.py
```diff
@@ -2985,8 +2988,10 @@ def message_command(
     name_localizations: Dict[Union[:class:`Locale`, :class:`str`], :class:`str`]
         Name(s) of the command for users of specific locales. The locale code should be the key, with the localized
         name as the value
-    guild_ids: Iterable[:class:`int`]
-        IDs of :class:`Guild`'s to add this command to. If unset, this will be a global command.
+    guild_ids: Optional[Iterable[:class:`int`]]
+        IDs of :class:`Guild`'s to add this command to. If set to :attr:`utils.MISSING` and :attr:`Client.default_guild_ids` is
```
ooliver1, line 2992: `MISSING` isnt technically documented right now, and its not an attribute but simply a `:data:` (non-class/function/method/...), so probably best to just say "if not passed"

nextcord/application_command.py
```diff
@@ -3036,8 +3041,10 @@ def user_command(
...
```
ooliver1, line 3045: All the same as above.

nextcord/application_command.py
```diff
@@ -2932,8 +2933,10 @@ def slash_command(
     description_localizations: Dict[Union[:class:`Locale`, :class:`str`], :class:`str`]
         Description(s) of the subcommand for users of specific locales. The locale code should be the key, with the
         localized description as the value.
-    guild_ids: Iterable[:class:`int`]
-        IDs of :class:`Guild`'s to add this command to. If unset, this will be a global command.
+    guild_ids: Optional[Iterable[:class:`int`]]
+        IDs of :class:`Guild`'s to add this command to. If set to :attr:`utils.MISSING` and :attr:`Client.default_guild_ids` is
```
ooliver1, line 2937: `Client.default_guild_ids` since its not documented (not a property or added in `Attributes`), it should likely be a properly, or
```
and ``default_guild_ids`` is set in :class:`Client`
```

nextcord/client.py
```diff
@@ -265,6 +265,10 @@ class Client:
         Warning: While enabling this will prevent "ghost" commands on guilds with removed code references, rolling out
         to ALL guilds with anything other than a very small bot will likely cause it to get rate limited.
 
+    default_guild_ids: Optional[List[:class:`int`]]
+        The default guild ids for every application command set. If the application command doesn't have any explicit
+        guild ids set and this list is not empty, then the application command's guild ids will be set to this.
+        Defaults to `None`.
```
ooliver1, line 271:
```suggestion
        Defaults to ``None``.
```

nextcord/client.py
```diff
@@ -2568,8 +2581,10 @@ def message_command(
         name_localizations: Dict[Union[:class:`Locale`, :class:`str`], :class:`str`]
             Name(s) of the command for users of specific locales. The locale code should be the key, with the localized
             name as the value
-        guild_ids: Iterable[:class:`int`]
-            IDs of :class:`Guild`'s to add this command to. If unset, this will be a global command.
+        guild_ids: Optional[Iterable[:class:`int`]]
+            IDs of :class:`Guild`'s to add this command to. If set to :attr:`utils.MISSING` and `default_guild_ids` is
+            set, then those default guild ids will be used instead. If both of those are unset, then the command will
+            be a global command. Defaults to `MISSING`.
```
ooliver1, line 2587:
```suggestion
            IDs of :class:`Guild`'s to add this command to. If set to :attr:`utils.MISSING` and ``default_guild_ids`` is
            set, then those default guild ids will be used instead. If both of those are unset, then the command will
            be a global command. Defaults to ``MISSING``.
```
and similar comment about `` :attr:`utils.MISSING` ``

nextcord/client.py
```diff
@@ -2623,8 +2638,10 @@ def slash_command(
         description_localizations: Dict[Union[:class:`Locale`, :class:`str`], :class:`str`]
             Description(s) of the command for users of specific locales. The locale code should be the key, with the
             localized description as the value.
-        guild_ids: Iterable[:class:`int`]
-            IDs of :class:`Guild`'s to add this command to. If unset, this will be a global command.
+        guild_ids: Optional[Iterable[:class:`int`]]
+            IDs of :class:`Guild`'s to add this command to. If set to :attr:`utils.MISSING` and `default_guild_ids` is
+            set, then those default guild ids will be used instead. If both of those are unset, then the command will
```
ooliver1, line 2643: and again

ooliver1 (approved):

ooliver1 (changes requested):
Didn't actually mean to approve, mb

nextcord/application_command.py
```diff
@@ -2911,7 +2912,7 @@ def slash_command(
     *,
     name_localizations: Optional[Dict[Union[Locale, str], str]] = None,
     description_localizations: Optional[Dict[Union[Locale, str], str]] = None,
-    guild_ids: Optional[Iterable[int]] = None,
+    guild_ids: Optional[Iterable[int]] = MISSING,
```
DenverCoder1, line 2915: There are other places in the code with a `guild_ids` parameter. While it may not be as easily instantiated, should these also default to `MISSING`?

Example:
image

nextcord/application_command.py
DenverCoder1: MISSING has actually never been referenced in the docs as far as I can tell. Usually we just say "unset".

The "Defaults to `MISSING`" can probably be left out entirely.

nextcord/application_command.py
DenverCoder1: I think this is what should be used everywhere MISSING is referenced in docstrings in this PR. That would be consistent with what has usually been done.

nextcord/application_command.py
```diff
@@ -1815,6 +1815,8 @@ def __init__(
             Callback to make the application command from, and to run when the application command is called.
         guild_ids: Iterable[:class:`int`]
             An iterable list/set/whatever of guild ID's that the application command should register to.
+            If not passed and :attr:`Client.default_guild_ids` is set, then those default guild ids will
+            be used instead. If both of those are unset, then the command will be a global command. Defaults to ``MISSING``.
```
DenverCoder1, line 1819: I don't think it should say "Defaults to MISSING" anywhere in the docs. To the user of lib, this just means it's not passed.


```suggestion
            be used instead. If both of those are unset, then the command will be a global command.
```

DenverCoder1 (approved):

DenverCoder1:
Do you have example code? It might be nice, although not necessary, to have a detailed example of how to use this in interactions.rst (Application Commands page).

docs/interactions.rst
```diff
@@ -266,3 +266,23 @@ Shown below is an example of a simple command running in a cog:
             await interaction.response.send_message("Hello I am a slash command in a cog!")
 
 Context menu commands can also be made in cogs using the :meth:`nextcord.user_command` or :meth:`nextcord.message_command` decorators.
+
+
+Default Guild Ids
+-----------------
+
+If you would like for all of your application commands to have the same guild ids unless explicitly stated, then you can set the ``default_guild_ids`` keyword argument in :class:`nextcord.ext.commands.Bot` and :class:`nextcord.ext.commands.AutoShardedBot`!
```
DenverCoder1, line 274: This works with Client too, right?

```suggestion
If you would like for all of your application commands to have the same guild ids unless explicitly stated, then you can set the ``default_guild_ids`` keyword argument in :class:`~nextcord.Client`, :class:`~nextcord.ext.commands.Bot`, or :class:`~nextcord.ext.commands.AutoShardedBot`!
```

docs/interactions.rst
```diff
@@ -266,3 +266,23 @@ Shown below is an example of a simple command running in a cog:
             await interaction.response.send_message("Hello I am a slash command in a cog!")
 
 Context menu commands can also be made in cogs using the :meth:`nextcord.user_command` or :meth:`nextcord.message_command` decorators.
+
+
+Default Guild Ids
```
DenverCoder1, line 271:
```suggestion
Default Guild IDs
```

docs/interactions.rst
EmreTech: It does, but it looks like the rest of the examples use `ext.commands` instead of the main library.

DenverCoder1 (approved):

ooliver1 (approved):

PR #834, opened 2022-10-03
feat: http ratelimit events

### LINKED ISSUE #154: Ratelimit events
Title says it all.

EmreTech:
## Summary

This adds a new ratelimit event that is dispatched whenever an HTTP request exhausts a bucket or gets a 429 response. Closes #154.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -255,9 +255,9 @@ to handle it, which defaults to print a traceback and ignoring the exception.
     :type remaining: :class:`int`
     :param reset_after: The amount of time we have to wait before making another request under the same bucket.
     :type reset_after: :class:`float`
-    :param bucket: The hash correlating to the bucket of the request from Discord.
+    :param bucket: The hash correlating to the bucket of the request from Discord. This hash denotes the rate limit being encountered.
     :type bucket: :class:`str`
-    :param scope: If we get a 429, the scope of the 429 response.
+    :param scope: If we get a 429, the scope of the 429 response. This value can either be "user" (rate limit relating to the user) or "shared" (rate limit relating to a resource).
     :type scope: Optional[:class:`str`]
 
 .. function:: on_global_http_ratelimit(retry_after)
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -247,7 +247,7 @@ to handle it, which defaults to print a traceback and ignoring the exception.
 
     If the 429 response is a global ratelimit, then use :func:`on_global_http_ratelimit` instead.
 
-    .. versionadded:: 2.3
+    .. versionadded:: 2.4
 
     :param limit: The amount of requests that have been made under the bucket that the request correlates to.
     :type limit: :class:`int`
@@ -266,6 +266,8 @@ to handle it, which defaults to print a traceback and ignoring the exception.
 
     If the 429 response is a non-global ratelimit or you want to track when the bucket expires,
     then use :func:`on_http_ratelimit` instead.
+
+    .. versionadded:: 2.4
 
     :param retry_after: The amount of time we have to wait before making another request.
     :type retry_after: :class:`float`
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -244,6 +244,7 @@ to handle it, which defaults to print a traceback and ignoring the exception.
 .. function:: on_http_ratelimit(limit, remaining, reset_after, bucket, scope)
 
     Called when a HTTP request in progress either exhausts its bucket or gets a 429 response.
+    For more information on how a ratelimit bucket is defined, check out the [Discord API Docs](https://discord.dev/topics/rate-limits).
 
     If the 429 response is a global ratelimit, then use :func:`on_global_http_ratelimit` instead.
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -245,6 +245,8 @@ to handle it, which defaults to print a traceback and ignoring the exception.
 
     Called when a HTTP request in progress either exhausts its bucket or gets a 429 response.
 
+    If the 429 response is a global ratelimit, then use :func:`on_global_http_ratelimit` instead.
+
     .. versionadded:: 2.3
 
     :param limit: The amount of requests that have been made under the bucket that the request correlates to.
@@ -257,6 +259,16 @@ to handle it, which defaults to print a traceback and ignoring the exception.
     :type bucket: :class:`str`
     :param scope: If we get a 429, the scope of the 429 response.
     :type scope: Optional[:class:`str`]
+
+.. function:: on_global_http_ratelimit(retry_after)
+
+    Called when a HTTP request in progress gets a 429 response and the scope is global.
+
+    If the 429 response is a non-global ratelimit or you want to track when the bucket expires,
+    then use :func:`on_http_ratelimit` instead.
+
+    :param retry_after: The amount of time we have to wait before making another request.
+    :type retry_after: :class:`float`
 
 .. function:: on_ready()
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -241,6 +241,23 @@ to handle it, which defaults to print a traceback and ignoring the exception.
     :param shard_id: The shard ID that has disconnected.
     :type shard_id: :class:`int`
 
+.. function:: on_http_ratelimit(limit, remaining, reset_after, bucket, scope)
+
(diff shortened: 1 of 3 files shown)


nextcord/http.py
```diff
@@ -341,6 +344,14 @@ async def request(
                                 bucket,
                                 delta,
                             )
+                            self._dispatch(
+                                "http_ratelimit",
+                                5,
+                                0,
```
Skelmis, line 350: Why are these hard coded?

nextcord/http.py
```diff
@@ -361,6 +372,15 @@ async def request(
                             retry_after: float = data["retry_after"]
                             _log.warning(fmt, retry_after, bucket)
 
+                            self._dispatch(
+                                "http_ratelimit",
+                                5,
+                                0,
```
Skelmis, line 378: Same here?

Skelmis (commented):
Just looking for some explanation

nextcord/http.py
EmreTech: Well I made them hard coded because the limit is a static value and if we get ratelimited the remaining value should be 0. Maybe since these are static I shouldn't provide them in the event, especially the remaining value.

Skelmis: I was not aware of these. Can you link the resources where the limit is stated to be 5 for all routes?

EmreTech: Well I assumed that because Nextcord/discord.py treats it like a static value in the `request` function (https://github.com/nextcord/nextcord/blob/master/nextcord/http.py#L308).

Skelmis: Thats not the ratelimit...? Thats how many times to retry a request if it fails for any reason

[Ratelimits](https://discord.com/developers/docs/topics/rate-limits) are dynamic, and different per bucket + discord says not to hard code them so we shouldn't where possible. Even on the examples there are multiple where the `limit` is different

EmreTech: Resolved in 79e39a4.

nextcord/http.py
```diff
@@ -330,7 +333,9 @@ async def request(
                         data = await json_or_text(response)
 
                         # check if we have rate limit header information
-                        remaining = response.headers.get("X-Ratelimit-Remaining")
+                        limit = response.headers.get("X-RateLimit-Limit") or "5"
+                        remaining = response.headers.get("X-Ratelimit-Remaining") or "0"
```
Skelmis, line 337:
```suggestion
                        limit = response.headers.get("X-RateLimit-Limit")
                        remaining = response.headers.get("X-Ratelimit-Remaining")
```

Skelmis (changes requested):

nextcord/http.py
Skelmis: Actually hmm, hold this change. I need to look into better handling for global limits

Skelmis:
I feel there are two approaches to also correctly propagating global ratelimits here.
1. We add a param to the current event such as `is_global`
2. We create another event `on_global_http_ratelimit` which only takes params related to global ratelimits

I personally like 2 as it removes a lot of the optional arguments and need to do checks within the event yourself.

Either way, we should conditionally check to see if `X-RateLimit-Global` is a set header, and that it's `True` to determine which event fires with the relevant info.


N.b. Do we also want to expose `X-RateLimit-Scope`?

nextcord/http.py
Skelmis: Okay cool, inline with my new comment I still think this change is valid.

However this info should be for non global ratelimits only

nextcord/http.py
```diff
@@ -341,6 +353,14 @@ async def request(
                                 bucket,
                                 delta,
                             )
+                            self._dispatch(
+                                "http_ratelimit",
+                                int(limit),
+                                int(remaining),
+                                delta,
+                                bucket,
+                                None,
```
Skelmis, line 362:
```suggestion
                                response.headers.get("X-RateLimit-Scope"),
```

ooliver1 (changes requested):
Could the docs perhaps refer to https://discord.com/developers/docs/topics/rate-limits#exceeding-a-rate-limit-rate-limit-response-structure so `scope` and `bucket`'s meanings are fully mentioned

nextcord/http.py
```diff
@@ -330,7 +333,16 @@ async def request(
...
                         # check if we have rate limit header information
-                        remaining = response.headers.get("X-Ratelimit-Remaining")
+                        is_global = bool(response.headers.get("X-RateLimit-Global", False))
+
+                        if is_global:
+                            limit = response.headers.get("X-RateLimit-Limit", "")
+                            remaining = response.headers.get("X-Ratelimit-Remaining", "")
+                        else:
+                            limit = response.headers["X-RateLimit-Limit"]
+                            remaining = response.headers["X-RateLimit-Remaining"]
```
Skelmis, line 343:
```suggestion
                        limit = response.headers.get("X-RateLimit-Limit", "")
                        remaining = response.headers.get("X-Ratelimit-Remaining", "")
```

As this is called for every request, and not all requests are rate limited. For example, logging in will KeyError

nextcord/http.py
```diff
@@ -330,7 +333,16 @@ async def request(
                         data = await json_or_text(response)
 
                         # check if we have rate limit header information
-                        remaining = response.headers.get("X-Ratelimit-Remaining")
+                        is_global = bool(response.headers.get("X-RateLimit-Global", False))
```
Skelmis, line 336: Just plays nicer with types (at-least in pycharm)

```suggestion
                        is_global = bool(response.headers.get("X-RateLimit-Global", ""))
```

Skelmis (changes requested):

nextcord/http.py
EmreTech: Can you elaborate how this plays nicer with types? I don't see how the current code would cause a type error.

Skelmis (approved):
Tested the http ratelimit case, didn't manage to hit the global limit during testing.

Code [here](https://workbin.dev/?id=166661048938684)

ooliver1:
> didn't manage to hit the global limit during testing.
likely since this was in one guild and channel, so the ratelimit scoped to there is smaller than the global.

docs/api.rst
```diff
@@ -241,6 +241,35 @@ to handle it, which defaults to print a traceback and ignoring the exception.
     :param shard_id: The shard ID that has disconnected.
     :type shard_id: :class:`int`
 
+.. function:: on_http_ratelimit(limit, remaining, reset_after, bucket, scope)
+
+    Called when a HTTP request in progress either exhausts its bucket or gets a 429 response.
+
+    If the 429 response is a global ratelimit, then use :func:`on_global_http_ratelimit` instead.
+
+    .. versionadded:: 2.3
```
ooliver1, line 250:
```suggestion
    .. versionadded:: 2.4
```

docs/api.rst
```diff
@@ -241,6 +241,35 @@ to handle it, which defaults to print a traceback and ignoring the exception.
...
+    :type bucket: :class:`str`
+    :param scope: If we get a 429, the scope of the 429 response. This value can either be "user" (rate limit relating to the user) or "shared" (rate limit relating to a resource).
+    :type scope: Optional[:class:`str`]
+
+.. function:: on_global_http_ratelimit(retry_after)
+
+    Called when a HTTP request in progress gets a 429 response and the scope is global.
+
+    If the 429 response is a non-global ratelimit or you want to track when the bucket expires,
+    then use :func:`on_http_ratelimit` instead.
```
ooliver1, line 268:
```suggestion
    then use :func:`on_http_ratelimit` instead.

    .. versionadded:: 2.4
```

ooliver1 (changes requested):
It may make sense to refer to https://discord.dev for ratelimit info on what a "bucket" is.

Got the global limit to hit though:
```
DEBUG:nextcord.http:DELETE https://discord.com/api/v10/channels/1054761482321137715 with None has returned 429
WARNING:nextcord.http:We are being rate limited. Retrying in 0.30 seconds. Handled under the bucket "None"
WARNING:nextcord.http:Global rate limit has been hit. Retrying in 0.30 seconds.
DEBUG:nextcord.client:Dispatching event global_http_ratelimit
WARNING:nextcord.http:retry_after=0.3
```

Skelmis (approved):

ooliver1 (approved):

PR #841, opened 2022-10-09
refactor(abc.py): remove _undefined and _Undefined

### LINKED ISSUE #840: Remove unnecessary `_undefined` sentinel in abc.py
### Summary

Remove the unnecessary `_undefined` sentinel in abc.py.

The core library

### The Problem

Nextcord has a sentinel for when things aren't provided and that's `utils.MISSING`. However, in `abc.py`, there are two sentinels: `utils.MISSING` and a less defined version, `_undefined`. Seeing how there's no reason not to use `utils.MISSING` when `_undefined` is used, `_undefined` isn't needed.

### The Ideal Solution

Remove `_undefined` and its class definition `_Undefined` and replace all references of `_undefined` with `MISSING` (an alias for `utils.MISSING` is made in this file BTW).

### The Current Solution

_No response_

### Additional Context

_No response_

EmreTech:
## Summary

This PR removes the unnecessary `_undefined` sentinel from `abc.py` in favor of `MISSING`. This attribute was used internally, so this will not be a breaking change for users. Closes #840.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -96,14 +96,6 @@ if TYPE_CHECKING:
 MISSING = utils.MISSING
 
 
-class _Undefined:
-    def __repr__(self) -> str:
-        return "see-below"
-
-
-_undefined: Any = _Undefined()
-
-
 @runtime_checkable
 class Snowflake(Protocol):
     """An ABC that details the common operations on a Discord model.
@@ -311,7 +303,7 @@ class GuildChannel:
         payload = []
         for index, c in enumerate(channels):
             d: Dict[str, Any] = {"id": c.id, "position": index}
-            if parent_id is not _undefined and c.id == self.id:
+            if parent_id is not MISSING and c.id == self.id:
                 d.update(parent_id=parent_id, lock_permissions=lock_permissions)
             payload.append(d)
 
@@ -323,7 +315,7 @@ class GuildChannel:
         try:
             parent = options.pop("category")
         except KeyError:
-            parent_id = _undefined
+            parent_id = MISSING
         else:
             parent_id = parent and parent.id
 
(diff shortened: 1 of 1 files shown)


teaishealthy (changes requested):
Instead of aliasing MISSING just `from .utils import MISSING`

EmreTech:
> Instead of aliasing MISSING just `from .utils import MISSING`

I think this would be well suited for another refactor because there are quite a few files that use the same alias.

EmreTech:
> Instead of aliasing MISSING just `from .utils import MISSING`

This will be fixed in #842.

teaishealthy:
> This will be fixed in #842.

I don't see why not fix it here, not doing this in this PR would stall the other PR?

ooliver1:
^, why not import normally in your PR, this would still be properly scoped.

ooliver1:
> > This will be fixed in #842.
> 
> I don't see why not fix it here, not doing this in this PR would stall the other PR?

This PR does not edit the import, so it is not conflicting with the other

EmreTech:
> > > This will be fixed in #842.
> > 
> > 
> > I don't see why not fix it here, not doing this in this PR would stall the other PR?
> 
> This PR does not edit the import, so it is not conflicting with the other

I get it.
Resolved in 2b777c4

teaishealthy (approved):

ooliver1 (approved):

DenverCoder1 (approved):
Should there be an alias of `_undefined` to `MISSING`, similar to the `EmptyEmbed` issue, in order to not make it a breaking change. This could be breaking if someone were to be manually passing `_undefined` to `set_permissions` as a default?

I suppose the thinking is that, as it is underscore-prefixed it's not part of the public API, so that's somebody else's problem. If so, that's fine by me.

EmreTech:
> Should there be an alias of `_undefined` to `MISSING`, similar to the `EmptyEmbed` issue, in order to not make it a breaking change. This could be breaking if someone were to be manually passing `_undefined` to `set_permissions` as a default?
> 
> I suppose the thinking is that, as it is underscore-prefixed it's not part of the public API, so that's somebody else's problem. If so, that's fine by me.

I don't think there should be an alias because yeah it's not under the public API.

PR #842, opened 2022-10-09
refactor: stop importing utils directly

EmreTech:
## Summary

This PR removes all direct imports to `nextcord.utils` in favor of importing what's needed directly. This was done because some newer parts of the code have already made these changes while the older parts directly from discord.py have been unchanged.

## Checklist

- [ ] This PR fixes an issue.


DenverCoder1:
The code seems fine, but what's the purpose? Either way of importing works just fine and it all really comes down to a matter of preference. (Tbh I think the way it was before is better since it makes it more clear where the function is coming from when you use it).

EmreTech:
> The code seems fine, but what's the purpose? Either way of importing works just fine and it all really comes down to a matter of preference. (Tbh I think the way it was before is better since it makes it more clear where the function is coming from when you use it).

It's to be consistent with the rest of the code. Newer, more updated files don't import from utils directly, so why not the rest of the code? I have a lot of PRs and issues that follow the same ideology.

DenverCoder1:
It's fine with me either way. A mix between module imports and function/class imports is pretty common in most Python projects and this is one case where I think there doesn't need to be a convention.

I've looked up some style guides to see some different opinions:

> Use `import` statements for packages and modules only, not for individual classes or functions.
> ~ Google Style Guide https://github.com/google/styleguide/blob/gh-pages/pyguide.md#22-imports

> Import the module using `import numpy` or import individual functions as needed, e.g `from numpy import array, zeros`
> ~ Numpy Style Guide https://numpydoc.readthedocs.io/en/latest/example.html

It seems that Google chooses a convention, being the opposite of this PR, and the way I think is better for readability. Numpy seems to be indifferent and allows either depending on the case. PEP8 similarly does not show any preference to using either way as a convention.

If others decide this is worth merging, that's fine with me, but if it up to me, I'd change the other imports to be module imports rather than the other way around.

Skelmis:
I personally do not sway either way, so I'll happily approve if others want the changes but yea.

PR #853, opened 2022-10-20
feat(app-cmds): add support for 3.10 unions

alentoghostflame:
Signed-off-by: Alex Schoenhofen <alexanderschoenhofen@gmail.com>

## Summary

Adds PEP 604 support to application commands, allowing the Python 3.10 style of unions.
Example: `arg1: str | None`

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -95,6 +95,7 @@ else:
 
 if sys.version_info[1] >= 10:
     from types import UnionType
+
     # UnionType is the annotation origin when doing Python 3.10 unions. Example: "str | None"
 else:
     UnionType = None
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -27,6 +27,7 @@ from __future__ import annotations
 import asyncio
 import inspect
 import logging
+import sys
 import typing
 import warnings
 from inspect import Parameter, signature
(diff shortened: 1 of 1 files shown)


nextcord/application_command.py
```diff
@@ -92,6 +93,13 @@
 else:
     _CustomTypingMetaBase = object
 
+if sys.version_info[1] >= 10:
```
DenverCoder1, line 96: This is more specific and readable. Also supported by conditional Pyright checks I believe (though I'm not 100% sure).

```suggestion
if sys.version_info >= (3, 10):
```

alentoghostflame: Tested on 3.10 and 3.8 and commited.

EmreTech (approved):
Haven't tested
Although, I think the title should be changed to explicitly state that this adds support for 3.10 unions because giving the pep number doesn't showcase this right away.

DenverCoder1 (approved):
Haven't tested but looks good to me.

nextcord/application_command.py
```diff
@@ -92,6 +93,13 @@
 else:
     _CustomTypingMetaBase = object
 
+if sys.version_info >= (3, 10):
```
DenverCoder1, line 96: I think `utils.PY_310` can be used for this. I just came across it and it looks like it's exactly this condition but in a constant.

teaishealthy (approved):

ooliver1 (approved):
Tested with channel type unions.

PR #858, opened 2022-10-21
fix(abc): handle timeouts in GuildChannel.permissions_for

### LINKED ISSUE #852: channel.permissions_for does not take timeouts into account
### Summary

Timeouts can refuse practically every permission on a channel, `permissions_for` does not do so.

### Reproduction Steps

1. Timeout bot
2. use `permissions_for`
3. `send_messages` is still `True`

### Minimal Reproducible Code

_No response_

### Expected Results

Timeouts to be considered by nextcord so no local checking is required.

### Actual Results

Timeouts are not taken into account

### Intents

`guilds`, `guild_messages`, `voice_states`

### System Information

- Python v3.10.7-final
- nextcord v2.2.0-final
- aiohttp v3.8.3
- system info: Linux 5.15.0-47-generic #​51-Ubuntu SMP Thu Aug 11 07:51:15 UTC 2022

### Checklist

### Additional Context

_No response_

e-codes-stuff:
## Summary

Fixes #852

## Checklist

  - [ ] I have updated the documentation to reflect the changes. (N/A, fixes implicit behavior)
- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -629,7 +629,6 @@ class GuildChannel:
 
             return base
 
-        # Handle Member case
         roles = obj._roles
         get_role = self.guild.get_role
 
@@ -686,11 +685,11 @@ class GuildChannel:
             base.value &= ~denied.value
 
         # if you are timed out then you lose all permissions except view_channel and read_message_history
-        if obj.communication_disabled_until != None:
-            denied = Permissions.none()
-            denied.view_channel = True
-            denied.read_message_history = True
-            base.value &= denied.value
+        if obj.communication_disabled_until is not None:
+            allowed = Permissions.none()
+            allowed.view_channel = True
+            allowed.read_message_history = True
+            base.value &= allowed.value
 
         return base
 
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -629,6 +629,7 @@ class GuildChannel:
 
             return base
 
+        # Handle Member case
         roles = obj._roles
         get_role = self.guild.get_role
 
@@ -683,6 +684,13 @@ class GuildChannel:
         if not base.read_messages:
             denied = Permissions.all_channel()
             base.value &= ~denied.value
+
+        # if you are timed out then you lose all permissions except view_channel and read_message_history
+        if obj.communication_disabled_until != None:
+            denied = Permissions.none()
+            denied.view_channel = True
+            denied.read_message_history = True
+            base.value &= denied.value
 
         return base
 
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -686,9 +686,7 @@ class GuildChannel:
 
         # if you are timed out then you lose all permissions except view_channel and read_message_history
         if obj.communication_disabled_until is not None:
-            allowed = Permissions.none()
-            allowed.view_channel = True
-            allowed.read_message_history = True
+            allowed = Permissions(view_channel=True, read_message_history=True)
             base.value &= allowed.value
 
         return base
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -570,6 +570,9 @@ class GuildChannel:
         - The default role permission overwrites
         - The permission overwrites of the role used as a parameter
 
+        .. versionchanged:: 2.3
+            Only ``view_channel`` and ``read_message_history`` can be returned for timed-out members
+
         .. versionchanged:: 2.0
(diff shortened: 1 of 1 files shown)


nextcord/abc.py
```diff
@@ -629,6 +629,7 @@ def permissions_for(self, obj: Union[Member, Role], /) -> Permissions:
 
             return base
 
+        # Handle Member case
```
EmreTech, line 632: What is this comment for?

nextcord/abc.py
```diff
@@ -684,6 +685,13 @@ def permissions_for(self, obj: Union[Member, Role], /) -> Permissions:
             denied = Permissions.all_channel()
             base.value &= ~denied.value
 
+        # if you are timed out then you lose all permissions except view_channel and read_message_history
+        if obj.communication_disabled_until != None:
```
EmreTech, line 689:
```suggestion
        if obj.communication_disabled_until:
```

EmreTech (commented):
Looks good otherwise

nextcord/abc.py
ooliver1: or `is not None` for explicitness (`!=` and `==` shouldn't be used for `None` since it is a sentinel)

nextcord/abc.py
```diff
@@ -684,6 +685,13 @@ def permissions_for(self, obj: Union[Member, Role], /) -> Permissions:
...
+            denied = Permissions.none()
+            denied.view_channel = True
+            denied.read_message_history = True
+            base.value &= denied.value
```
ooliver1, line 693: `denied` doesn't really represent what this does, since those are the only allowed permissions

ooliver1 (changes requested):

nextcord/abc.py
e-codes-stuff: I added it as a way to more easily scope the part of the code I was reading but I will remove it if it seems unnecessary

nextcord/abc.py
e-codes-stuff: I'll change it to `is not None` since I think being explicit seems like the right choice here

nextcord/abc.py
```diff
@@ -684,6 +684,13 @@ def permissions_for(self, obj: Union[Member, Role], /) -> Permissions:
             denied = Permissions.all_channel()
             base.value &= ~denied.value
 
+        # if you are timed out then you lose all permissions except view_channel and read_message_history
+        if obj.communication_disabled_until is not None:
+            allowed = Permissions.none()
+            allowed.view_channel = True
+            allowed.read_message_history = True
```
DenverCoder1, line 691: Looks good :+1: I might suggest making this more concise

```suggestion
            allowed = Permissions(view_channel=True, read_message_history=True)
```

DenverCoder1 (approved):

nextcord/abc.py
```diff
@@ -684,6 +684,11 @@ def permissions_for(self, obj: Union[Member, Role], /) -> Permissions:
             denied = Permissions.all_channel()
             base.value &= ~denied.value
 
+        # if you are timed out then you lose all permissions except view_channel and read_message_history
+        if obj.communication_disabled_until is not None:
```
DenverCoder1, line 688: Has this been tested with obj being a `Role` object? I feel this might cause a problem.

e-codes-stuff: This code will never be run on a `Role` object, the comment I added was removed earlier, but this code handles the `Member` case only. `Role` is handled earlier and always returns earlier.

DenverCoder1: Tested and seems fine. Role returns early, so does not reach this code.

DenverCoder1 (approved):
Documentation suggestion:

```diff
        This function takes into consideration the following cases:

        - Guild owner
        - Guild roles
        - Channel overrides
        - Member overrides
+       - Timed out members
```

Besides that, looks good to me :+1:

DenverCoder1:
Just one more small suggestion, you can add a version directive like this above the 2.0 version changed.
```diff
+       .. versionchanged:: 2.3
+           Only ``view_channel`` and ``read_message_history`` can be returned for timed-out members

        .. versionchanged:: 2.0
            The object passed in can now be a role object.
```

e-codes-stuff:
That should be everything, tell me if the code needs any other changes

ooliver1 (approved):

PR #863, opened 2022-10-23
docs: Fix single-backtick words to double-backtick or links

DenverCoder1:
## Summary

Made 100+ fixes to single-backtick words in rst files to double-backtick phrases or reference links depending on the case.

This helps with readability since single-backtick phrases are italicized rather than set in inline-codeblocks. 

A few additional small docs fixes were also made (eg. missing colon, etc.) in certain cases.

image

image

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -593,7 +593,7 @@ class CallbackMixin:
 
         Parameters
         ----------
-        callback: Optional[:class:`Callable`]
+        callback: Optional[:data:`~typing.Callable`]
             Callback to create options from and invoke. If provided, it must be a coroutine function.
         parent_cog: Optional[:class:`ClientCog`]
             Class that the callback resides on. Will be passed into the callback if provided.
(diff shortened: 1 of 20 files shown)


Skelmis (approved):

nextcord/application_command.py
```diff
@@ -593,7 +593,7 @@ def __init__(self, callback: Optional[Callable] = None, parent_cog: Optional[Cli
 
         Parameters
         ----------
-        callback: Optional[Callable]
+        callback: Optional[:class:`Callable`]
```
ooliver1, line 596: Nearly every other reference to typing.* doesn't use links so the types in the docstrings are shorter.

nextcord/application_command.py
```diff
@@ -970,7 +970,7 @@ def __init__(
 
         Parameters
         ----------
-        autocomplete_callback: `Callable`
+        autocomplete_callback: Optional[:class:`Callable`]
```
ooliver1, line 973: Same as above.

ooliver1 (changes requested):

teaishealthy (approved):

ooliver1 (approved):

PR #867, opened 2022-10-26
fix: Python3.11 dataclass changes

### LINKED ISSUE #866: Importing commands raises ValueError in Python 3.11
### Summary

Importing the commands module in Python 3.11 raises a ValueError

### Reproduction Steps

Run the `from nextcord.ext import commands` in a Python 3.11 environment with either the latest stable or latest development build of nextcord installed.

### Minimal Reproducible Code

```python

pass
```

### Expected Results

`commands` module gets successfully imported into the current running environment, as happens under Python 3.10 and earlier

### Actual Results

The following ValueError is raised
```
  File "<string>", line 1, in <module>
  File "C:\Users\User\AppData\Roaming\Python\Python311\site-packages\nextcord\ext\commands\__init__.py", line 19, in <module>
    from .flags import *
  File "C:\Users\User\AppData\Roaming\Python\Python311\site-packages\nextcord\ext\commands\flags.py", line 70, in <module>
    @dataclass
     ^^^^^^^^^
  File "C:\Program Files\Python311\Lib\dataclasses.py", line 1221, in dataclass
    return wrap(cls)
           ^^^^^^^^^
  File "C:\Program Files\Python311\Lib\dataclasses.py", line 1211, in wrap
    return _process_class(cls, init, repr, eq, order, unsafe_hash,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\Python311\Lib\dataclasses.py", line 959, in _process_class
    cls_fields.append(_get_field(cls, name, type, kw_only))
                      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "C:\Program Files\Python311\Lib\dataclasses.py", line 816, in _get_field
    raise ValueError(f'mutable default {type(f.default)} for field '
ValueError: mutable default <class 'nextcord.utils._MissingSentinel'> for field name is not allowed: use default_factory

```

### Intents

All

### System Information

- Python v3.11.0-final
- nextcord v2.3.0-alpha
    - nextcord pkg_resources: v2.3.0a4137+g6e0999d1
- aiohttp v3.8.3
- system info: Windows 10 10.0.22621

### Checklist

### Additional Context

_No response_

Skelmis:
## Summary

This closes #866 by migrating the syntax for ``flags.py`` inline with changes to Python 3.11

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/ext/commands/flags.py
+++ b/nextcord/ext/commands/flags.py
@@ -94,13 +94,13 @@ class Flag:
         Whether multiple given values overrides the previous value.
     """
 
-    name: str = MISSING
+    name: str = field(default_factory=MISSING)
     aliases: List[str] = field(default_factory=list)
-    attribute: str = MISSING
-    annotation: Any = MISSING
-    default: Any = MISSING
-    max_args: int = MISSING
-    override: bool = MISSING
+    attribute: str = field(default_factory=MISSING)
+    annotation: Any = field(default_factory=MISSING)
+    default: Any = field(default_factory=MISSING)
+    max_args: int = field(default_factory=MISSING)
+    override: bool = field(default_factory=MISSING)
     cast_to_dict: bool = False
 
     @property
--- a/nextcord/ext/commands/flags.py
+++ b/nextcord/ext/commands/flags.py
@@ -94,13 +94,13 @@ class Flag:
         Whether multiple given values overrides the previous value.
     """
 
-    name: str = field(default_factory=MISSING)
+    name: str = MISSING
     aliases: List[str] = field(default_factory=list)
-    attribute: str = field(default_factory=MISSING)
-    annotation: Any = field(default_factory=MISSING)
-    default: Any = field(default_factory=MISSING)
-    max_args: int = field(default_factory=MISSING)
-    override: bool = field(default_factory=MISSING)
+    attribute: str = MISSING
+    annotation: Any = MISSING
+    default: Any = MISSING
+    max_args: int = MISSING
+    override: bool = MISSING
     cast_to_dict: bool = False
 
     @property
--- a/nextcord/utils.py
+++ b/nextcord/utils.py
@@ -365,7 +365,7 @@ def time_snowflake(dt: datetime.datetime, high: bool = False) -> int:
         The snowflake representing the time given.
     """
     discord_millis = int(dt.timestamp() * 1000 - DISCORD_EPOCH)
-    return (discord_millis << 22) + (2 ** 22 - 1 if high else 0)
+    return (discord_millis << 22) + (2**22 - 1 if high else 0)
 
 
 def find(predicate: Callable[[T], Any], seq: Iterable[T]) -> Optional[T]:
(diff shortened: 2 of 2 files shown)


DenverCoder1:
Looks good assuming tested, but any idea why it seems `_MissingSentinel` is considered a mutable default?

Skelmis:
In 3.11 defaults must be hashable which it is not.

[Change log](https://docs.python.org/3.11/whatsnew/3.11.html#dataclasses)

DenverCoder1:
Is there any harm in making it hashable?

Skelmis:
Possibly yes, as objects with the same hash should return equal. Given ``MISSING`` always returns ``False`` for comparisons it would be a change in behaviour

DenverCoder1:
> Possibly yes, as objects with the same hash should return equal. Given `MISSING` always returns `False` for comparisons it would be a change in behaviour

I think it would make sense if `MISSING == MISSING` returning `True`. I don't really see any potential harm in changing that behavior.

```py
def __eq__(self, other):
    return self is other

def __hash__(self):
    return id(self)
```
(Code suggested by GitHub Copilot)

I think it would be a bit of a stretch to consider it a breaking change. Would that work?

Skelmis:
I can certainly give it a go later on and see what happens

TheGhostOfInky:
I tried both fixes and both seem to solve the issue, making `MISSING` hashable certainly seems like a better solution since having to assign a constant as a mutable is counter-productive. It also would break user code that uses `MISSING` in dataclasses upon upgrading to 3.11, unless they're already aware of the issue.

teaishealthy (changes requested):
`default_factory` must be a callable that returns the object to be used (as indicated by `_factory`). This code would fail with `TypeError: '_MissingSentinel' object is not callable` if any argument was not passed to the dataclass.

ooliver1:
@MaskDuck did you see tea's comment? It looks like you randomly approved

Skelmis:
I have modified the changes to instead have ``MISSING`` be hashable, a quick search in files appears to show this modification should not cause any issues internally.

Note: Given the file has now been touched the latter change relates to `black` being run on it

DenverCoder1 (approved):

teaishealthy (approved):

ooliver1 (approved):

lol219:
Does this mean that nextcord is now fixed and supports 3.11 in this PR.

TheGhostOfInky:
Yes, it does, I'm using it with the repo's current code and it works.

lol219:
> Yes, it does, I'm using it with the repo's current code and it works.

Thanks

DenverCoder1:
> Does this mean that nextcord is now fixed and supports 3.11 in this PR.

I'd recommend not using it with 3.11 for a while. While it may work for the most part, there may still be some issues since it's brand new.

We are also seeing many type issues still which could potentially cause problems (see https://github.com/nextcord/nextcord/pull/868)

lol219:
@DenverCoder1 got it

PR #869, opened 2022-10-26
typing(abc): type set_permissions parameters

### LINKED ISSUE #839: set_permissions has untyped parameters
### Summary

`GuildChannel.set_permissions` has untyped parameters in the actual implementation of the function.

The core library

### The Problem

As I have discussed in past issues, it's important to type as much as possible so static type checkers can fully understand the code we're using and give appropriate errors if necessary.

### The Ideal Solution

Type the parameters in the actual implementation of `GuildChannel.set_permissions` (i.e. not the overloads).

### The Current Solution

_No response_

### Additional Context

_No response_

EmreTech:
## Summary

This PR types the parameters of the actual implementation (aka not overloaded version) of `GuildChannel.set_permissions`. The actual implementation's parameters are changed slightly to include a catch-all variable keyword parameter because of how different the two other overloads are.

Closes #839. 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -834,7 +834,7 @@ class GuildChannel:
         elif isinstance(overwrite, PermissionOverwrite):
             (allow, deny) = overwrite.pair()
             await http.edit_channel_permissions(
-                self.id, target.id, allow.value, deny.value, perm_type, reason=reason
+                self.id, target.id, str(allow.value), str(deny.value), perm_type, reason=reason
             )
         else:
             raise InvalidArgument("Invalid overwrite type provided.")
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -731,10 +731,10 @@ class GuildChannel:
         ...
 
     async def set_permissions(
-        self, 
-        target: Union[Member, Role], 
-        *, 
-        reason: Optional[str] = None, 
+        self,
+        target: Union[Member, Role],
+        *,
+        reason: Optional[str] = None,
         **kwargs: Any,
     ):
         r"""|coro|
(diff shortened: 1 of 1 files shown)


nextcord/abc.py
```diff
@@ -730,7 +730,13 @@ async def set_permissions(
     ) -> None:
         ...
 
-    async def set_permissions(self, target, *, overwrite=MISSING, reason=None, **permissions):
+    async def set_permissions(
+        self,
+        target: Union[Member, Role],
+        *,
+        reason: Optional[str] = None,
+        **kwargs: Any,
```
ooliver1, line 738: They are both optional and keyword only args, so
```py
overwrite: Optional[PermissionOverwrite] = MISSING,
**permissions: bool,
```
should work just fine right?

ooliver1 (changes requested):

teaishealthy (approved):
Just ooli's comment else lgtm 👍

nextcord/abc.py
EmreTech: My apologies for not elaborating on why I used a catch-all variable keyword parameter.
You see, if I add what you're suggesting, I get this type error: 
```
Overloaded implementation is not consistent with signature of overload 2
  Type "(self: Self@GuildChannel, target: Member | Role, *, reason: str | None = None, overwrite: PermissionOverwrite | None = MISSING, **permissions: bool) -> Coroutine[Any, Any, None]" cannot be assigned to type "(self: Self@GuildChannel, target: Member | Role, *, reason: str | None = ..., **permissions: bool) -> Coroutine[Any, Any, None]"
    Parameter 5: type "bool" cannot be assigned to type "PermissionOverwrite | None"
      Type "bool" cannot be assigned to type "PermissionOverwrite | None"
        "bool" is incompatible with "PermissionOverwrite"
        Type cannot be assigned to type "None"
```

ooliver1 (approved):

PR #870, opened 2022-10-27
refactor(guild): use Optional for limit kwargs

ToasterUwU:
## Summary

Add the `Optional` annotation for limit in two cases, since it was both times type hinted to be a int, but also accepted None for no limit.

Testing is not needed, since its just annotation for values that already are used for it.

## Checklist

- [X] This PR fixes an issue.

### DIFF
--- a/nextcord/guild.py
+++ b/nextcord/guild.py
@@ -3036,7 +3036,7 @@ class Guild(Hashable):
 
         Parameters
         ----------
-        limit: :class:`int`
+        limit: Optional[:class:`int`]
             The number of entries to retrieve. If ``None`` retrieve all entries.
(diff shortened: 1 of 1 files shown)


EmreTech (approved):

DenverCoder1:
The docstring for audit_logs should probably also be corrected

ToasterUwU:
> The docstring for audit_logs should probably also be corrected

Agreed, totally overlooked that one.

ToasterUwU:
@DenverCoder1 Think im done. Docstrings are both right, no other guild.py limit argument was a Optional with wrong annotation.

DenverCoder1 (approved):

EmreTech (approved):

TAG-Epic (approved):

ooliver1 (approved):

PR #877, opened 2022-10-31
refactor: correctly type utils

ooliver1:
## Summary

Improves public facing types from nextcord.utils to be fully annotated.
Sets pyright to strict (apart from rules which cause errors from `array.array`).
`pyright --verifytypes nextcord` is now improved by .9% :sweat_smile:

## Checklist

- [x] This PR fixes an issue.


Skelmis:
Does not also have the side effect of marking some of these utils now as public API? At-least per the documented reasons

ooliver1:
Yeah, probably, they were made public since `utils._*` outside of `utils`, and `from utils import _*` are both using private members outside of where it is defined.

Skelmis:
Fair enough, I don't see much of an issue with making them public

Skelmis (approved):

DenverCoder1 (approved):

teaishealthy (approved):

PR #880, opened 2022-11-01
refactor(ui): improve public annotations

ooliver1:
## Summary

This improves public-facing annotations in the ui module.
The overall pyright type completeness score increases by 2.3%.
Unlike #877, this does not use `pyright: strict`, since there are significant archtectual changes required to get strict to be happy, such as importing and using private members, and using raw `**kwargs` for 'underlying' objects.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/components.py
+++ b/nextcord/components.py
@@ -313,9 +313,9 @@ class SelectOption:
         emoji: Optional[Union[str, Emoji, PartialEmoji]] = None,
         default: bool = False,
     ) -> None:
-        self.label = label
-        self.value = label if value is MISSING else value
-        self.description = description
+        self.label: str = label
+        self.value: str = label if value is MISSING else value
+        self.description: Optional[str] = description
 
         if emoji is not None:
             if isinstance(emoji, str):
@@ -327,8 +327,8 @@ class SelectOption:
                     f"Expected emoji to be str, Emoji, or PartialEmoji not {emoji.__class__}"
                 )
 
-        self.emoji = emoji
-        self.default = default
+        self.emoji: Optional[PartialEmoji] = emoji
+        self.default: bool = default
 
     def __repr__(self) -> str:
         return (
--- a/nextcord/ui/button.py
+++ b/nextcord/ui/button.py
@@ -40,9 +40,9 @@ __all__ = (
 
 if TYPE_CHECKING:
     from ..emoji import Emoji
+    from ..interactions import ClientT
     from ..types.components import ButtonComponent as ButtonComponentPayload
     from .view import View
-
 
 B = TypeVar("B", bound="Button")
 V = TypeVar("V", bound="View", covariant=True)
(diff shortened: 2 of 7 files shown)


nextcord/components.py
```diff
@@ -313,9 +313,9 @@ def __init__(
         emoji: Optional[Union[str, Emoji, PartialEmoji]] = None,
         default: bool = False,
     ) -> None:
-        self.label = label
-        self.value = label if value is MISSING else value
-        self.description = description
+        self.label: str = label
+        self.value: str = label if value is MISSING else value
+        self.description: Optional[str] = description
```
DenverCoder1, line 318: I don't think annotations are necessary here since these types can be inferred from the annotations in the arguments

nextcord/ui/button.py
```diff
@@ -134,7 +134,7 @@ def style(self) -> ButtonStyle:
         return self._underlying.style
 
     @style.setter
-    def style(self, value: ButtonStyle):
+    def style(self, value: ButtonStyle) -> None:
```
DenverCoder1, line 137: I don't think it's necessary to annotate return types of functions that do not have return statements. It can be inferred that the return type is None.

DenverCoder1 (approved):
Everything looks accurate, but some parts seem unnecessary.

nextcord/components.py
ooliver1:
```
nextcord.ui.button.Button.label
  /.../nextcord/ui/button.py:error: Return type annotation is missing
```
pyright thought otherwise (`pyright --verifytypes nextcord`)

nextcord/ui/button.py
ooliver1:
```
nextcord.ui.button.Button.label
  /.../nextcord/ui/button.py:error: Return type annotation is missing
```
pyright thought otherwise (`pyright --verifytypes nextcord`)

nextcord/components.py
ooliver1: I thought so too, pyright thought otherwise, maybe its so if the source type changes, the attribute type errors out, not sure.
```
  /.../nextcord/components.py:316:14 - error: Type is missing type annotation and could be inferred differently by type checkers
    Inferred type is "str"
```

DenverCoder1: I find that strange, it's usually better to avoid writing the same thing twice since that can lead to more errors if it needs to change.

nextcord/ui/button.py
DenverCoder1: Are there ways to tell it not to check those rules? It seems weird to annotate None when the function doesn't return anything.

nextcord/components.py
teaishealthy: Pyright is a much more powerful type checker than for example pyre or mypy, additionally those are fundamentally implemented differently which could cause them to infer those types differently.

nextcord/ui/button.py
teaishealthy: Similar reasoning to other comment, but there is a tool to do this automatically  [autotyping](https://pypi.org/projects/autotyping)

teaishealthy (approved):

PR #886, opened 2022-11-05
feat(View): add prevent_update param

TrapDrap:
## Summary

Adding prevent_update to View. Option preventing storing permanent views for separate messages by default.
(since there's no limit to how many different buttons get stored, and if not stopped properly can cause issues over time when running a bot for long periods without reboot, and users are not notified of these being stored)

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/ui/view.py
+++ b/nextcord/ui/view.py
@@ -164,7 +164,13 @@ class View:
 
         cls.__view_children_items__ = children
 
-    def __init__(self, *, timeout: Optional[float] = 180.0, auto_defer: bool = True, prevent_update: bool = True) -> None:
+    def __init__(
+        self,
+        *,
+        timeout: Optional[float] = 180.0,
+        auto_defer: bool = True,
+        prevent_update: bool = True,
+    ) -> None:
         self.timeout = timeout
         self.auto_defer = auto_defer
         self.prevent_update = True if timeout else prevent_update
--- a/nextcord/ui/view.py
+++ b/nextcord/ui/view.py
@@ -138,6 +138,13 @@ class View:
         Whether or not to automatically defer the component interaction when the callback
         completes without responding to the interaction. Set this to ``False`` if you want to
         handle view interactions outside of the callback.
+    prevent_update: :class:`bool` = True
+        This option only affects persistent views.
+        Whether or not to store the view separately for each message.
+        The stored views are not automatically cleared, and can cause issues if
+        you run the bot continously for long periods of time if views are not properly stopped.
+        Setting this to False will force the client to find a persistent view added with
+        `Bot.add_view` and not store the view separately.
     """
 
     __discord_ui_view__: ClassVar[bool] = True
@@ -146,18 +153,21 @@ class View:
     def __init_subclass__(cls) -> None:
         children: List[ItemCallbackType] = []
         for base in reversed(cls.__mro__):
-            for member in base.__dict__.values():
-                if hasattr(member, "__discord_ui_model_type__"):
-                    children.append(member)
+            children.extend(
+                member
+                for member in base.__dict__.values()
+                if hasattr(member, "__discord_ui_model_type__")
+            )
 
         if len(children) > 25:
             raise TypeError("View cannot have more than 25 children")
 
         cls.__view_children_items__ = children
 
-    def __init__(self, *, timeout: Optional[float] = 180.0, auto_defer: bool = True) -> None:
+    def __init__(self, *, timeout: Optional[float] = 180.0, auto_defer: bool = True, prevent_update: bool = True) -> None:
         self.timeout = timeout
         self.auto_defer = auto_defer
+        self.prevent_update = True if timeout else prevent_update
         self.children: List[Item] = []
         for func in self.__view_children_items__:
             item: Item = func.__discord_ui_model_type__(**func.__discord_ui_model_kwargs__)
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -1554,7 +1554,7 @@ class Messageable:
             )
 
         ret = state.create_message(channel=channel, data=data)
-        if view:
+        if view and view.prevent_update:
             state.store_view(view, ret.id)
 
         if delete_after is not None:
--- a/nextcord/interactions.py
+++ b/nextcord/interactions.py
@@ -443,7 +443,7 @@ class Interaction(Hashable, Generic[ClientT]):
 
         # The message channel types should always match
         message = InteractionMessage(state=self._state, channel=self.channel, data=data)  # type: ignore
-        if view and not view.is_finished():
+        if view and not view.is_finished() and view.prevent_update:
             self._state.store_view(view, message.id)
         return message
 
@@ -906,7 +906,7 @@ class InteractionResponse:
                 for file in files:
                     file.close()
 
-        if view is not MISSING:
+        if view is not MISSING and view.prevent_update:
             if ephemeral and view.timeout is None:
                 view.timeout = 15 * 60.0
 
@@ -1077,7 +1077,7 @@ class InteractionResponse:
                 for file in files:
                     file.close()
 
-        if view and not view.is_finished() and message_id is not None:
+        if view and not view.is_finished() and message_id is not None and view.prevent_update:
             state.store_view(view, message_id)
 
         self._responded = True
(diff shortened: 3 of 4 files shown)


EmreTech:
If the `auto_stop` parameter limits storing persistent views for specific messages, `force_global` or just `global` would be a better name.

EmreTech:
Also, to avoid breaking changes, the default for `auto_stop` or whatever the parameter will end up being named should be `False`.

TrapDrap:
> If the `auto_stop` parameter limits storing persistent views for specific messages, `force_global` or just `global` would be a better name.

I'm thinking something like `use_persistent` or `persistent` could also work.

EmreTech:
> > If the `auto_stop` parameter limits storing persistent views for specific messages, `force_global` or just `global` would be a better name.
> 
> I'm thinking something like `use_persistent` or `persistent` could also work.

That doesn't make sense imo. The whole point of this PR is to limit storing persistent views with a specific message, right?

TrapDrap:
> Also, to avoid breaking changes, the default for `auto_stop` or whatever the parameter will end up being named should be `False`.

Well, the main part of this is to fix the issue of views being stored without the user's knowledge, since most people probably call their views without knowing they need to properly stop it, it being set to True makes sense. Will wait on more input tho.

EmreTech:
> Well, the main part of this is to fix the issue of views being stored without the user's knowledge

Can you elaborate how this happens? Also, wouldn't more precise docs be better instead of this breaking change?

TrapDrap:
> > Well, the main part of this is to fix the issue of views being stored without the user's knowledge
> 
> Can you elaborate how this happens? Also, wouldn't more precise docs be better instead of this breaking change?

On every message sent with a view, all components are added to a list specific to that message. Over time, these are not deleted, even when the messages are uncached, accumulating in the background. It's a very rare scenario you would need to use message-specific data if you're using a persistent view, so it's wouldn't be a noticeable breaking change to the large majority of View users.

EmreTech:
> > > Well, the main part of this is to fix the issue of views being stored without the user's knowledge
> > 
> > 
> > Can you elaborate how this happens? Also, wouldn't more precise docs be better instead of this breaking change?
> 
> On every message sent with a view, all components are added to a list specific to that message. Over time, these are not deleted, even when the messages are uncached, accumulating in the background. It's a very rare scenario you would need to use message-specific data if you're using a persistent view, usually, you get all the message data from `interaction.message`, so it wouldn't be a noticeable breaking change to the large majority of View users.

Ah I understand now.

TrapDrap:
that's really weird, changing the var and it's suddenly failing

TrapDrap:
nevermind it's just a the python global, guess cant use "global" as name

EmreTech:
> nevermind it's just a the python global, guess cant use "global" as name

You can, but if you use the builtin global function anywhere in that file, you have to import the `builtins` module and use `builtins.global`.

EmreTech:
Ignore my last comment, I confused the `global` keyword with `globals`.

TrapDrap:
hold on i used the wrong version 💀

TrapDrap:
I changed the name to `prevent_update` to match with the already existing ...prevent_update... function when setting the message-specific view

nextcord/ui/view.py
```diff
@@ -177,9 +184,16 @@ def __init_subclass__(cls) -> None:
...
+    def __init__(
+        self,
+        *,
+        timeout: Optional[float] = 180.0,
+        auto_defer: bool = True,
+        prevent_update: bool = False,
+    ):
         self.timeout = timeout
         self.auto_defer = auto_defer
+        self.prevent_update = prevent_update if not timeout else True
```
Skelmis, line 196: By default this will always be enabled? I thought it was going to be disabled by default for backwards compatibility

TrapDrap: This feat is supposed to be a bug fix as well. In most if not all cases, you don't want a persistent view to be individually stored for each message, as the view is persistent for a reason: so you don't have to create a view for each message, and that it works across reboots. This feat should fix memory leaks caused by views unknowingly being stored in ViewStore and not being deleted because the view timeout is `None`. By default, this being enabled *should* be backwards compatible to anyone correctly storing their persistent views with `bot.add_view`, and shouldn't affect those using non-persistent views.

ooliver1: This is not backwards compatible though. Since they may use attributes on the view for the first lifecycle. You may say "they shouldn't do that", but it is still possible and is used. This change in behaviour is a breaking change and should be opt-in.

TrapDrap: I've changed it to default to opt-in. 
But that kind of defeats the original purpose of this PR if they don't enable it. Why would a breaking change requiring a small amount of people to fix their code be bad in order to fix a memory leak issue, arguably a larger issue

ooliver1: Because it isn't a "small amount of people". You cannot base how many users use a feature based off your own experience, since the majority of nextcord users are not even in the Discord server.

nextcord/ui/view.py
```diff
@@ -168,18 +175,27 @@ class View:
...
+    def __init__(
+        self,
+        *,
+        timeout: Optional[float] = 180.0,
+        auto_defer: bool = True,
+        prevent_update: bool = True,
+    ):
         self.timeout = timeout
         self.auto_defer = auto_defer
+        self.prevent_update = True if timeout else prevent_update
```
Skelmis, line 198: Simply changing the default value for `prevent_update` and then **also** modifying the if statement so you in fact have not changed the default for this variable when you said you had is not cool.

Skelmis (approved):

nextcord/ui/view.py
```diff
@@ -160,6 +160,13 @@ class View:
         Whether or not to automatically defer the component interaction when the callback
         completes without responding to the interaction. Set this to ``False`` if you want to
         handle view interactions outside of the callback.
+    prevent_update: :class:`bool` = True
+        This option only affects persistent views.
```
toolifelesstocode, line 164: Might want to put the line somewhere towards the bottom of this document

nextcord/ui/view.py
```diff
@@ -160,6 +160,13 @@ class View:
         Whether or not to automatically defer the component interaction when the callback
         completes without responding to the interaction. Set this to ``False`` if you want to
         handle view interactions outside of the callback.
+    prevent_update: :class:`bool` = True
```
toolifelesstocode, line 163: Usually the default is mentioned in the actual documentation and not by the name. In other words:
```suggestion
    prevent_update: :class:`bool`
```

toolifelesstocode (changes requested):

toolifelesstocode:
Also correct me if I'm wrong, but I don't think this exactly fixes an issue; And also this is not a breaking change since you're just adding a parameter.

image

TrapDrap:
> Also correct me if I'm wrong, but I don't think this exactly fixes an issue; And also this is not a breaking change since you're just adding a parameter.
> 
> image

Ah yes, some of those checks are outdated. It use to be a breaking change & fixes an issue but now is an optional param.

nextcord/ui/view.py
TrapDrap: What do you mean? I don't get it. The other docs show the default there as well.

nextcord/ui/view.py
TrapDrap: I think it's important to let users know at first glance that this is a parameter used for persistent views. If they're not using persistent views, saves some reading

nextcord/ui/view.py
EmreTech: This is best left for another PR since the surrounding documentation in this module uses the same style for parameter defaults.

nextcord/ui/view.py
EmreTech: I don't think that really changes anything since the table of contents for a class is organized alphabetically.

toolifelesstocode: I meant the `This option only affects persistent views.` part specifically, GitHub just made it look like I highlighted the 2 lines

TrapDrap: I intentionally put it at the top. As aforementioned, I believe it is important to address that it is clear to the user that it should only affect persistent views, and so non-persistent view users wouldn't need to know the rest of the parameter's description before knowing that it isn't necessary for them.

TrapDrap:
I have no idea why there's a conflict.

TrapDrap:
Reopen, still adding back changes

EmreTech (approved):

PR #902, opened 2022-11-12
Interactions in converters

xXenvy:
## Summary

Pull request adding interaction support to converters

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/ext/commands/converter.py
+++ b/nextcord/ext/commands/converter.py
@@ -50,8 +50,8 @@ from .errors import *
 if TYPE_CHECKING:
     from typing import Callable
 
-    from nextcord.member import Member
     from nextcord.interactions import Interaction
+    from nextcord.member import Member
     from nextcord.message import PartialMessageableChannel
     from nextcord.user import User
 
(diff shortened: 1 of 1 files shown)


nextcord/ext/commands/converter.py
```diff
@@ -284,17 +288,21 @@ class UserConverter(IDConverter[nextcord.User]):
...
         state = ctx._state
 
+        if isinstance(ctx, nextcord.Interaction):
+            bot = ctx.client
+        else:
+            bot = ctx.bot
         if match is not None:
             user_id = int(match.group(1))
-            result = ctx.bot.get_user(user_id) or _utils_get(ctx.message.mentions, id=user_id)
+            result = bot.get_user(user_id) or _utils_get(ctx.message.mentions, id=user_id)
```
ooliver1, line 302: `ctx.message` in interactions is different to `Context`, and may be `None`.

EmreTech:
Why is this needed? Application Commands have their own converters.

ooliver1 (changes requested):
These might do well with a [`.. versionchanged`](https://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-versionchanged).

ooliver1:
It might be better to utilise `OptionConverter` since that is *designed* for application commands, instead of retroactively applying to `ext.commands`.

ooliver1:
>  [x] If code changes were made then they have been tested. 

I'm also not too sure if you did in fact test this, since the pyright issues show this will not work in some cases.

EmreTech:
`OptionConverter` already takes in an interaction argument though, in its `convert` method.

ooliver1:
> `OptionConverter` already takes in an interaction argument though, in its `convert` method.

As in, providing default converters.

PR #907, opened 2022-11-15
refactor(mentions): use None instead of _FakeBool

teaishealthy:
## Summary

This PR removes the `_FakeBool` sentinel value used in mentions.py over `None`

## Checklist

- [ ] This PR fixes an issue.


nextcord/mentions.py
```diff
@@ -114,20 +100,20 @@ def to_dict(self) -> AllowedMentionsPayload:
         parse = []
         data = {}
 
-        if self.everyone:
+        if self.everyone or self.everyone is None:
```
TAG-Epic, line 103: This seems over complicated. Why not check if its None and make it True?

teaishealthy: It's needed for .merge, which is used throughout the library.

EmreTech: I agree that this is over-complicated, but the proposed solution isn't a good idea. Here's a better approach.
```suggestion
        if self.everyone is not False:
```

teaishealthy: @EmreTech This does not work for checking `self.users`, as that can be List[Snowflake] too

EmreTech (approved):

ooliver1 (approved):

PR #910, opened 2022-11-17
typing: make MISSING type friendly

EmmmaTech:
## Summary

This PR makes the `MISSING` sentinel more type friendly. This is accomplished by migrating the `MISSING` sentinel definition to another file then making a custom type stub for that file where `_MissingSentinel` (the underlying class of `MISSING`) is defined to be an enum with one member. There is also a custom union made named `MissingOr` that is defined to be `Union[x, _MissingSentinel]`.

### Why do we need this?

For too long, discord.py has had their `MISSING` sentinel implicitly typed. The actual `MISSING` constant was defined with an `Any` type hint which tricks type checkers into thinking that this sentinel is compatible with everything (as that's how the `Any` type works). 

The problem with this is that there are no guarantees that a particular parameter in a function or an attribute in a class is `MISSING` or some other type. This is not good because it can lead to flaws in the code. Instead, explicitly typing the `MISSING` sentinel exposes these flaws for the developer to see and fix.

## Checklist

- [x] This PR fixes an issue.


ooliver1 (commented):
This might go in hand with documenting `MISSING` and `MissingOr` (`.. autodata`) as that is now shown as a possible type to give.

ooliver1 (commented):
Any progress? There seems to be some type issues that have risen from this.

ooliver1:
Closing this for now due to no progress

PR #917, opened 2022-11-23
refactor: move listeners to the main library

### LINKED ISSUE #892: Port event listeners from commands.Bot -> nextcord.Client
### Summary

Port event listeners from commands.Bot to nextcord.Client.

The core library

### The Problem

Currently with `nextcord.Client`, you can only have one callback per event. You can override event callbacks with your own behavior, but you can't have multiple listeners that don't affect the default event callback.

### The Ideal Solution

Port event listeners from `commands.Bot` but keep the event override functionality since that will break many bots that rely on this. This can be performed with a simple dictionary that maps event names -> lists of coroutines that are event listeners. Then, when a certain event is dispatched, we loop through all of the event listeners mapped to that event and run them all.

### The Current Solution

This functionality has already been implemented in `ext.commands`, so you can use that.

### Additional Context

_No response_

EmmmaTech:
## Summary

This PR moves listeners from `commands.Bot` to `nextcord.Client`. Closes #892.

## Checklist

- [ ] This PR fixes an issue.


Skelmis (approved):
Untested, lgtm

nextcord/ext/commands/bot.py
```diff
@@ -508,93 +508,13 @@ def after_invoke(self, coro: CFT) -> CFT:
 
     # listener registration
 
+    @nextcord.utils.copy_doc(nextcord.Client.add_listener)
```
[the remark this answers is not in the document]
EmreTech: `BotBase` doesn't inherit from `nextcord.Client` at all, and `add_listener` & `remove_listener` are both referenced later on in the `BotBase` code.
This is also why I kept in `dispatch` even though it just references `nextcord.Client.dispatch`.

nextcord/client.py
```diff
@@ -1298,6 +1303,94 @@ async def on_ready():
...
+        events from different places e.g. such as :func:`.on_ready`
+
+        The functions being listened to must be a :ref:`coroutine <coroutine>`.
+
+        Example
+        -------
+
+        .. code-block:: python3
+
+            @bot.listen()
```
spifory, line 1370:
```suggestion
            @client.listen()
```
Since this appears under client now, it should be `client.listen` for the sake of consistent naming, as defining `nextcord.Client` as `bot` (`bot = Client(...)`) is just weird

nextcord/client.py
```diff
@@ -1298,6 +1303,94 @@ async def on_ready():
...
+            async def on_message(message):
+                print('one')
+
+            # in some other file...
+
+            @bot.listen('on_message')
```
spifory, line 1376:
```suggestion
            @client.listen('on_message')
```
Same here

spifory (changes requested):

nextcord/ext/commands/bot.py
```diff
@@ -236,12 +237,11 @@ def __init__(
 
     # internal helpers
 
+    # kept in since BotBase isn't a direct subclass of client and thus doesn't know
+    # about inheriting this method
     def dispatch(self, event_name: str, *args: Any, **kwargs: Any) -> None:
         # super() will resolve to Client
         super().dispatch(event_name, *args, **kwargs)  # type: ignore
-        ev = "on_" + event_name
```
ooliver1, line 242: This is not included in the `Client` implementation. This is important as otherwise `extra_events` is holding `on_x` and `event_name` is just `x`.
Found by the following code not printing ready:
```py
import os

from dotenv import load_dotenv

from nextcord import Client, Intents, Message

load_dotenv()


client = Client(intents=Intents(guilds=True, guild_messages=True, message_content=True))


@client.listen()
async def on_ready():
    print("Bot is ready")


async def on_message(message: Message):
    if message.content == "ping":
        await message.channel.send("pong")
        client.remove_listener(on_message)


client.add_listener(on_message)
client.run(os.environ["TOKEN"])
```

EmreTech: This problem was actually with the line below the one you commented on, as I forgot to fetch with the modified listener name (so `"on_" + event`). Fixed in 1e376a6.

nextcord/client.py
```diff
@@ -1292,6 +1297,92 @@ async def on_ready():
         _log.debug("%s has successfully been registered as an event", coro.__name__)
         return coro
 
+    def add_listener(self, func: CoroFunc, name: str = MISSING) -> None:
+        """The non decorator alternative to :meth:`.listen`.
```
ooliver1, line 1301: I'm not sure as to if these should have versionadded, since they have been added to `Client`.

ooliver1 (commented):
Works fine. Test code:
```python
import os

from dotenv import load_dotenv

from nextcord import Client, Intents, Message

load_dotenv()


client = Client(intents=Intents(guilds=True, guild_messages=True, message_content=True))


@client.listen()
async def on_ready():
    print("Bot is ready")


async def on_message(message: Message):
    if message.content == "ping":
        await message.channel.send("pong")
        client.remove_listener(on_message)


client.add_listener(on_message)
client.run(os.environ["TOKEN"])
```

nextcord/client.py
EmreTech: e2724047

Skelmis:
@ooliver1 just waiting on your re-review now things are resolved

ooliver1 (approved):

PR #918, opened 2022-11-24
Created basic voice example with slash commands

MichalRajzer:
## Summary

This adds a slash command basic voice example. 

This was needed a few days ago when I started out with nextcord, so I thought maybe some people would benefit from this kind of example code.

## Checklist

  - [x] I have updated the documentation to reflect the changes. (Kind of, there is no documentation for the examples)
- [ ] This PR fixes an issue.

### DIFF
--- a/examples/slash_voice.py
+++ b/examples/slash_voice.py
@@ -78,11 +78,9 @@ class Music(commands.Cog):
                 await self.join(interaction, channel=interaction.user.voice.channel)
             else:
                 await interaction.send("You are not connected to a voice channel.")
-                raise commands.CommandError(
-                    "Author not connected to a voice channel.")
+                raise commands.CommandError("Author not connected to a voice channel.")
         source = nextcord.PCMVolumeTransformer(nextcord.FFmpegPCMAudio(query))
-        self.voiceClient.play(source, after=lambda e: print(
-            f"Player error: {e}") if e else None)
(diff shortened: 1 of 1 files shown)


ooliver1 (changes requested):
I have a few issues/questions
- Is this not just <https://github.com/nextcord/nextcord/blob/master/examples/basic_voice.py> but merged with a slash command? Examples do not need to list every possibility with everything.
- This does not follow [PEP8](https://peps.python.org/pep-0008/) naming schemes.
- Is it not possible to use `interaction.guild.voice_client` like the `basic_voice.py` example?

Soheab:
I don't see why this almost got a rewrite compared to the prefix voice example since the only things that needed changing and I can see are the following:
- `ctx` -> `interaction`
- `ctx.voice_client...` -> `interaction.guild.voice_client...`
- `ctx.author` -> `interaction.user`
- `async with ctx.typing():` -> `await interaction.response.defer()`

And it's bad practise to call nextcord commands manually, consider moving the content of the function to a free standing one and calling that instead but that's not needed here as said above.

+ everything from [Ooliver's review](https://github.com/nextcord/nextcord/pull/918#pullrequestreview-1193694488)

ooliver1:
Reason for close: example is unnecessary, examples don't need to be every permutation of other examples.

PR #894, opened 2022-11-08
typing: resolve issues with pyright 1.1.279

ooliver1:
## Summary

This PR fixes the type issues for `1.1.279` so it can be bumped for #872

## Checklist

- [ ] If code changes were made then they have been tested.
  - [ ] I have updated the documentation to reflect the changes.
  - [x] I have run `task pyright` and fixed the relevant issues.
- [x] This PR fixes an issue.
- [ ] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

PR #926, opened 2022-11-30
fix(commands): stop setting private attribute directly

EmreTech:
## Summary

This PR fixes a problem with #894 where an attribute in `Command` was changed to be set directly to the underlying private attribute which missed the property callback which sets another crucial attribute.

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/ext/commands/core.py
+++ b/nextcord/ext/commands/core.py
@@ -433,19 +433,17 @@ class Command(_BaseCommand, Generic[CogT, P, T]):
         Callable[Concatenate[CogT, ContextT, P], Coro[T]],
         Callable[Concatenate[ContextT, P], Coro[T]],
     ]:
-        # May require adding ContextT to the Generic of Command
-        return self._callback  # type: ignore
+        return self._callback
 
     @callback.setter
     def callback(
         self,
         function: Union[
-            Callable[Concatenate[CogT, Context, P], Coro[T]],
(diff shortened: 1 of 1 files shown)


nextcord/ext/commands/core.py
```diff
@@ -439,12 +439,13 @@ def callback(
     @callback.setter
     def callback(
         self,
-        function: Union[
+        function: Union[t
```
ooliver1, line 442: What is the lowercase `t` for?

EmreTech: Typo probably.

nextcord/ext/commands/core.py
```diff
@@ -320,7 +320,8 @@ def __init__(
             raise TypeError("Name of a command must be a string.")
         self.name: str = name
 
-        self._callback = func
+        # pyright complains that ContextT is incompatible with plain Context
```
ooliver1, line 323: Becuase it isn't. Consider fixing the issue or rewrite this comment to not seem so opposed to pyright.

ooliver1 (changes requested):

nextcord/ext/commands/core.py
EmreTech: I don't know how to fix this problem, so I'll rephrase the comment.

ArjunSharda (approved):

ooliver1 (approved):

PR #950, opened 2022-12-13
refactor!: rewrite async iterators

### LINKED ISSUE #122: Refactor async iterators to actually be async iterators and remove a lot of legacy jank.
### Links

https://github.com/Rapptz/discord.py/projects/3#card-59040298
https://discordpy.readthedocs.io/en/latest/migrating.html#moving-away-from-custom-asynciterator (more context on what exactly will be changed)

EmmmaTech:
## Summary

This PR rewrites async iterators to be contained inside of a function instead of a class to get rid of bloat.

Since this PR is a huge breaking change and removes the ability to use methods like `_AsyncIterator.flatten`, the following utility methods are modern replacements:

- `anext` has been built into python since 3.10 (next python minimum version for nextcord will be 3.12)
- `utils.as_chunks` as a replacement for `_AsyncIterator.chunk` (added in v2.0)

Closes #122.

## Checklist

- [ ] This PR fixes an issue.


ooliver1:
Alias what to what? This is overall a breaking change so won't be in 2.x.x, so breaking changes in a...breaking change is fine.

nextcord/abc.py
```diff
@@ -1585,8 +1586,8 @@ def history(
         after: Optional[SnowflakeTime] = None,
         around: Optional[SnowflakeTime] = None,
         oldest_first: Optional[bool] = None,
-    ) -> HistoryIterator:
-        """Returns an :class:`~nextcord.AsyncIterator` that enables receiving the destination's message history.
+    ) -> AsyncIterator[Message]:
+        """Returns an async iterator that enables receiving the destination's message history.
```
ooliver1, line 1590: It may make sense to refer to python's [definition](https://docs.python.org/3/glossary.html#term-asynchronous-iterator) of an async iterator, be it an inline link or a replacement (like `|coro|`, defined in `docs/conf.py`).

EmreTech: 0f20203

ooliver1 (commented):
Seems good to me, once conflicts are resolved I can test this.

ooliver1 (approved):
Much less code now, all simpler.
Except for the comment and the Discord message about old docs on `AsyncIterator` in `api.rst`, looks good. Untested.

PR #957, opened 2022-12-24
feat: add ability to format strings like text channel names

squeakerdev:
## Summary

Adds `format_as_channel_name`, a function to convert text to a text-channel compatible name, same as Discord does in the background when creating channels via the API. 

Example: "Joey's Cool Channel" -> "joeys-cool-channel"

This is my first PR, I've looked at the contributing guidelines but I'm unsure about what to do regarding docs. Please could someone take a look? Thanks! 😅 

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/utils.py
+++ b/nextcord/utils.py
@@ -1131,6 +1131,30 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None) 
     return f"<t:{int(dt.timestamp())}:{style}>"
 
 
+def format_as_channel_name(s: str, include_hashtag: bool = False) -> str:
+    """Strip a string of whitespace & special characters, like Discord does when creating a text channel.
+
+    Parameters
+    ----------
+    s: :class:`str`
+        The string to format.
+    include_hashtag: :class:`bool`
+        Whether or not to include the leading hashtag.
+
+    Returns
+    -------
+    :class:`str`
+        The formatted string.
+    """
+
+    if include_hashtag:
+        return "#" + re.sub(_TEXT_CHANNEL_REGEX, '-', s).strip().lower()
+    else:
+        return re.sub(_TEXT_CHANNEL_REGEX, '-', s).strip().lower()
+
+
+_TEXT_CHANNEL_REGEX = r"[^\w\s]+"
+
 _FUNCTION_DESCRIPTION_REGEX = re.compile(r"\A(?:.|\n)+?(?=\Z|\r?\n\r?\n)", re.MULTILINE)
 
 _ARG_NAME_SUBREGEX = r"(?:\\?\*)*(?P<name>[^\s:\-]+)"
--- a/nextcord/utils.py
+++ b/nextcord/utils.py
@@ -1148,9 +1148,9 @@ def format_as_channel_name(s: str, include_hashtag: bool = False) -> str:
     """
 
     if include_hashtag:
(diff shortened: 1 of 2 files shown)


nextcord/utils.py
```diff
@@ -1131,6 +1131,30 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None)
     return f"<t:{int(dt.timestamp())}:{style}>"
 
 
+def format_as_channel_name(s: str, include_hashtag: bool = False) -> str:
```
EmreTech, line 1134:
```suggestion
def format_as_channel_name(string: str, *, include_hashtag: bool = False) -> str:
```
Just to be more descriptive.

nextcord/utils.py
```diff
@@ -1131,6 +1131,30 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None)
...
+    """Strip a string of whitespace & special characters, like Discord does when creating a text channel.
+
+    Parameters
+    ----------
+    s: :class:`str`
```
EmreTech, line 1139:
```suggestion
    string: :class:`str`
```

EmreTech:
@passivity I added this in the docs for you in dca3857.

nextcord/utils.py
EmreTech: Why was this resolved without any conversation?

Skelmis: +1, I don't see a reason for these to be resolved without any conversation or explanation

nextcord/utils.py
```diff
@@ -1131,6 +1131,30 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None)
...
+    Returns
+    -------
+    :class:`str`
+        The formatted string.
+    """
+
+    if include_hashtag:
+        return "#" + re.sub(_TEXT_CHANNEL_REGEX, "-", s).strip().lower()
+    else:
+        return re.sub(_TEXT_CHANNEL_REGEX, "-", s).strip().lower()
```
Skelmis, line 1153: I'm not adamant on this, but its a suggestion if you'd like it to be more inline
```suggestion
    new_string = re.sub(_TEXT_CHANNEL_REGEX, "-", s).strip().lower()
    return f"#{new_string}" if include_hashtag else new_string
```

nextcord/utils.py
```diff
@@ -1131,6 +1131,30 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None)
...
+
+
+_TEXT_CHANNEL_REGEX = r"[^\w\s]+"
```
ooliver1, line 1156: Emojis sometimes work in channel names, are you sure this is the pattern that should be used?

nextcord/utils.py
ooliver1: Just an extension to this, one character variable names are awful for readability, and the key word argument being forced is because `format_as_channel_name("name", True)` has no semantic meaning for `True`, but `format_as_channel_name("name", include_hashtag=True)` does.

ooliver1 (changes requested):

Skelmis:
Hi @squeakerdev, just following up on this. How are you feeling on the PR regarding the review changes?

ooliver1:
I'm closing this for now as there has been no activity for a year. I am not sure about the need for this either, as Discord formats them API-side themselves, so there is no error from the API to mitigate.

PR #965, opened 2023-01-06
refactor: use audit log `after` query param

ooliver1:
## Summary

Since the after parameter in audit log retrieval is documented, this can be used over filtering, reversing and fetching tons of logs.
<https://github.com/discord/discord-api-docs/commit/38f01b07531b672880e434692d215233e45eb0db>
[Test Code](https://paste.nextcord.dev/?id=1673044721454197)

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/iterators.py
+++ b/nextcord/iterators.py
@@ -282,11 +282,8 @@ class HistoryIterator(_AsyncIterator["Message"]):
         if isinstance(around, datetime.datetime):
             around = Object(id=time_snowflake(around))
 
-        self.reverse: bool
-        if oldest_first is None:
-            self.reverse = after is not None
-        else:
-            self.reverse = oldest_first
+        # Maintain compatibility with `None` being passed in for `oldest_first`
+        self.reverse: bool = bool(oldest_first)
 
         self.messageable: Messageable = messageable
         self.limit: Optional[int] = limit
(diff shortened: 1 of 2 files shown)


nextcord/iterators.py
```diff
@@ -521,7 +521,7 @@ def __init__(
 
         self.reverse: bool
         if oldest_first is None:
-            self.reverse = after is not None
+            self.reverse = False
         else:
             self.reverse = oldest_first
```
EmreTech, line 526: Can't you just set `reverse` to `bool(oldest_first)` instead? `None` would be evaluated to `False`.
I would make a suggestion for this, but GitHub doesn't allow me to make suggestions with deleted lines.

nextcord/iterators.py
```diff
@@ -282,11 +282,8 @@ def __init__(
         if isinstance(around, datetime.datetime):
             around = Object(id=time_snowflake(around))
 
-        self.reverse: bool
-        if oldest_first is None:
-            self.reverse = after is not None
-        else:
-            self.reverse = oldest_first
+        # Maintain compatibility with `None` being passed in for `oldest_first`
+        self.reverse: bool = bool(oldest_first)
```
EmreTech, line 286: This is the history iterator, not the audit log iterator.

EmreTech (approved):

PR #970, opened 2023-01-09
refactor: nextcord.Interaction subclassing

### LINKED ISSUE #952: Creating Subclasses of Common Class Objects
### Summary

There should be subclasses for most classes. Example: `Interaction` is the base class, `SlashCommandInteraction` is the class that should be used.

The core library

### The Problem

I feel that there should be subclasses for most of the classes in the library since it would help provide the relevant attributes for certain situations. An example would be how `<Interaction>.message` should not work for modals (I think).

### The Ideal Solution

Creating subclasses for commonly used classes in the library. Example:
```py
class Interaction:
	...

class ModalInteraction(Interaction):
	...

class SlashCommandInteraction(Interaction):
	...
```

### The Current Solution

_No response_

### Additional Context

Disnake _does_ do something like this, for reference.

Gangsta-Muffin:
## Summary
PR for #952 

Create Subclasses for `nextcord.Interaction`

Subclasses will include:
-------------------------

`ApplicationCommandInteraction`
Interaction object for application commands

`ApplicationAutocompleteInteraction`
Interaction object for slash command autocomplete

`ViewInteraction`
Interaction object for message components

`ModalSubmitInteraction`
Interaction object for Modal submits

Following object names will/may still change:
- `ApplicationAutocompleteInteraction` 
- `ViewInteraction` to `MessageComponentInteraction`

## Checklist (currently incomplete)

- [x] SlashApplicationInteraction (Interaction subclass)
- [x] ApplicationAutocompleteInteraction (Interaction subclass)
- [x] ViewInteraction (Interaction subclass)
- [x] ModalSubmitInteraction (Interaction subclass)
- [x] InteractionResponse subclasses
- [x] Interaction subclasses in `nextcord.types.interactions`
-------------

- [x] Update external Modules to use/identify correct Object
- [x] Add new subclasses to `__all__` Set in interactions.py
- [x] Update docs
- [x] Update additional stuff (for example: `__slots__`, type hints, bugs, ...)

-------------
- [ ] Final test (for bugs) - fix if necessary


nextcord/interactions.py
```diff
@@ -559,6 +539,485 @@ async def send(
...
+    def _from_data(self, data: InteractionPayload):
+        super()._from_data(data=data)
+
+        self.app_command_name: str = self.data["name"]  # type: ignore # self.data should be present here
+        self.app_command_id: int = self.data["id"]  # type: ignore # self.data should be present here
+
+        try:
+            self.options: list = self._get_application_options(self.data["options"])  # type: ignore # self.data should be present here
+        except KeyError:
+            self.options: list = []
```
EmreTech, line 643: Fully define this type hint.

nextcord/interactions.py
```diff
@@ -559,6 +539,485 @@ async def send(
...
+        The token to continue the interaction. These are valid
+        for 15 minutes.
+    data: :class:`dict`
+        The raw application command specific interaction data.
+    attached: :class:`InteractionAttached`
+        The attached data of the interaction. This is used to store any data you may need inside the interaction for convenience. This data will stay on the interaction, even after a :meth:`Interaction.application_command_before_invoke`.
+    application_command: Union[SlashApplicationSubcommand, BaseApplicationCommand]
+        The application command that triggered the interaction
+    app_command_name: :class:`str`
+        The name of the application command that triggered the interaction
```
EmreTech, line 593:
```suggestion
        The name of the application command that triggered the interaction.
```

nextcord/interactions.py
```diff
@@ -559,6 +539,485 @@ async def send(
...
+    guild_locale: Optional[:class:`str`]
+        The guilds preferred locale, if invoked in a guild.
+    application_id: :class:`int`
+        The application ID that the interaction was for.
+    user: Optional[Union[:class:`User`, :class:`Member`]]
+        The user or member that sent the interaction.
+    token: :class:`str`
+        The token to continue the interaction. These are valid
+        for 15 minutes.
+    data: :class:`dict`
```
EmreTech, line 586: It would be nice if this `data` attribute was uniquely typed for each interaction subclass to use a corresponding typed dict.

nextcord/interactions.py
```diff
@@ -559,6 +539,485 @@ async def send(
             allowed_mentions=allowed_mentions,
         )
 
+
+class ApplicationCommandInteraction(Interaction):
+    """Represents the base class for all Discord interactions.
```
EmreTech, line 544:
```suggestion
    """Represents an application command interaction.
```

nextcord/interactions.py
```diff
@@ -559,6 +539,485 @@ async def send(
...
+        options_collection: list = []
+
+        for option in data:
+            options_collection.append(SlashOptionData(option))
+
+        return options_collection
+
+
+class ApplicationAutocompleteInteraction(Interaction):
+    """Represents the base class for all Discord interactions.
```
EmreTech, line 660:
```suggestion
    """Represents an autocomplete application command interaction.
```

nextcord/interactions.py
```diff
@@ -559,6 +539,485 @@ async def send(
...
+    def _get_application_options(self, data):
+        options_collection: list = []
+
+        for option in data:
+            options_collection.append(SlashOptionData(option))
+
+        return options_collection
+
+
+class ApplicationAutocompleteInteraction(Interaction):
```
EmreTech, line 659: If this class is going to copy a lot of logic from `ApplicationCommandInteraction`, shouldn't this class be a subclass of `ApplicationCommandInteraction`?

nextcord/interactions.py
```diff
@@ -559,6 +539,485 @@ async def send(
...
+        The user or member that sent the interaction.
+    token: :class:`str`
+        The token to continue the interaction. These are valid
+        for 15 minutes.
+    data: :class:`dict`
+        The raw View specific interaction data.
+    attached: :class:`InteractionAttached`
+        The attached data of the interaction. This is used to store any data you may need inside the interaction for convenience. This data will stay on the interaction, even after a :meth:`Interaction.application_command_before_invoke`.
+    message: :class:`Message`
+        The message the component is attached to
```
EmreTech, line 830:
```suggestion
        The message the component is attached to.
```

nextcord/interactions.py
```diff
@@ -559,6 +539,485 @@ async def send(
...
+    component_id: :class:`int`
+        The ID of the component that triggered the Interaction
```
EmreTech, line 832:
```suggestion
        The ID of the component that triggered the interaction.
```

nextcord/interactions.py
```diff
@@ -559,6 +539,485 @@ async def send(
...
+    value: :class:`str`
+        The value of the component that trgigered the interaction
```
EmreTech, line 834:
```suggestion
        The value of the component that triggered the interaction.
```

nextcord/interactions.py
```diff
@@ -559,6 +539,485 @@ async def send(
...
+        "_app_permissions",
+        "_state",
+        "_session",
+        "_original_message",
+        "_cs_response",
+        "_cs_followup",
+        "_cs_channel",
+    )
+
+    def __init__(self, *, data: InteractionPayload, state: ConnectionState):
```
EmreTech, line 627: It would also be nice if the `data` parameter was more specifically typed. Right now, it allows for any interaction payload, but in this case we want to limit it to application command interaction payloads.

nextcord/interactions.py
```diff
@@ -559,6 +539,485 @@ async def send(
...
+        super().__init__(data=data, state=state)
+
+        self.application_command: Optional[
+            Union[SlashApplicationSubcommand, BaseApplicationCommand]
+        ] = None
+
+    def _from_data(self, data: InteractionPayload):
```
EmreTech, line 634: Same here.

nextcord/interactions.py
```diff
@@ -559,6 +539,485 @@ async def send(
...
+    token: :class:`str`
+        The token to continue the interaction. These are valid
+        for 15 minutes.
+    data: :class:`dict`
+        The raw View specific interaction data.
+    attached: :class:`InteractionAttached`
+        The attached data of the interaction. This is used to store any data you may need inside the interaction for convenience. This data will stay on the interaction, even after a :meth:`Interaction.application_command_before_invoke`.
+    message: :class:`Message`
+        The message the component is attached to
+    component_id: :class:`int`
```
EmreTech, line 831: It looks like the component id attribute is really the custom id. Custom ids for components are typed to be strings, not integers.

EmmmaTech (changes requested):
Two more things that I decided to leave out of individual review comments:
- The rest of the library isn't typed to use these new subclasses.
- It would be a lot cleaner imo if the interaction subclasses were split into multiple files under an `interactions` folder. This also includes the base interaction class.

FYI, some of the review comments I specifically made on `ApplicationCommandInteraction` apply to the rest of the subclasses, I didn't want to copy all of those review comments over and over again.

MaskDuck (changes requested):
also i'm unsure whether should we keep documenting Interaction -- as it doesn't have any real-world usage now.

i think it's a bit breaking but fine, acceptable

nextcord/interactions.py
ooliver1: Issue is, `nextcord.types` in general isn't documented, would it make more sense to simply link to the discord docs?

EmreTech: I suppose.

MaskDuck (changes requested):
would need to add backward compatibilities or fix pyright error
i actually prefer add backward compatibilities bcuz it would not take much time

but you also need to re-typehint the whole lib where just use interactions, thx =))

EmmmaTech (commented):
I would rename the submodules of `nextcord.interactions` to remove the `_interaction(s)` suffix because the name of the parent module already implies that the submodules contain interaction subclasses.

A good example of where this was done right is `nextcord.ui.select`.

nextcord/interactions.py
```diff
@@ -594,7 +799,120 @@ async def edit(self, *args, **kwargs) -> Optional[Message]:
...
+
+    .. container:: operations
+
+        .. describe:: x == y
+
+            Checks if two interactions are equal.
+
+        .. describe:: x != y
+
+            Checks if two interactions are not equal.
```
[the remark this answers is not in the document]
Skelmis: No, the current documented addition is consistent with the rest of the library

nextcord/interactions.py
```diff
@@ -594,7 +799,120 @@ async def edit(self, *args, **kwargs) -> Optional[Message]:
...
+class ModalSubmitInteraction(Interaction):
+    """Represents the interaction for all modal submits.
+
+    This interaction gets triggered by :class:`nextcord.ui.Modal`
+
+    .. container:: operations
+
+        .. describe:: x == y
+
+            Checks if two interactions are equal.
```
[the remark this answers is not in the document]
Skelmis: No, the current documented addition is consistent with the rest of the library

nextcord/interactions.py
```diff
@@ -1301,3 +1619,23 @@ class InteractionMessage(_InteractionMessageMixin, Message):
...
 
     pass
+
+
+# Lives here - can't go into application_commands.py due to circular import
+class SlashOptionData:
+    """Class description"""
+
+    def __init__(self, data) -> None:
+        # Still missing `options?` Data from (view Discord Docs)
```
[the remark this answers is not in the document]
Skelmis: What do you mean "add the property", I am unsure on what is required to resolve this review comment

nextcord/interactions.py
Skelmis: As an side, possibly we should discuss introducing something like `discord-typings` if we want to go that route? @ooliver1

ooliver1: `discord-typings` too does not have documentation

nextcord/interactions/message_component.py
```diff
@@ -0,0 +1,109 @@
...
+
+if TYPE_CHECKING:
+    from ..state import ConnectionState
+    from ..types.interactions import Interaction as InteractionPayload
+
+
+class MessageComponentInteraction(Interaction):
+    """Represents the interaction for all messsage components.
+
+    This interaction is get's triggered by :class:`nextcord.ui.View`
```
EmreTech, line 19:
```suggestion
    This interaction gets triggered by a :class:`nextcord.ui.View`.
```

nextcord/interactions/message_component.py
```diff
@@ -0,0 +1,109 @@
...
+        "value",
+    )
+
+    def __init__(self, *, data: InteractionPayload, state: ConnectionState) -> None:
+        super().__init__(data=data, state=state)
+
+    def _from_data(self, data: InteractionPayload) -> None:
+        super()._from_data(data=data)
+
+        message = data["message"]  # type: ignore # should be present here - not sure why its causing issues
```
EmreTech, line 57: According to the type hint for `data` in this function (which is `nextcord.types.Interaction`), the `message` key might not be available because `message` in the `Interaction` typed dict is not required.

nextcord/interactions/message_component.py
```diff
@@ -0,0 +1,109 @@
...
+        self.message = self._state._get_message(
+            int(message["id"])
+        ) or Message(  # ---- check for and create correct message object ----
+            state=self._state, channel=self.channel, data=message  # type: ignore
+        )
+        self.component_id = self.data["custom_id"]  # type: ignore # self.data should be present here
```
EmreTech, line 63: `self.data` is typed as optional, so according to Pyright it could not be present.

nextcord/state.py
```diff
@@ -1377,11 +1377,13 @@ def parse_interaction_create(self, data) -> None:
         if data["type"] == 3:  # interaction component
             custom_id = interaction.data["custom_id"]  # type: ignore
             component_type = interaction.data["component_type"]  # type: ignore
-            self._view_store.dispatch(component_type, custom_id, interaction)
+            self._view_store.dispatch(component_type, custom_id, interaction)  # type: ignore
+            # Some Interaction types incompatible with this method -> accounted for via if-statement
```
EmreTech, line 1381: Not in Pyright's eyes. Checking if a certain value in the raw interaction data used to make an `Interaction` is another value doesn't mean that the created object is magically that type. You have to add more to the if statement to justify this.

nextcord/ui/item.py
```diff
@@ -4,7 +4,7 @@
 
 from typing import TYPE_CHECKING, Any, Callable, Coroutine, Generic, Optional, Tuple, TypeVar
 
-from ..interactions import ClientT, Interaction
+from ..interactions.base import ClientT, Interaction
```
EmreTech, line 7: Shouldn't this be using `MessageComponentInteraction`?

nextcord/interactions/modal_submit.py
```diff
@@ -0,0 +1,106 @@
...
+
+if TYPE_CHECKING:
+    from ..state import ConnectionState
+    from ..types.interactions import Interaction as InteractionPayload
+
+
+class ModalSubmitInteraction(Interaction):
+    """Represents the interaction for all modal submits.
+
+    This interaction gets triggered by :class:`nextcord.ui.Modal`
```
EmreTech, line 19:
```suggestion
    This interaction is triggered by a :class:`nextcord.ui.Modal`.
```

nextcord/interactions/application.py
```diff
@@ -0,0 +1,177 @@
...
+__all__ = ("ApplicationCommandInteraction", "ApplicationAutocompleteInteraction", "SlashOptionData")
+
+if TYPE_CHECKING:
+    from ..application_command import BaseApplicationCommand, SlashApplicationSubcommand
+    from ..state import ConnectionState
+    from ..types.interactions import Interaction as InteractionPayload
+
+
+# class is here due to circular import issues
+class SlashOptionData:
```
EmreTech, line 16: Why is there documentation for this class? If it's copy-pasted here for circular import issues, then it would be internally used.

nextcord/interactions/application.py
```diff
@@ -0,0 +1,177 @@
...
+
+            if len(options) == 0:
+                # return empty list if no options exist
+                return []
+
+        # If we are here, then the user provided an input for some options
+        options_collection: List[SlashOptionData] = []
+
+        for option in options:
+            options_collection.append(SlashOptionData(option))
```
EmreTech, line 134: There's an easy way to get around the circular import issues if this is what `SlashOptionData` is being used for in the code. Just import whatever you need in the scope of the function. 
For example:

```python
def foo():
    from .bar import soap
    soap()
```

And for any time `SlashOptionData` is being used for type hints outside the scope of a function, import it under a `TYPE_CHECKING` block and enable `from __future__ import annotations`.

EmmmaTech (changes requested):
Additionally, every new submodule in the `nextcord.interactions` module is missing the license identifier comment (`# SPDX-License-Identifier: MIT`).

nextcord/interactions/application.py
Gangsta-Muffin: The point behind this comment was the fact that I was originally trying to add the class to `nextcord.application_command` since there is a class of similar nature in that file: `SlashOption` (line 1218 ff.). The `SlashOptionData` class is designed to contain SlashOption data provided by a user, while the `SlashOption` class is designed to take requirements for an application command option - requirements set by the developer.

nextcord/interactions/application.py
Gangsta-Muffin: I assume `foo()` represents the `_get_application_options()` method in this case? It is customary to have all imports at the top of a python file. Is it really a good idea to import within a function mid file? 

(Yes, this is why the class is in this file- was originally going to put it in `nextcord.application_command.py`. Since you asked about this in the other comment, and the discussion will continue here, I shall resolve that conversation)

nextcord/interactions/application.py
Gangsta-Muffin: conversation continueing in other conversation

nextcord/interactions/application.py
```diff
@@ -0,0 +1,818 @@
...
+
+    def _set_application_command(
+        self, app_cmd: Union[SlashApplicationSubcommand, BaseApplicationCommand]
+    ) -> None:
+        self.application_command = app_cmd
+
+    def _get_application_options(self, data):
+        options = data
+
+        if len(options) == 0:
```
teaishealthy, line 107: Empty containers are falsy
```suggestion
        if not options:
```

nextcord/interactions/message_component.py
```diff
@@ -447,191 +341,62 @@ async def edit_original_message(
...
-        if self.message is not None:
-            return await self.message.edit(*args, **kwargs)
-        raise InvalidArgument(
-            "Interaction.message is None, this method can only be used in "
-            "response to a component or modal submit interaction."
+
+        if self._original_message is not None:
+            return self._original_message
+
+        # TODO: fix later to not raise?
```
teaishealthy, line 371: ?

nextcord/interactions/message_component.py
```diff
@@ -1,102 +1,44 @@
...
+    from ..state import ConnectionState
+    from ..types.interactions import MessageComponentInteraction as MessageComponentPayload
+    from ..ui.modal import Modal
+    from ..ui.view import View
+    from .modal_submit import ModalSubmitInteraction
 
-    def __repr__(self) -> str:
-        return f"<InteractionAttached {super().__repr__()}>"
 
+class MessageComponentInteraction(Interaction):
```
teaishealthy, line 36: A lot of this code is duplicated across both classes in application.py. Move into a seperate super class or base.py

teaishealthy (changes requested):

teaishealthy:
also needs version docs thingies

teaishealthy:
🦅

nextcord/interactions/__init__.py
```diff
@@ -0,0 +1,15 @@
+"""
+nextcord.interactions
+~~~~~~~~~~~~~~~~~~
+
+Interactions collection for the Discord API
+
+:copyright: (c) 2015-2021 Rapptz
```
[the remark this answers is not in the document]
Gangsta-Muffin: > this file contains no code from Rapptz

What should I write there instead?

nextcord/interactions/application.py
```diff
@@ -0,0 +1,818 @@
...
+from .. import utils
+from ..embeds import Embed
+from ..enums import InteractionResponseType, InteractionType
+from ..errors import ClientException, HTTPException, InteractionResponded, InvalidArgument
+from ..file import File
+from ..flags import MessageFlags
+from ..message import Attachment
+from ..webhook.async_ import WebhookMessage, async_context, handle_message_parameters
+from .base import (
+    MISSING,
```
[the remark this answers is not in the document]
Gangsta-Muffin: that's what MISSING from .base is. should I do that anyways?

nextcord/interactions/application.py
```diff
@@ -0,0 +1,818 @@
...
+            self.options: List[SlashOptionData] = self._get_application_options(self.data["options"])  # type: ignore # Data should be defined here
+        except KeyError:
+            self.options: List[SlashOptionData] = []
+
+    def _set_application_command(
+        self, app_cmd: Union[SlashApplicationSubcommand, BaseApplicationCommand]
+    ) -> None:
+        self.application_command = app_cmd
+
+    def _get_application_options(self, data):
```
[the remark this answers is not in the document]
Gangsta-Muffin: I originally Had `ApplicationAutocompleteInteraction` inherit from `ApplicationCommandInteraction` as the Autocomplete interaction contains similar data as the app command interaction class. In a later commit, I realized I still had some methods I needed to transfer from the base Interaction class, methods that would be compatible for App commands, but not for autocompletes, so I changed the inherit for Autocomplete, and copy pasted the relevant methods Autocomplete was inheriting, and put it in the class.

I suppose a Payload class would be better, which both autocomplete and app command would use? Would reduce redundancy I suppose.

nextcord/interactions/application.py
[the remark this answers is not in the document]
teaishealthy: yes

nextcord/interactions/message_component.py
Gangsta-Muffin: This is a Todo inserted by someone else - was just copied over. I am guessing it still wants doing at some point

nextcord/interactions/modal_submit.py
```diff
@@ -0,0 +1,764 @@
...
+            token=self.token,
+            session=self._session,
+        )
+        state = _InteractionMessageState(self, self._state)
+        message = InteractionMessage(state=state, channel=channel, data=data)  # type: ignore
+        self._original_message = message
+        return message
+
+
+class ModalSubmitInteractionResponse(InteractionResponse):
```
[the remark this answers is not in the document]
Gangsta-Muffin: What code are you referencing specifically?

nextcord/interactions/application.py
```diff
@@ -0,0 +1,818 @@
...
+                view=view,
+                tts=tts,
+                ephemeral=ephemeral,
+                delete_after=delete_after,
+                allowed_mentions=allowed_mentions,
+                flags=flags,
+                suppress_embeds=suppress_embeds,
+            )
+        return await self.followup.send(
+            content=content,  # type: ignore
```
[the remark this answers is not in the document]
Gangsta-Muffin: --- from old code; Discussed in nextcord support server ---

nextcord/interactions/application.py
```diff
@@ -0,0 +1,818 @@
...
+        ------
+        HTTPException
+            Deferring the interaction failed.
+        InteractionResponded
+            This interaction has already been responded to before.
+        """
+        if self._responded:
+            raise InteractionResponded(self._parent)
+
+        defer_type: int = 0
```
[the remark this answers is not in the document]
Gangsta-Muffin: --- from old code; Discussed in nextcord support server ---

nextcord/interactions/application.py
```diff
@@ -0,0 +1,818 @@
...
+        ClientException
+            The channel for the message could not be resolved.
+
+        Returns
+        -------
+        InteractionMessage
+            The original interaction response message.
+        """
+
+        # TODO: fix later to not raise?
```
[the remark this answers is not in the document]
Gangsta-Muffin: --- from old code; Discussed in nextcord support server ---

nextcord/interactions/message_component.py
Gangsta-Muffin: --- from old code; Discussed in nextcord support server ---

nextcord/client.py
```diff
@@ -2869,7 +3001,38 @@ class be provided, it should be a subclass of :py:class:`Interaction`.
         .. note::
 
             This is synchronous due to how slash commands are implemented.
-        """
+
+        Warns
+        --------
+        ~FutureWarning
```
EmreTech, line 3007: I don't think it's worth documenting this warning, considering that other functions throughout the library do not either.

nextcord/client.py
```diff
@@ -2869,7 +3001,38 @@ class be provided, it should be a subclass of :py:class:`Interaction`.
...
+            Warns when the cls parameter is redefined to a custom class. This serves as a warning
+            in the event that defining this parameter was a mistake, and other parameters were to be defined instead.
+        """
+        # check if they are creating custom interaction subclasses
+        if cls != Interaction:
```
EmreTech, line 3012:
```suggestion
        if cls is not Interaction:
```
`!=` is usually used with literals.

nextcord/client.py
```diff
@@ -2869,7 +3001,38 @@ class be provided, it should be a subclass of :py:class:`Interaction`.
...
+            warnings.warn(
+                "The 'cls' parameter was given a custom Interaction object. "
```
EmreTech, line 3014:
```suggestion
                "The 'cls' parameter was given a custom Interaction class. "
```

nextcord/components.py
```diff
@@ -77,6 +77,10 @@ def __repr__(self) -> str:
     def _raw_construct(cls, **kwargs) -> Self:
         self = cls.__new__(cls)
         for slot in get_slots(cls):
+            if slot == "type" and getattr(self, slot, None) is not None:
+                # nextcord/issues/1103
+                continue
```
EmreTech, line 82: Wasn't this fixed by #1104?

EmmmaTech (changes requested):

nextcord/types/interactions.py
```diff
@@ -202,16 +202,46 @@ class Interaction(TypedDict):
     type: InteractionType
     token: str
     version: int
-    data: NotRequired[InteractionData]
+    # data: NotRequired[InteractionData]
```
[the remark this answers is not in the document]
Gangsta-Muffin: Because the subclasses recieved their own attribute for data. I did this partially due to typehinting issues, but also because the PingInteraction class (not currently in nextcord, but exists in the dev docs) does not have a data key. (But I mainly did it due to typehint issues - since the `InteractionData` type is a Union of all InteractionData types, and in the various subclasses, pyright would complain in various places, that various code was okay provided if the typehint was the correct one, but would have issues if the type was that used for the other subclasses)

nextcord/interactions/application.py
```diff
@@ -0,0 +1,808 @@
...
+        super().__init__(data=data, state=state)
+
+        self.application_command: Optional[
+            Union[SlashApplicationSubcommand, BaseApplicationCommand]
+        ] = None
+
+    def _from_data(self, data: ApplicationCommandPayload) -> None:
+        super()._from_data(data=data)
+
+        self.data: InteractionData = data.get("data")
```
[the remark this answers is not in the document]
Gangsta-Muffin: They subclasses do in fact use the correct types. The imports follow similar patter to the following: 
`from ..types.interactions import ApplicationCommandInteractionData as InteractionData`

nextcord/interactions/application.py
```diff
@@ -0,0 +1,808 @@
...
+
+        options = self.data.get("options")
+        self.options: List[SlashOptionData] = (
+            self._get_application_options(options) if options else []
+        )
+
+    def _set_application_command(
+        self, app_cmd: Union[SlashApplicationSubcommand, BaseApplicationCommand]
+    ) -> None:
+        self.application_command = app_cmd
```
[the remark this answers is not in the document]
Gangsta-Muffin: This is part of old code not written by me. From searching through the library, by the looks of things, it gets used in `CallbackMixing.invoke_callback_with_hooks()`. Im guessing it serves a purpose, but since its not my code, I've not looked into it

nextcord/interactions/application.py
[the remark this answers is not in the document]
EmreTech: Ah, I missed that part. I don't see why it should be aliased to `InteractionData`, it makes the code a little confusing to read through.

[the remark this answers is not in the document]
Gangsta-Muffin: I aliased them for the following reasons:
- It was aliased in the original Interaction file
- For the sake of reducing the length of the typehints since they are quite big
- For the sake of simplifying things a bit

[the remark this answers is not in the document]
EmreTech: 1. No it wasn't? I just looked at the source code, and only the `Interaction` typed dict was aliased to not conflict with the actual `Interaction` class defined in the file.
2. That's not a strong argument to use this alias. To me, the whole point of type hints is not only for static type checkers, but also for people reading the code to understand what a specific variable/attribute/parameter is. Shortening it like how you did it here makes it less obvious what specifically the type of the `data` attribute is. Plus, we already have linters to take care of long lines.
3. How would this alias simplify things? I don't understand.

nextcord/types/interactions.py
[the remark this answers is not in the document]
EmreTech: Yeah, I forgot about that while reviewing this PR.

nextcord/interactions/application.py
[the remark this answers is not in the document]
Gangsta-Muffin: Regarding point 3: My thought was mainly to Simply the class name, since `ApplicationCommandInteractionData` was very long and a bit overwhelming, and I thought the alias simplified things and in result, allowed for better understanding of the code - Albeit, based on your counter arguments, I can see the flaws and reasons for why this isn't a good idea; will be changing the code

nextcord/application_command.py
```diff
@@ -1236,6 +1248,46 @@ def error_name(self) -> str:
...
+        The type of the application command option. This indicates what value the option accepts as an input
+    name: :class:`str`
+        The name of this option
+    focused: :class:`bool`
+        Whether the user is currently creating an input for this option. This is useful for :meth:`on_autocomplete`
+    """
+
+    __slots__: Tuple[str, ...] = ("data", "value", "type", "name", "focused")
+
+    def __init__(self, data) -> None:
```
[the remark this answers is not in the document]
Gangsta-Muffin: This type doesn't exist - Are you suggesting I should make a new type for this?

nextcord/components.py
Gangsta-Muffin: Possibly? I haven't been keeping track on more recent PRs. Please let me know what I should do with this code, provided changes are in order :)

nextcord/application_command.py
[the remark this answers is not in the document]
EmreTech: https://github.com/nextcord/nextcord/blob/5d2eff1810edd76359b66f37f731537a520c3d2c/nextcord/types/interactions.py#L128-L135

It does exist, and it's already imported in this module.

nextcord/components.py
EmreTech: All that PR did when it was merged was remove `"type"` from the `__slots__` of the `Component` class. As long as your fork is up to date with the upstream and you remove this code I commented on, all should be good.

nextcord/application_command.py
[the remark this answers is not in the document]
Gangsta-Muffin: Must have missed that somehow. Either way, as part of that list, is the class  `_ApplicationCommandInteractionDataOptionSubcommand`, which the SlashOptionData class doesn't use. (It may possibly not use Snowflake either but I can't say this for sure). Would it be worth making a new dedicated type for this? or is there some typehinting approach I can take, to reassure pyright, that this class does, indeed, not get used?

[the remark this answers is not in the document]
EmreTech: Well, it does get used to tell Pyright what keys and types of values a specific dictionary has. You could make a new union without the subcommand one, but I'm not sure what to call it.

nextcord/interactions/application.py
```diff
@@ -0,0 +1,808 @@
...
+        super()._from_data(data=data)
+
+        self.data: InteractionData = data.get("data")
+        self.app_command_name: str = self.data["name"]
+        self.app_command_id: Snowflake = self.data["id"]
+
+        options = self.data.get("options")
+        self.options: List[SlashOptionData] = (
+            self._get_application_options(options) if options else []
+        )
```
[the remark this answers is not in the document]
teaishealthy: preview is broken for some reason, but what Emre said.
```suggestion
        options = self.data.get("options", [])
        self.options: List[SlashOptionData] = (
            self._get_application_options(options)
        )
```

nextcord/interactions/application.py
teaishealthy: Forgot to apply this

nextcord/interactions/application.py
```diff
@@ -0,0 +1,808 @@
...
+
+            if not options:
+                # return empty list if no options exist
+                return []
+
+        from ..application_command import (  # Importing here due to circular import issues
+            SlashOptionData,
+        )
+
+        return [SlashOptionData(option) for option in options]
```
teaishealthy, line 128: pyright is not happy with this because you are changing the type of an already assigned variable.

nextcord/interactions/application.py
```diff
@@ -0,0 +1,808 @@
...
+            type=InteractionResponseType.modal.value,
+            data=modal.to_dict(),
+        )
+
+        self._responded = True
+
+        self._parent._state.store_modal(modal, self._parent.user.id)  # type: ignore
+
+
+class ApplicationAutocompleteInteraction(Interaction):
```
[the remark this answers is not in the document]
teaishealthy: @Gangsta-Muffin ?

teaishealthy (changes requested):

nextcord/client.py
```diff
@@ -2869,7 +3001,38 @@ class be provided, it should be a subclass of :py:class:`Interaction`.
...
-        """
+
+        Warns
+        --------
+        ~FutureWarning
+            Warns when the cls parameter is redefined to a custom class. This serves as a warning
+            in the event that defining this parameter was a mistake, and other parameters were to be defined instead.
+        """
+        # check if they are creating custom interaction subclasses
+        if cls is not Interaction:
```
ooliver1, line 3012: This warning seems a bit weird, the signature for `get_interaction` seems to imply you should pass a custom interaction when passing other custom interaction classes. The better check would be to look at the other classes and if none are custom, but that just seems a bit excessive IMO.
```python
class Client(nextcord.Client):
    def get_interaction(self, data: InteractionPayload):
        return super().get_interaction(
            data,
            application_autocomplete_interaction=CustomAutocompleteInteraction,
            application_interaction=CustomAppCmdInteraction,
            message_component_interaction=CustomComponentInteraction,
            modal_submit_interaction=CustomModalSubmitInteraction,
            cls=CustomInteraction,
        )
```

nextcord/ui/item.py
```diff
@@ -2,9 +2,11 @@
 
 from __future__ import annotations
 
-from typing import TYPE_CHECKING, Any, Callable, Coroutine, Generic, Optional, Tuple, TypeVar
+from typing import TYPE_CHECKING, Any, Callable, Coroutine, Generic, Optional, Tuple, TypeVar, Union
 
-from ..interactions import ClientT, Interaction
+from ..interactions.base import ClientT, Interaction
+from ..interactions.message_component import MessageComponentInteraction
+from ..interactions.modal_submit import ModalSubmitInteraction
```
ooliver1, line 9: `ItemCallbackType` should probably use the component interaction class not the base one:
```python
Argument of type "(self: Self@View, button: Button[Unknown], interaction: CustomComponentInteraction) -> Coroutine[Any, Any, None]" cannot be assigned to parameter of type "ItemCallbackType[Button[V@button], ClientT@button]"
  Type "(self: Self@View, button: Button[Unknown], interaction: CustomComponentInteraction) -> Coroutine[Any, Any, None]" cannot be assigned to type "ItemCallbackType[Button[V@button], ClientT@button]"
    Parameter 3: type "Interaction[ClientT@button]" cannot be assigned to type "CustomComponentInteraction"
      "Interaction[ClientT@button]" is incompatible with "CustomComponentInteraction"Pylance[reportGeneralTypeIssues](https://github.com/microsoft/pyright/blob/main/docs/configuration.md#reportGeneralTypeIssues)
```

ooliver1 (changes requested):
Tested. Only issues were the warning and button type error mentioned in these comments. https://paste.nextcord.dev/?id=1702153210128711

PR #979, opened 2023-01-23
feat(message): send message flag

### LINKED ISSUE #978: Add support for setting flags when sending a message
### Summary

The library currently doesn't support setting Message flags

The core library

### The Problem

Link to API Docs: https://discord.com/developers/docs/resources/channel#create-message-jsonform-params

As you can see in the above linked docs, a field called "flags" is shown but the library currently doesn't support that field.

### The Ideal Solution

Preferably a `flags` kwarg that takes a list of `MessageFlags` should be added to `Messageable.send` and locked down to the currently supported flags like `SUPPRESS_EMBEDS` or an user friendlier solution would be a `suppress_embeds` kwarg that takes a bool and defaults to False.

### The Current Solution

N/A

### Additional Context

_No response_

MaskDuck:
## Summary

closes #978 

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -1419,8 +1419,7 @@ class Messageable:
             you specified both ``file`` and ``files``,
             or you specified both ``embed`` and ``embeds``,
             or the ``reference`` object is not a :class:`~nextcord.Message`,
-            :class:`~nextcord.MessageReference` or :class:`~nextcord.PartialMessage`,
-            or you specified both ``flags`` and ``suppress_embeds``.
+            :class:`~nextcord.MessageReference` or :class:`~nextcord.PartialMessage`
 
         Returns
         -------
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -1436,7 +1436,7 @@ class Messageable:
             flags = MessageFlags()
         if suppress_embeds is True:
             flags.suppress_embeds = True
-            
+
         flag_value: Optional[int] = flags.value if flags is not None else None
 
         embed_payload: Optional[EmbedData] = None
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -1436,7 +1436,7 @@ class Messageable:
         if suppress_embeds is not None:
             flags.suppress_embeds = suppress_embeds
 
-        flag_value: int = flags.value if flags.value is not 0 else None
+        flag_value: Optional[int] = flags.value if flags.value is not 0 else None
 
         embed_payload: Optional[EmbedData] = None
         embeds_payload: Optional[List[EmbedData]] = None
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -1246,6 +1246,7 @@ class Messageable:
         mention_author: Optional[bool] = ...,
         view: Optional[View] = ...,
         flags: Optional[MessageFlags] = ...,
+        suppress_embeds: Optional[bool] = ...,
     ) -> Message:
         ...
 
@@ -1265,6 +1266,7 @@ class Messageable:
         mention_author: Optional[bool] = ...,
         view: Optional[View] = ...,
         flags: Optional[MessageFlags] = ...,
+        suppress_embeds: Optional[bool] = ...,
     ) -> Message:
         ...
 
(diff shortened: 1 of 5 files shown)


ooliver1 (commented):
Test your own PRs and fix the pyright issues once again please.

teaishealthy (dismissed):
untested

nextcord/abc.py
```diff
@@ -1318,6 +1326,8 @@ async def send(
         reference: Optional[Union[Message, MessageReference, PartialMessage]] = None,
         mention_author: Optional[bool] = None,
         view: Optional[View] = None,
+        flags: Optional[MessageFlags] = None,
+        suppress_embeds: Optional[bool] = None,
```
EmreTech, line 1330: Why `Optional[bool]`? Why not just make the default `False`?

nextcord/abc.py
```diff
@@ -1411,6 +1432,12 @@ async def send(
         channel = await self._get_channel()
         state = self._state
         content = str(content) if content is not None else None
+        if flags is None:
+            flags = MessageFlags()
+        if suppress_embeds is True:
+            flags.suppress_embeds = True
```
EmreTech, line 1438: 1. `if x is True` violates PEP 8 recommendations.
2. Accompanied with my previous review comment, why not just set `flags.suppress_embeds` directly to `suppress_embeds`?

nextcord/abc.py
ooliver1: Same reason as `mention_author`, default to the API default.

nextcord/abc.py
ooliver1: This should instead simply check for `None` then pass it explicitly, as this relies on the API default - consistency.
```suggestion
        if suppress_embeds is not None:
            flags.suppress_embeds = suppress_embeds
```

nextcord/abc.py
```diff
@@ -1388,6 +1398,16 @@ async def send(
...
 
             .. versionadded:: 2.0
+        flags: Optional[:class:`~nextcord.MessageFlags`]
+            The message flags being set for this message.
+            Currently only :class:`~nextcord.MessageFlags.suppress_embeds` is able to be set.
+
+            .. versionadded:: 2.4
+        suppress_embeds: Optional[:class:`bool`]
+            Whether to suppress embeds on this message.
+            This cannot be set with ``flags``.
```
Skelmis, line 1408: Looks like it can be set in conjunction though?

ooliver1: I don't even see much point in `flags`, since if a flag isnt implemented, you can't really send it.

Soheab: it allows for future proofing I guess

ooliver1 (commented):
This seems to be settable when creating a forum post.

ooliver1 (approved):

nextcord/abc.py
```diff
@@ -1400,7 +1419,8 @@ async def send(
             you specified both ``file`` and ``files``,
             or you specified both ``embed`` and ``embeds``,
             or the ``reference`` object is not a :class:`~nextcord.Message`,
-            :class:`~nextcord.MessageReference` or :class:`~nextcord.PartialMessage`.
+            :class:`~nextcord.MessageReference` or :class:`~nextcord.PartialMessage`,
+            or you specified both ``flags`` and ``suppress_embeds``.
```
EmreTech, line 1423: `nextcord.InvalidArgument` isn't raised for this case?

nextcord/webhook/async_.py
```diff
@@ -481,6 +481,9 @@ def handle_message_parameters(
     view: Optional[View] = MISSING,
     allowed_mentions: Optional[AllowedMentions] = MISSING,
     previous_allowed_mentions: Optional[AllowedMentions] = None,
+    ephemeral: Optional[bool] = None,
+    flags: Optional[MessageFlags] = None,
```
EmreTech, line 485: Shouldn't an exception be raised for when `suppress_embeds`/`ephemeral` are mixed with `flags`?

EmreTech (changes requested):

nextcord/webhook/async_.py
EmreTech: What do you mean?

ooliver1: ephemeral just overrides flags

EmreTech (approved):

PR #980, opened 2023-01-23
refactor(app-cmds): use a metaclass for String and Range

ooliver1:
## Summary

microsoft/pyright#3799
`__class_getitem__` is apparently for generics,
This means that pyright believes they are `Type[Self]` not `Type[int]|Type[str]`.
Using a metaclass means pyright recognises parameters as the correct type.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -3437,31 +3437,9 @@ def unpack_annotation(
     return type_ret, literal_ret
 
 
-class Range:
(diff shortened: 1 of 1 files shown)


ooliver1:
Wow that diff, I would recommend using `split` view to review that.
image

Skelmis (approved):

DenverCoder1 (approved):

PR #989, opened 2023-02-08
feat(app-cmds): cog_command_error for application commands

### LINKED ISSUE #825: cog_command_error For Slash/Application Commands
### Summary

Setting cog_command_error to work with Slash Commands instead of only prefix commands.

Application Commands

### The Problem

As the name says, `cog_command_error` cannot be used for Slash Commands in cogs. Using this is useful for error handlers on specific cogs, I have these set for almost all my cogs but for Slash Commands due to them not working properly.

### The Ideal Solution

Either adding support for `cog_command_error` with Slash Commands or creating a new version of it named `cog_application_error`. Either works, as long as they are run once the cog errors due to an application command.

### The Current Solution

Using `on_application_command_error` in your Events could solve this issue. Simply using it the same way as `on_command_error` but with `interaction: nextcord.Interaction` instead of `ctx`, etc. (Note: This does not work for Cogs, its for all Slash/Application Commands)

### Additional Context

_No response_

ToasterUwU:
Fixes #825

## Summary

This adds `cog_application_command_error` which is the application command equivalent of `cog_command_error`.

## Checklist

- [X] This PR fixes an issue.


ToasterUwU:
have to add doc stuff, will do that in a moment

ToasterUwU:
Ok, so this should be working fine (i tested it) and should be documented completly.

HOWEVER, please check this super carefully. This is the first time that i added code to nextcord, not just changed it or updated the docs. 

So check carefully and give me feedback. Its possible i did something really wrong or stupid, because im not familiar with the codes structure yet.

nextcord/application_command.py
```diff
@@ -510,6 +515,13 @@ def _read_application_commands(self) -> None:
                     value.from_callback(value.callback)
                     self.__cog_application_commands__.append(value)
 
+    def has_error_handler(self) -> bool:
+        """:class:`bool`: Checks whether the cog has an error handler.
+
+        .. versionadded:: 2.4
+        """
+        return not hasattr(self.cog_application_command_error.__func__, "__cog_special_method__")
```
Skelmis, line 523: I'd say this naming + documentation is slightly misleading as it also provides the assumption the check handles prefix commands also.

I propose renaming it to something like `has_application_command_error_handler` and the documentation accordingly

nextcord/application_command.py
```diff
@@ -722,6 +761,12 @@ def from_callback(
         if self.name is None:
             self.name = self.callback.__name__
 
+        if self.error_callback is None:
+            if self.parent_cog:
+                self.error_callback = self.parent_cog._get_overridden_method(
+                    self.parent_cog.cog_application_command_error
+                )
+
```
Skelmis, line 769: May I ask what purpose this serves here?

None of the values check / used appear to be passed into this method so isn't this just the same as calling this within `__init__`?

nextcord/client.py
```diff
@@ -574,10 +574,12 @@ async def on_application_command_error(
         if interaction.application_command and interaction.application_command.has_error_handler():
             return
 
-        # TODO implement cog error handling
-        # cog = context.cog
-        # if cog and cog.has_error_handler():
-        #     return
+        if interaction.application_command == None:
```
Skelmis, line 577:
```suggestion
        if interaction.application_command is None:
```

nextcord/application_command.py
```diff
@@ -917,11 +962,10 @@ async def invoke_error(self, interaction: Interaction, error: Exception) -> None
...
         """
-        if self.has_error_handler() and self.error_callback is not None:
-            if self.parent_cog:
-                await self.error_callback(self.parent_cog, interaction, error)
-            else:
-                await self.error_callback(interaction, error)
+        if (
+            self.has_error_handler() and self.error_callback is not None
+        ):  # TODO Why is this here? self.has_error_handler does the same as the second check
+            await self.error_callback(interaction, error)
```
Skelmis, line 968: How does this handle passing `self` to callbacks if required? Just seems to be something from the removed code, so I'm just curious if that is still applicable here or not

Skelmis (changes requested):
I haven't tested this yet, will do after getting some more context from my current review items

nextcord/application_command.py
ToasterUwU: It is required, because the init method doesn't get a parent_cog object with it, if the slash_command decorator is used.
But when the ClientCog object is initialized, it goes trough all the application commands it contains and updates them with its own object.

So this is required to make the decorator way of adding slash commands work.

nextcord/application_command.py
ToasterUwU: It automatically does. The code I removed didn't work.
It passed self automatically, than another self via the object and than the rest args.
Which gave me the error of having 4 args when 3 were exspected.

The code I replaced it with works.

nextcord/application_command.py
Skelmis: Sweet, thanks

nextcord/application_command.py
```diff
@@ -722,6 +761,12 @@ def from_callback(
         if self.name is None:
             self.name = self.callback.__name__
 
+        if self.error_callback is None:
+            if self.parent_cog:
```
EmreTech, line 765: Couldn't you combine these two if statements together with an and statement?

nextcord/application_command.py
```diff
@@ -917,11 +962,10 @@ async def invoke_error(self, interaction: Interaction, error: Exception) -> None
         """|coro|
         Invokes the error handler if available.
         """
-        if self.has_error_handler() and self.error_callback is not None:
-            if self.parent_cog:
-                await self.error_callback(self.parent_cog, interaction, error)
```
EmreTech, line 922: Why isn't the parent cog supplied as a parameter anymore?

nextcord/application_command.py
ToasterUwU: Done

nextcord/application_command.py
ToasterUwU: Take a look at the [last disscussion with Skelmis](https://github.com/nextcord/nextcord/pull/989#discussion_r1103737943) on this exact same thing.

nextcord/application_command.py
```diff
@@ -510,6 +515,13 @@ def _read_application_commands(self) -> None:
                     value.from_callback(value.callback)
                     self.__cog_application_commands__.append(value)
 
+    def has_application_command_error_handler(self) -> bool:
+        """:class:`bool`: Checks whether the cog has an error handler for application commands.
+
+        .. versionadded:: 2.4
```
ooliver1, line 521:
```suggestion
        .. versionadded:: 2.5
```

nextcord/application_command.py
```diff
@@ -536,6 +548,27 @@ def cog_application_command_check(self, interaction: Interaction) -> bool:
         """
         return True
 
+    @_cog_special_method
+    async def cog_application_command_error(
+        self, interaction: Interaction, error: ApplicationError
+    ) -> None:
+        """A special method that is called whenever an error
+        is dispatched inside this cog.
```
ooliver1, line 556:
```suggestion
        """A special method that is called whenever an error is dispatched inside this cog.
```
Summaries should be on the first line.

ooliver1 (changes requested):

nextcord/application_command.py
```diff
@@ -510,6 +515,13 @@ def _read_application_commands(self) -> None:
                     value.from_callback(value.callback)
                     self.__cog_application_commands__.append(value)
 
+    def has_application_command_error_handler(self) -> bool:
+        """:class:`bool`: Checks whether the cog has an error handler for application commands.
+
+        .. versionadded:: 2.5
```
Skelmis, line 521:
```suggestion
        .. versionadded:: 2.6
```

nextcord/application_command.py
```diff
@@ -536,6 +548,28 @@ def cog_application_command_check(self, interaction: Interaction) -> bool:
...
+        self, interaction: Interaction, error: ApplicationError
+    ) -> None:
+        """A special method that is called whenever an error is dispatched inside this cog.
+
+        This is similar to :func:`.on_application_command_error` except only applying
+        to the commands inside this cog.
+
+        This **must** be a coroutine.
+
+        .. versionadded:: 2.5
```
Skelmis, line 562:
```suggestion
        .. versionadded:: 2.6
```

nextcord/client.py
```diff
@@ -574,10 +574,12 @@ async def on_application_command_error(
...
-        # TODO implement cog error handling
-        # cog = context.cog
-        # if cog and cog.has_error_handler():
-        #     return
+        if interaction.application_command is None:
+            return  # Not supposed to ever happen
+
+        cog = interaction.application_command.parent_cog
+        if cog and cog.has_application_command_error_handler():
+            return
```
Skelmis, line 582: Sound logic, but wouldn't this be better suited to living in `application_command.has_error_handler`?

Skelmis (changes requested):

nextcord/client.py
ToasterUwU: i simply copied the structure from the prefix commands and all that. Thats how it works there, so i think we should keep it consistend and keep it with slash commands like this as well.

nextcord/application_command.py
```diff
@@ -917,11 +962,10 @@ async def invoke_error(self, interaction: Interaction, error: Exception) -> None
...
         Invokes the error handler if available.
         """
-        if self.has_error_handler() and self.error_callback is not None:
-            if self.parent_cog:
-                await self.error_callback(self.parent_cog, interaction, error)
-            else:
-                await self.error_callback(interaction, error)
+        if (
+            self.has_error_handler() and self.error_callback is not None
+        ):  # TODO Why is this here? self.has_error_handler does the same as the second check
```
teaishealthy, line 967:
```suggestion
        if self.has_error_handler()
```

nextcord/client.py
```diff
@@ -574,10 +574,12 @@ async def on_application_command_error(
         if interaction.application_command and interaction.application_command.has_error_handler():
             return
 
-        # TODO implement cog error handling
-        # cog = context.cog
-        # if cog and cog.has_error_handler():
-        #     return
+        if interaction.application_command is None:
+            return  # Not supposed to ever happen
```
teaishealthy, line 578:
```suggestion
        if interaction.application_command is None:
            return  # Not supposed to ever happen

        if interaction.application_command.has_error_handler():
            return
```

teaishealthy (changes requested):

ooliver1 (approved):

alentoghostflame (approved):
Haven't tested, but LGTM! If you could merge from master and double check that everything works still, that would be sick.

ToasterUwU:
> Haven't tested, but LGTM! If you could merge from master and double check that everything works still, that would be sick.

I dont have the time to write a test and test it again, but I checked if anything has changed that is used by this addition, and non of the code I changed or use for the new stuff has been changed in any of the commits I just merged into the branch. So unless there is something that evades all logic, it should still work just as well. @alentoghostflame

nextcord/application_command.py
```diff
@@ -947,11 +992,8 @@ async def invoke_error(self, interaction: Interaction, error: Exception) -> None
         """|coro|
         Invokes the error handler if available.
         """
-        if self.has_error_handler() and self.error_callback is not None:
-            if self.parent_cog:
-                await self.error_callback(self.parent_cog, interaction, error)
-            else:
-                await self.error_callback(interaction, error)
+        if self.has_error_handler():
```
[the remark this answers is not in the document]
ToasterUwU: At that point, why even have the function to check it? I mean, if we never use it.

nextcord/client.py
```diff
@@ -578,13 +578,15 @@ async def on_application_command_error(
 
         This only fires if you do not specify any listeners for command error.
         """
-        if interaction.application_command and interaction.application_command.has_error_handler():
+        if interaction.application_command is None:
+            return  # Not supposed to ever happen
+
+        if interaction.application_command.has_error_handler():
             return
```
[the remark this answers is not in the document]
ToasterUwU: Don't understand the question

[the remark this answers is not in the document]
EmreTech: These if statements used to be combined together with the `and` operator. Why did you change that?

nextcord/application_command.py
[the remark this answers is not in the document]
EmreTech: Useful shorthand for users? Just because it's defined doesn't mean it has to be used by the underlying internal code.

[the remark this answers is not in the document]
ToasterUwU: I don't think this is a good idea to change. That would mean there is not one spot to change if it ever needs a change, but multiple.
And that, at least for me, is a bigger issue then pyright not being able to put a function with one line into context.

nextcord/client.py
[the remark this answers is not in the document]
ToasterUwU: I don't remember anymore, but I think it was because I was fiddling around with it, changed it a lot, and it just ended up like this without me giving it much more thought.

Or I did for the sake of readability, not sure after all this time

nextcord/application_command.py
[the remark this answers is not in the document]
EmreTech: > I don't think this is a good idea to change. That would mean there is not one spot to change if it ever needs a change, but multiple.

`has_error_callback` is currently only referenced once in `application_command.py`. With this suggestion, if error callbacks were changed in some way (maybe to allow multiple or something), it would be a small change. It's really not a big deal in this case.

If you're still not convinced, you can alternatively add this under the if statement:
```python
if TYPE_CHECKING:
    error_callback = cast(Callable, self.error_callback)

await error_callback(interaction, error)
```

nextcord/application_command.py
```diff
@@ -512,6 +517,13 @@ def _read_application_commands(self) -> None:
                     value.from_callback(value.callback)
                     self.__cog_application_commands__.append(value)
 
+    def has_application_command_error_handler(self) -> bool:
+        """:class:`bool`: Checks whether the cog has an error handler for application commands.
+
+        .. versionadded:: 2.6
```
EmreTech, line 523:
```suggestion
        .. versionadded:: 3.0
```

nextcord/application_command.py
```diff
@@ -538,6 +550,28 @@ def cog_application_command_check(self, interaction: Interaction) -> bool:
...
+        self, interaction: Interaction, error: ApplicationError
+    ) -> None:
+        """A special method that is called whenever an error is dispatched inside this cog.
+
+        This is similar to :func:`.on_application_command_error` except only applying
+        to the commands inside this cog.
+
+        This **must** be a coroutine.
+
+        .. versionadded:: 2.6
```
EmreTech, line 564:
```suggestion
        .. versionadded:: 3.0
```

EmreTech (changes requested):
This PR was not considered for 2.6.

ToasterUwU:
> This PR was not considered for 2.6.

I mean.. it was.. just like it was for 2.5

It just got moved back twice.

EmreTech (approved):
LGTM, haven't tested

Skelmis (changes requested):
The following simplistic bot errors, please fix.
Main: https://paste.nextcord.dev/?id=17002155473974
Cog: https://paste.nextcord.dev/?id=1700215573244250
Error: https://paste.nextcord.dev/?id=1700215592682863

ToasterUwU:
Found the issue and fixed it, Because fo the Type checking check, only when type checking there actually will be a error_callback assigned. So tests pass, running it doesn't work lol

Anyways, stupid oversight, fixed now.

Skelmis (approved):
Tested, lgtm

PR #995, opened 2023-02-22
refactor: use ffprobe even if located under /usr/bin

dotlambda:
## Summary
If I call `_probe_codec_native(source, "/bin/ffmpeg")` I want it to use `/bin/ffprobe`.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/player.py
+++ b/nextcord/player.py
@@ -557,7 +557,7 @@ class FFmpegOpusAudio(FFmpegAudio):
     def _probe_codec_native(
         source, executable: str = "ffmpeg"
     ) -> Tuple[Optional[str], Optional[int]]:
-        exe = executable[:2] + "probe" if executable in ("ffmpeg", "avconv") else executable
+        exe = executable[:-4] + "probe" if executable.endswith(("ffmpeg", "avconv")) else executable
         args = [
             exe,
             "-v",


dotlambda:
> Generally you have user installed binaries in `/usr/bin`.

True, I just wanted to shorten the commit message.

> I'm not against this change but the reason is a bit odd, shouldn't that be in your `$PATH`?

Not necessarily. I might e.g. want to test it with an ffmpeg executable I compiled locally to test a change in ffmpeg.

nextcord/player.py
```diff
@@ -557,7 +557,7 @@ async def probe(
     def _probe_codec_native(
         source, executable: str = "ffmpeg"
     ) -> Tuple[Optional[str], Optional[int]]:
-        exe = executable[:2] + "probe" if executable in ("ffmpeg", "avconv") else executable
+        exe = executable[:-4] + "probe" if executable.endswith(("ffmpeg", "avconv")) else executable
```
teaishealthy, line 560: Just a general question, shouldn't proper path manipulation be used for this?
`Path(executable).parent / "ffprobe"` or similar

Skelmis:
Hey @dotlambda, just bumping this. How are you feeling regarding the requested review comments?

ooliver1:
> Not necessarily. I might e.g. want to test it with an ffmpeg executable I compiled locally to test a change in ffmpeg.

That would then be `/usr/local/bin`, the order of your `PATH` variable would determine which binary is used.

Skelmis:
I do not necessarily agree with this for the sake of a change which may only benefit a small subset of users. Can you provide some further reasoning why the library should do this as a default versus you as a developer simply setting an item in the PATH with a higher precedence to be used?

ooliver1:
Closed for the above reasons with no response.

nextcord/player.py
dotlambda: You're right. I think
```suggestion
        exe = Path(executable).parent / "ffprobe" if Path(executable).name in ("ffmpeg", "avconv") else executable
```
makes the most sense.

dotlambda:
> Can you provide some further reasoning why the library should do this as a default versus you as a developer simply setting an item in the PATH with a higher precedence to be used?

My specific use case is in [Nixpkgs](https://github.com/NixOS/nixpkgs/tree/master/pkgs/development/python-modules/nextcord) where I want to set a default ffmpeg executable that's used by any user of nextcord. I can't set PATH for those users because they import a Python module rather than run an executable.
So what I'm trying to achieve is the ability to set `executable` to an absolute path.

PR #998, opened 2023-02-22
feat(file): add context manager to File

msnyder5:
## Summary

Add a context manager to `nextcord.File`. This will automatically call `close()` on the file when the context manager exits scope.

Also change `force_close` to be optional default to None so that we can set `force_close` to True when using context manager if `False` was not provided to init.

### Old Method
```python
file = nextcord.File(filepath)
await channel.send(file=file)
file.close()
```

### New Method (w/ Context Manager)
```python
with nextcord.File(filepath) as file:
    await channel.send(file=file)
```

## Checklist

- [ ] This PR fixes an issue.
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed) *Edit: This actually isn't a breaking change as you can still use the old syntax.*

### DIFF
--- a/nextcord/file.py
+++ b/nextcord/file.py
@@ -152,7 +152,7 @@ class File:
             self.force_close = True
         return self
 
-    def __close__(self, *_) -> None:
+    def __exit__(self, *_) -> None:
         self.close()
 
     def reset(self, *, seek: Union[int, bool] = True) -> None:
--- a/nextcord/file.py
+++ b/nextcord/file.py
@@ -13,6 +13,12 @@ class File:
     r"""A parameter object used for :meth:`abc.Messageable.send`
     for sending file objects.
 
+    .. versionadded:: 2.4
+
+        You can now use nextcord.File as a context manager. This will
+        automatically call `close()` when the context manager exits scope.
+        When using the context manager, force_close will default to True.
+
     .. note::
 
         File objects are single use and are not meant to be reused in
@@ -47,6 +53,7 @@ class File:
         This will also make the file bytes unusable by flushing it from
         memory after it is sent once.
         Enable this if you don't wish to reuse the same bytes.
+        Defaults to True when using context manager.
 
         .. versionadded:: 2.2
 
@@ -138,12 +145,14 @@ class File:
             self.filename is not None and self.filename.startswith("SPOILER_")
         )
 
-    def __enter__(self):
+    def __enter__(self) -> File:
+        # Set force_close to true when using context manager
+        # and force_close was not provided to __init__
         if self.force_close is None:
             self.force_close = True
         return self
 
-    def __close__(self):
+    def __close__(self) -> None:
         self.close()
 
     def reset(self, *, seek: Union[int, bool] = True) -> None:
--- a/nextcord/file.py
+++ b/nextcord/file.py
@@ -152,7 +152,7 @@ class File:
             self.force_close = True
         return self
 
-    def __close__(self) -> None:
+    def __close__(self, *_) -> None:
         self.close()
 
     def reset(self, *, seek: Union[int, bool] = True) -> None:
--- a/nextcord/file.py
+++ b/nextcord/file.py
@@ -90,7 +90,7 @@ class File:
         filename: Optional[str]
         description: Optional[str]
         spoiler: bool
-        force_close: bool
+        force_close: Optional[bool]
 
     def __init__(
         self,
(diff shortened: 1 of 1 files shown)


nextcord/file.py
```diff
@@ -13,6 +13,12 @@ class File:
     r"""A parameter object used for :meth:`abc.Messageable.send`
     for sending file objects.
 
+    .. versionadded:: 2.4
```
EmreTech, line 16:
```suggestion
    .. versionchanged:: 2.5
```

nextcord/file.py
```diff
@@ -47,6 +53,7 @@ class File:
         This will also make the file bytes unusable by flushing it from
         memory after it is sent once.
         Enable this if you don't wish to reuse the same bytes.
+        Defaults to True when using context manager.
```
ooliver1, line 56: To show as `True`.
```suggestion
        Defaults to ``True`` when using context manager.
```

nextcord/file.py
```diff
@@ -99,7 +106,7 @@ def __init__(
         *,
         description: Optional[str] = None,
         spoiler: bool = False,
-        force_close: bool = False,
+        force_close: Optional[bool] = None,
```
EmreTech, line 109: I don't get what changing the default of `force_close` accomplishes?

nextcord/file.py
ooliver1, line 16: 2.5 is the upcoming version now, 2.4 is released.
```suggestion
    .. versionadded:: 2.5
```

nextcord/file.py
msnyder5: Yeah I think that might be slightly opinionated. The thought process is that if you're using a context manager you want to make sure the file is closed, as leaving it open could result in unexpected behavior.

nextcord/file.py
```diff
@@ -13,6 +13,12 @@ class File:
     r"""A parameter object used for :meth:`abc.Messageable.send`
     for sending file objects.
 
+    .. versionadded:: 2.4
+
+        You can now use nextcord.File as a context manager. This will
+        automatically call `close()` when the context manager exits scope.
```
ooliver1, line 19: This will make a link in the docs. We use Sphinx to generate docs right now, and also ReStructuredText (RST). Here's the link to that "role"'s documentation:
<https://www.sphinx-doc.org/en/master/usage/restructuredtext/domains.html#role-py-meth>
image

```suggestion
        automatically call :meth:`close` when the context manager exits scope.
```

ooliver1 (approved):
Nice first PR, thanks for the explanation!

ooliver1 (changes requested):
Did not need to make that approved.

nextcord/file.py
EmreTech: Okay, but changing the default of the `force_close` parameter from `False` to `None` doesn't help. If the default was left alone, then when the context manager is used, the `force_close` attribute is automatically set to `True`, regardless of what its last value was.

msnyder5: I did it this way to allow `force_close` to be set to `False` while using the context manager if someone wants to.

EmreTech: I see.

nextcord/file.py
msnyder5: Isn't 2.4 the current version? I don't know how you guys version, but this isn't a breaking change so I think it's ok to be added to the current version.

Skelmis: Correct 2.4 is the current version, which means it's already released. This feature at the earliest can get into 2.5

EmreTech: We use a variant of [semantic versioning](https://semver.org/), which states that once a version is released, it cannot be modified.

msnyder5: Understood. Thanks for the info.

nextcord/file.py
```diff
@@ -138,6 +145,16 @@ def __init__(
...
         )
 
+    def __enter__(self) -> File:
+        # Set force_close to true when using context manager
+        # and force_close was not provided to __init__
+        if self.force_close is None:
+            self.force_close = True
+        return self
+
+    def __close__(self, *_) -> None:
```
EmreTech, line 155:
```suggestion
    def __exit__(self, *_) -> None:
```
https://docs.python.org/3/reference/datamodel.html#with-statement-context-managers

EmreTech (changes requested):

nextcord/file.py
msnyder5: Yeah, brainfart while copying it from one repo to the other.

Skelmis (approved):

EmreTech (approved):
Not tested.

alentoghostflame (approved):
Haven't tested but it LGTM!

ooliver1 (approved):

PR #999, opened 2023-02-23
feat(raw-models): add guild and user attr to RawMemberRemoveEvent

ooliver1:
## Summary

Previously, it was not possible to see what user left. This adds the attribute to `RawMemberRemoveEvent`, using a full user object ( :D finally a non-partial ) to get a full `User`.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/state.py
+++ b/nextcord/state.py
@@ -1691,8 +1691,6 @@ class ConnectionState:
             )
 
         raw = RawMemberRemoveEvent(data=data, state=self)
-        user = User(state=self, data=data["user"])
-        raw.user = user
         self.dispatch("raw_member_remove", raw)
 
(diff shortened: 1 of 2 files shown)


Skelmis (approved):

nextcord/state.py
```diff
@@ -1695,6 +1690,9 @@ def parse_guild_member_remove(self, data) -> None:
                 data["guild_id"],
             )
```
teaishealthy, line 1691: Remove this else block or update the log message as the event is still dispatched when the guild hasn't been found (raw.guild is None)

teaishealthy (changes requested):

teaishealthy (approved):

PR #1000, opened 2023-02-23
chore: add ruff linter

### LINKED ISSUE #1062: datetime.datetime.utcnow is deprecated in Python 3.12
See https://github.com/python/cpython/issues/103857

`utils.utcnow()` creates a timezone-aware datetime in the recommended way so it is fine to use as an alternative in most cases.

For an exact replacement, `utils.utcnow().replace(tzinfo=None)` can be used.

Most instances of `utcnow` and `utcfromtimestamp` were already fixed prior to the creation of nextcord, however one was still found.

```diff
index 725f498a76..c40a6ebbae 100644
--- a/nextcord/member.py
+++ b/nextcord/member.py
@@ -801,7 +801,7 @@ async def edit(
                 if not suppress:
                     voice_state_payload[
                         "request_to_speak_timestamp"
-                    ] = datetime.datetime.utcnow().isoformat()
+                    ] = utils.utcnow().isoformat()
                 await http.edit_voice_state(guild_id, self.id, voice_state_payload)
```

ooliver1:
## Summary

As the 1000th PR, here is a meta PR.
Ruff is a reimplementation of Flake8, popular plugins and other tools like isort.
This gives us an easy way to have a consistent code style without slowing down checks.
This should reduce the need for similar reviews on many PRs with similar style issues.
Unfortunately the ruff implementation does not allow for setting
different isort profiles per file, so examples show `nextcord` as a
first-party import for now.

Many rule types have been disabled with a `TODO` comment. This is due to previous attempts at this PR finding out that fixing large changes like `F` and `D` is a large change, and is better bit for separate PRs.

A list of diffs going through the rules fixed

[**B026**](https://beta.ruff.rs/docs/rules/star-arg-unpacking-after-keyword-arg/) - Star-arg unpacking after a keyword argument is strongly discouraged
```diff
@@ -267,10 +266,10 @@ def class_results_to_node(key, elements):
         ref = nodes.reference(
             "",
             "",
+            *[nodes.Text(element.label)],
             internal=True,
             refuri="#" + element.fullname,
             anchorname="",
-            *[nodes.Text(element.label)],
         )
```

[**RUF012**](https://beta.ruff.rs/docs/rules/mutable-class-default/) - Mutable class attributes should be annotated with `typing.ClassVar`
```diff
@@ -32,7 +34,7 @@ class DetailsDirective(Directive):
     final_argument_whitespace = True
     optional_arguments = 1

-    option_spec = {
+    option_spec: ClassVar[Dict[str, Any]] = {
         "class": directives.class_option,
         "summary-class": directives.class_option,
     }
```

[**B006**](https://beta.ruff.rs/docs/rules/mutable-argument-default/) - Do not use mutable data structures for argument defaults
```diff
-        lineno: int,
-        inliner: Inliner,
-        options: Dict = {},
-        content: List[str] = [],
+        _lineno: int,
+        _inliner: Inliner,
+        _options: Optional[Dict] = None,
+        _content: Optional[List[str]] = None,
```

[**PIE790**](https://beta.ruff.rs/docs/rules/unnecessary-pass/) - Unnecessary `pass` statement
```diff
--- a/examples/application_commands/sub_commands.py
+++ b/examples/application_commands/sub_commands.py
@@ -12,7 +12,6 @@ async def main(interaction: Interaction):
     This is the main slash command that will be the prefix of all commands below.
     This will never get called since it has subcommands.
     """
-    pass
```

[**RET505**](https://beta.ruff.rs/docs/rules/superfluous-else-return/) - Unnecessary `{branch}` after return statement
```diff
@@ -89,7 +89,7 @@ class TicTacToe(nextcord.ui.View):
             value = sum(across)
             if value == 3:
                 return self.O
-            elif value == -3:
+            if value == -3:
                 return self.X
```

[**SIM105**](https://beta.ruff.rs/docs/rules/suppressible-exception/) - Use `contextlib.suppress({exception})` instead of `try`-`except`-`pass`
```diff
-        try:
+        with contextlib.suppress(KeyError):
             options["rate_limit_per_user"] = options.pop("slowmode_delay")
-        except KeyError:
-            pass
```

[**ARG001**](https://beta.ruff.rs/docs/rules/unused-function-argument/) - Unused function argument: `{name}`
```diff
@@ -518,7 +516,7 @@ class GuildChannel:
         elif isinstance(obj, Role):
             predicate = lambda p: p.is_role()
         else:
-            predicate = lambda p: True
+            predicate = lambda _: True
```

[**PLR5501**](https://beta.ruff.rs/docs/rules/collapsible-else-if/) -  	Use `elif` instead of `else` then `if`, to reduce indentation
```diff
@@ -873,11 +871,10 @@ class GuildChannel:
                 raise InvalidArgument("No overwrite provided.")
                 raise InvalidArgument("No overwrite provided.")
             try:
                 overwrite = PermissionOverwrite(**permissions)
-            except (ValueError, TypeError):
-                raise InvalidArgument("Invalid permissions given to keyword arguments.")
-        else:
-            if len(permissions) > 0:
-                raise InvalidArgument("Cannot mix overwrite and keyword arguments.")
+            except (ValueError, TypeError) as e:
+                raise InvalidArgument("Invalid permissions given to keyword arguments.") from e
+        elif len(permissions) > 0:
+            raise InvalidArgument("Cannot mix overwrite and keyword arguments.")
```

[**B007**](https://beta.ruff.rs/docs/rules/unused-loop-control-variable/) - Loop control variable `{name}` not used within loop body
```diff
@@ -492,7 +491,7 @@ class ClientCog:
         """
         self.__cog_application_commands__ = []
         for base in reversed(self.__class__.__mro__):
-            for _, value in base.__dict__.items():
+            for value in base.__dict__.values():
```

[**SIM114**](https://beta.ruff.rs/docs/rules/if-with-same-arms/) - Combine `if` branches using logical `or` operator
```diff
     @description.setter
         if param_typing is self.parameter.empty:
             return ApplicationCommandOptionType.string
-        elif valid_type := self.option_types.get(param_typing, None):
-            return valid_type
-        elif (
-            # type(None) in typing.get_args(param_typing)  # TODO: Once Python 3.10 is standard, use this
-            type(None) in typing_extensions.get_args(param_typing)
-            and (
-                inner_type := find(
-                    lambda t: t is not type(None), typing_extensions.get_args(param_typing)
+
+        if (
+            # TODO: Once Python 3.10 is standard, use this
+            # type(None) in typing.get_args(param_typing)  # noqa: ERA001
+            (valid_type := self.option_types.get(param_typing, None))
+            or (
+                type(None) in typing_extensions.get_args(param_typing)
+                and (
+                    inner_type := find(
+                        lambda t: t is not type(None), typing_extensions.get_args(param_typing)
+                    )
                 )
+                and (valid_type := self.option_types.get(inner_type, None))
             )
-            and (valid_type := self.option_types.get(inner_type, None))
         ):
             return valid_type
```

[**TRY200**](https://beta.ruff.rs/docs/rules/reraise-no-cause/) - Use `raise from` to specify exception cause
```diff
@@ -1724,8 +1721,10 @@ class SlashCommandOption(BaseCommandOption, SlashOption, AutocompleteOptionMixin
                 # this looks messy but is too much effort to handle
                 # feel free to use typing.cast and if statements and raises
                 resolved_attachment_data = interaction.data["resolved"]["attachments"][value]  # type: ignore
-            except (AttributeError, ValueError, IndexError):
-                raise ValueError("Discord did not provide us interaction data for the attachment")
+            except (AttributeError, ValueError, IndexError) as e:
+                raise ValueError(
+                    "Discord did not provide us interaction data for the attachment"
+                ) from e
```

[**SIM210**](https://beta.ruff.rs/docs/rules/if-expr-with-true-false/) -  	Use `bool({expr})` instead of True if `{expr}` else False
```diff
@@ -2070,12 +2068,12 @@ class BaseApplicationCommand(CallbackMixin, CallbackWrapperMixin):
         """:class:`bool`: Returns ``True`` if this command is or should be registered to any guilds."""
         guild_only_ids = set(self.command_ids.keys())
         guild_only_ids.discard(None)
-        return True if (self.guild_ids_to_rollout or guild_only_ids) else False
+        return bool(self.guild_ids_to_rollout or guild_only_ids)
```

[**SIM108**](https://beta.ruff.rs/docs/rules/if-else-block-instead-of-if-exp/) - Use ternary operator `{contents}` instead of `if`-`else`-block
```diff
@@ -2090,20 +2088,12 @@ class BaseApplicationCommand(CallbackMixin, CallbackWrapperMixin):
         guild: Union[:class:`int`, :class:`Guild`]
             Guild or Guild ID to add this command to roll out to.
         """
         guild: Union[:class:`int`, :class:`Guild`]
             Guild or Guild ID to add this command to roll out to.
         """
-        if isinstance(guild, Guild):
-            # I don't like doing `guild = guild.id` and this keeps it extendable.
-            guild_id = guild.id
-        else:
-            guild_id = guild
-
-        self.guild_ids_to_rollout.add(guild_id)
+        self.guild_ids_to_rollout.add(guild.id if isinstance(guild, Guild) else guild)
```

[**RET504**](https://beta.ruff.rs/docs/rules/unnecessary-assign/) - Unnecessary assignment to `{name}` before `return` statement
```diff
@@ -3190,7 +3174,7 @@ def slash_command(
         if isinstance(func, BaseApplicationCommand):
             raise TypeError("Callback is already an application command.")

-        app_cmd = SlashApplicationCommand(
+        return SlashApplicationCommand(
             callback=func,
             name=name,
             name_localizations=name_localizations,
@@ -3202,7 +3186,6 @@ def slash_command(
             nsfw=nsfw,
             force_global=force_global,
         )
-        return app_cmd
```

[**ASYNC101**](https://beta.ruff.rs/docs/rules/open-sleep-or-subprocess-in-async-function/) - Async functions should not call `open`, `time.sleep`, or `subprocess` methods
it is a little annoying `noqa`ing the instances of this, but it is better than not linting accidental uses
```diff
@@ -438,9 +438,8 @@ class Asset(AssetMixin):
                 fp.seek(0)
             return written
-        else:
-            with open(fp, "wb") as f:
-                return f.write(data)
+
+        with open(fp, "wb") as f:  # noqa: ASYNC101
+            return f.write(data)
```

[**B010**](https://beta.ruff.rs/docs/rules/set-attr-with-constant/) - Do not call `setattr` with a constant attribute value. It is not any safer than normal property access.
```diff
@@ -340,7 +340,7 @@ class AuditLogChanges:
         elem: List[RolePayload],
     ) -> None:
         if not hasattr(first, "roles"):
-            setattr(first, "roles", [])
+            first.roles = []
```

[**C417**](https://beta.ruff.rs/docs/rules/unnecessary-map/) - Unnecessary `map` usage (rewrite using a {object_type})
```diff
@@ -2784,7 +2782,7 @@ class GroupChannel(abc.Messageable, abc.PrivateChannel, Hashable, PinsMixin):
         if len(self.recipients) == 0:
             return "Unnamed"

-        return ", ".join(map(lambda x: x.name, self.recipients))
+        return ", ".join((x.name for x in self.recipients))
```

[**T201**](https://beta.ruff.rs/docs/rules/print/) - `print` found
`noqa`ing all `stderr` prints too is quite annoying, better than not catching accidental prints (for "debugging")
would be better if ruff ignored prints with `file=`
```diff
@@ -563,7 +568,7 @@ class Client:
         overridden to have a different implementation.
         Check :func:`~nextcord.on_error` for more details.
         """
-        print(f"Ignoring exception in {event_method}", file=sys.stderr)
+        print(f"Ignoring exception in {event_method}", file=sys.stderr)  # noqa: T201
```

[**SLOT000**](https://beta.ruff.rs/docs/rules/no-slots-in-str-subclass/) - Subclasses of `str` should define `__slots__`
may be a bug to report in ruff, as this is an enum and the `str` is messed around with by enums anyway
```diff
@@ -129,7 +129,7 @@ class IntEnum(int, Enum):
         return self.value

-class StrEnum(str, Enum):
+class StrEnum(str, Enum):  # noqa: SLOT000
```

[**RUF001**](https://beta.ruff.rs/docs/rules/ambiguous-unicode-character-string/) - String contains ambiguous {}. Did you mean {}?
"ambiguous" characters are intentional here
```diff
@@ -608,7 +603,7 @@ class Locale(StrEnum):
     bg = "bg"
     """Bulgarian | български"""
     ru = "ru"
-    """Russian | Pусский"""
+    """Russian | Pусский"""  # noqa: RUF001
```

[**RSE102**](https://beta.ruff.rs/docs/rules/unnecessary-paren-on-raise-exception/) - Unnecessary parentheses on raised exception
```diff
@@ -276,7 +276,7 @@ def has_role(item: Union[int, str]) -> AC:

     def predicate(interaction: Interaction) -> bool:
         if interaction.guild is None:
-            raise ApplicationNoPrivateMessage()
+            raise ApplicationNoPrivateMessage
```

[**SIM114**](https://beta.ruff.rs/docs/rules/if-with-same-arms/) - Combine `if` branches using logical `or` operator
`in (tuple)` would be technically better, but this is 'faster' (double in terms of nanoseconds, quite negligible  but not something I want to change)
```diff
@@ -596,9 +594,7 @@ class DiscordWebSocket:
         """
         try:
             msg = await self.socket.receive(timeout=self._max_heartbeat_timeout)
-            if msg.type is aiohttp.WSMsgType.TEXT:
-                await self.received_message(msg.data)
-            elif msg.type is aiohttp.WSMsgType.BINARY:
+            if msg.type is aiohttp.WSMsgType.TEXT or msg.type is aiohttp.WSMsgType.BINARY:
```

[**B009**](https://beta.ruff.rs/docs/rules/get-attr-with-constant/) - Do not call `getattr` with a constant attribute value. It is not any safer than normal property access.
```diff
@@ -796,7 +792,7 @@ class DiscordVoiceWebSocket:
         self._close_code: Optional[int] = None
         self.secret_key: Optional[List[int]] = None
         self._hook: Optional[Callable[..., Awaitable[None]]] = (
-            hook or getattr(self, "_hook", None) or getattr(self, "_default_hook")
+            hook or getattr(self, "_hook", None) or self._default_hook
         )
```

[**B028**](https://beta.ruff.rs/docs/rules/no-explicit-stacklevel/) - No explicit `stacklevel` keyword argument found
```diff
@@ -3010,11 +2996,12 @@ class Guild(Hashable):
             raise InvalidArgument(
                 "Cannot pass both delete_message_days and delete_message_seconds."
             )
-        elif delete_message_days is not None:
+        if delete_message_days is not None:
             warnings.warn(
                 DeprecationWarning(
-                    "delete_message_days is deprecated, use delete_message_seconds instead."
-                )
+                    "delete_message_days is deprecated, use delete_message_seconds instead.",
+                ),
+                stacklevel=2,
             )
```

[**RUF006**](https://beta.ruff.rs/docs/rules/asyncio-dangling-task/) - Store a reference to the return value of `asyncio.{method}`
```diff
@@ -479,12 +481,12 @@ class Interaction(Hashable, Generic[ClientT]):

             async def inner_call(delay: float = delay) -> None:
                 await asyncio.sleep(delay)
-                try:
+                with contextlib.suppress(HTTPException):
                     await delete_func
-                except HTTPException:
-                    pass

-            asyncio.create_task(inner_call())
+            task = asyncio.create_task(inner_call())
+            self._background_tasks.add(task)
+            task.add_done_callback(self._background_tasks.discard)
```

**DTZ003** - The use of `datetime.datetime.utcnow()` is not allowed, use `datetime.datetime.now(tz=)` instead
This was likely supposed to be used in the 2.0 datetime refactor
```diff
@@ -801,9 +799,7 @@ class Member(abc.Messageable, _UserTag):
                 await http.edit_my_voice_state(guild_id, voice_state_payload)
             else:
                 if not suppress:
-                    voice_state_payload[
-                        "request_to_speak_timestamp"
-                    ] = datetime.datetime.utcnow().isoformat()
+                    voice_state_payload["request_to_speak_timestamp"] = utils.utcnow().isoformat()
```

[**PLCO105**](https://beta.ruff.rs/docs/rules/type-name-incorrect-variance/) - `{kind}` name "{param_name}" does not reflect its {variance}; consider renaming it to "{replacement_name}"
```diff
@@ -24,10 +24,10 @@ if TYPE_CHECKING:
     from ..types.components import ButtonComponent as ButtonComponentPayload
     from .view import View

-V = TypeVar("V", bound="View", covariant=True)
+V_co = TypeVar("V_co", bound="View", covariant=True)

-class Button(Item[V]):
+class Button(Item[V_co]):
```

## Checklist

And this so happens to fix #1062


teaishealthy (dismissed):

nextcord/guild.py
```diff
@@ -1904,8 +1918,9 @@ async def active_threads(self) -> List[Thread]:
     def fetch_members(
         self, *, limit: Optional[int] = 1000, after: Optional[SnowflakeTime] = None
     ) -> MemberIterator:
```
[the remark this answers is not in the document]
teaishealthy: that is not what i selected

teaishealthy:
I'd also update all remaining references to `discord.com/developers` for consistency :+1:

ooliver1:
The majority of those changes will be fixed with other rules

ooliver1:
For the removing blank lines between first line and others, usually docstrings have one summary line then a blank line - noted by the `D` group from pydocstyle. The blank `TYPE_CHECKING` block should be auto fixed with `TCH`.

teaishealthy:
Awesome. Although I'd still like this one https://github.com/nextcord/nextcord/pull/1000#discussion_r1116743033 :)

ooliver1:
This will get redone once I have several consecutive days to complete this, since merge conflicts completely ruin a PR of this nature.

Skelmis (commented):
Partial review

nextcord/ext/commands/bot.py
```diff
@@ -67,7 +68,7 @@
 CXT = TypeVar("CXT", bound="Context")
 
 
-def when_mentioned(bot: Union[Bot, AutoShardedBot], msg: Message) -> List[str]:
+def when_mentioned(bot: Union[Bot, AutoShardedBot], _msg: Message) -> List[str]:
```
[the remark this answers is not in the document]
ooliver1: It *is* unused here, but custom prefixes can make use of that message to have per-guild prefixes for example. It is a breaking change though yeah since you can use this function within a custom prefix function.

nextcord/member.py
```diff
@@ -156,7 +157,7 @@ def flatten_user(cls):
             # However I'm not sure how I feel about "functions" returning properties
             # It probably breaks something in Sphinx.
             # probably a member function by now
-            def generate_function(x):
+            def generate_function(x, value=value):
```
Skelmis, line 160: Why add this arg?

nextcord/state.py
```diff
@@ -1810,7 +1793,9 @@ def parse_guild_create(self, data) -> None:
 
         # check if it requires chunking
         if self._guild_needs_chunking(guild):
-            asyncio.create_task(self._chunk_and_dispatch(guild, unavailable))
+            task = asyncio.create_task(self._chunk_and_dispatch(guild, unavailable))
+            task.add_done_callback(lambda _t: self._chunk_tasks.pop(guild.id, None))
+            self._chunk_tasks[guild.id] = task
```
Skelmis, line 1798: This has the potential to be a race condition, however extremely unlikely.
```suggestion
            self._chunk_tasks[guild.id] = task
            task.add_done_callback(lambda _t: self._chunk_tasks.pop(guild.id, None))
```

Skelmis (changes requested):

docs/extensions/attributetable.py
```diff
@@ -159,11 +158,11 @@ def build_lookup_table(env):
         "class",
     }
 
-    for fullname, _, objtype, _, _, _ in domain.get_objects():
+    for fullname, __, objtype, *__ in domain.get_objects():
         if objtype in ignored:
             continue
 
-        classname, _, child = fullname.rpartition(".")
+        classname, __, child = fullname.rpartition(".")
```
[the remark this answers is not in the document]
Skelmis: I do not, didn't realise we used `_` for sphinx localing

nextcord/state.py
ooliver1: I don't think the event loop would context switch in the middle of a synchronous function, but that makes sense to be changed anyway.

nextcord/member.py
ooliver1: I believe this is [`B203`](https://beta.ruff.rs/docs/rules/function-uses-loop-variable/), which could probably just be `noqa`'d if you like, as this is used straight after. Not actually sure as to why `generate_function` exists and why it can't just be inlined alltogether though.

Skelmis: Neither, but still yea

Skelmis: Hmm, interesting. I see the reasoning with that link, has this been tested? As if it was relying on non B023 behavior there may be change..?

Skelmis (approved):
Untested due to the amount of things changed but lgtm

EmreTech (approved):

PR #1004, opened 2023-03-04
fix(app-cmds): before and after invoke hooks not working for function inside cogs

okay1204:
## Summary

Fixes an issue where the `before_invoke` and `after_invoke` decorators were not able to be used on functions inside Cogs because of the `self` parameter not being addressed. This fix simply checks how many arguments the passed in hook has and decides if the parent cog should be passed in for the self argument or not.

## Checklist

- [X] This PR fixes an issue.


nextcord/application_command.py
```diff
@@ -904,13 +904,22 @@ async def invoke_callback_with_hooks(
...
+                    await self._invoke_hook_param_check(after_invoke, interaction)
+
+    async def _invoke_hook_param_check(
+        self, hook: ApplicationHook, interaction: Interaction
+    ) -> None:
+        # invoke with self parameter if it exists
+        if len(signature(hook).parameters) == 2:
+            await hook(self.parent_cog, interaction)  # type: ignore
+        else:
+            await hook(interaction)  # type: ignore
```
teaishealthy, line 922:
```suggestion
        await hook(interaction)  # type: ignore
```

teaishealthy (dismissed):

okay1204:
@teaishealthy Reverted the previous change because it broke the fix. It needs the else statement or else the hook is called again with just the interaction parameter.

teaishealthy:
> Reverted the previous change because it broke the fix. It needs the else statement or else the hook is called again with just the interaction parameter.

Yeah I completely forgot the return in the if statement

Skelmis (approved):
Untested but overall looks fine to me

alentoghostflame:
@okay1204 If you fix the merge conflicts, I'll look into merging this in.

ooliver1 (approved):

PR #1007, opened 2023-03-09
refactor(channel): add invitable argument in create_thread method

Kevin-OVI:
## Summary

The `invitable` parameter was not present in the `TextChannel.create_thread` method for some reason even if the underlying state supported it, so I added it.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/channel.py
+++ b/nextcord/channel.py
@@ -716,6 +716,7 @@ class TextChannel(abc.Messageable, abc.GuildChannel, Hashable, PinsMixin):
         message: Optional[Snowflake] = None,
         auto_archive_duration: ThreadArchiveDuration = MISSING,
         type: Optional[ChannelType] = None,
+        invitable: bool = True,
         reason: Optional[str] = None,
     ) -> Thread:
         """|coro|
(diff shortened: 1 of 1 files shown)


nextcord/channel.py
```diff
@@ -742,6 +743,9 @@ async def create_thread(
             The type of thread to create. If a ``message`` is passed then this parameter
             is ignored, as a thread created with a message is always a public thread.
             By default this creates a private thread if this is ``None``.
+        invitable: :class:`bool`
+            Whether non-moderators can add other non-moderators to this thread.
+            Only available for private threads and without message.
```
teaishealthy, line 748: >  and without message.

Could you elaborate on what this means?

ooliver1, line 748:
```suggestion
            Only available for private threads and created without ``message``.
```
It isn't too clear as to what "and without message" means (as seen above).

ooliver1 (changes requested):

nextcord/channel.py
```diff
@@ -742,6 +743,9 @@ async def create_thread(
             The type of thread to create. If a ``message`` is passed then this parameter
             is ignored, as a thread created with a message is always a public thread.
             By default this creates a private thread if this is ``None``.
+        invitable: :class:`bool`
+            Whether non-moderators can add other non-moderators to this thread.
+            Only available for private threads and created without ``message``.
```
teaishealthy, line 748:
```suggestion
            Only available for private threads and threads created without a ``message``.
```

teaishealthy (changes requested):

nextcord/channel.py
```diff
@@ -716,6 +716,7 @@ async def create_thread(
         message: Optional[Snowflake] = None,
         auto_archive_duration: ThreadArchiveDuration = MISSING,
         type: Optional[ChannelType] = None,
+        invitable: bool = True,
```
ooliver1, line 719: Ideally this should default to `None`, which means "default to Discord's default" (by not adding unnecessary fields to the payload)

ooliver1 (approved):

teaishealthy (approved):

PR #1021, opened 2023-03-26
fix: ForumChannel and Thread returning wrong ForumTags

Fenix5fire:
## Summary

Fixed ForumChannel.available_tags only returned last tag by wrong id: used wrong dict to get id
Fixed Thread.applied_tags and Thread.applied_tag_ids returned empty list: used non-existent key in getting tags data from payload
ForumTag id is no longer optional when getting from payload

## Checklist

- [x] This PR fixes an issue.


Skelmis (approved):
Not tested but appears to line up with docs + was in the help thread for this so probs fine by me

EmreTech (approved):
Tested.

PR #1026, opened 2023-03-31
feat(flags): add ApplicationFlags.application_auto_moderation_rule_create

spifory:
## Summary

Implements discord/discord-api-docs#6021

Note that I am unsure if the name of the flag is okay, I've just used what it was called in the documentation.

## Checklist

- [ ] This PR fixes an issue.

### DIFF
--- a/nextcord/flags.py
+++ b/nextcord/flags.py
@@ -1195,7 +1195,10 @@ class ApplicationFlags(BaseFlags):
 
     @flag_value
     def application_auto_moderation_rule_create(self):
-        """:class:`bool`: Returns ``True`` if the application uses the Auto Moderation API."""
+        """:class:`bool`: Returns ``True`` if the application uses the Auto Moderation API.
+
+        .. versionadded::
+        """
         return 1 << 6
 
     @flag_value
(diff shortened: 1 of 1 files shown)


toolifelesstocode:
The client does say that, however, remember that some names on the client are different on the API. Such as how there is 'guild' and 'server'

Skelmis (approved):
Looks good to me, i also agree with the naming as it keeps it consistent. All other values in the class mirror the documentation 1-1 so while this can be considered long it would also be consistent with both the class and discord docs

nextcord/flags.py
```diff
@@ -1193,6 +1193,11 @@ class ApplicationFlags(BaseFlags):
         rather than using this raw value.
     """
 
+    @flag_value
+    def application_auto_moderation_rule_create(self):
+        """:class:`bool`: Returns ``True`` if the application uses the Auto Moderation API."""
```
ooliver1, line 1198:
```suggestion
        """:class:`bool`: Returns ``True`` if the application uses the Auto Moderation API.

        .. versionadded:: 2.5
        """
```

ooliver1 (changes requested):

nextcord/flags.py
```diff
@@ -1193,6 +1193,14 @@ class ApplicationFlags(BaseFlags):
         rather than using this raw value.
     """
 
+    @flag_value
+    def application_auto_moderation_rule_create(self):
+        """:class:`bool`: Returns ``True`` if the application uses the Auto Moderation API.
+
+        .. versionadded:: 2.5
```
EmreTech, line 1200:
```suggestion
        .. versionadded:: 2.6
```

EmreTech (approved):

PR #1053, opened 2023-04-30
feat(slash_utils): add choices decorator

Soheab:
## Summary

I noticed an undocumented `describe` decorator at `ext/slash_utils` similar to discord.py but that also has one called `choices` which is what this PR is for.

## Checklist

- [ ] This PR fixes an issue.


Skelmis (commented):
I'm aware `describe` is not documented, but could we possibly document em as well?

Soheab:
> I'm aware `describe` is not documented, but could we possibly document em as well?

I have documented it under Application commands > command helpers. I don't think it needs a separate ext tab.

Issue #1002, opened 2023-02-27
User agents not always provided


vcokltfre:
### Summary

Webhooks, interactions, and CDN downloads use the default aiohttp UA, rather than the lib-specific one.

Related Disnake PR: https://github.com/DisnakeDev/disnake/pull/959
Why this matters (DDevs): https://discord.com/channels/613425648685547541/697138785317814292/879488478793060382

PR #1057, opened 2023-05-13
refactor: HTTPClient Re-Core

alentoghostflame:
## Summary

Massively reworks the core of HTTP, adding support for:
- Multiple authorization tokens/types with the `auth` kwarg, opening the way for OAuth2 extensions,
- Global ratelimits per authorization, to stay within the (default) 50 requests per second per auth maximum,
- Proper rate limit bucket handling, allowing multiple requests in a single bucket to run simultaneously, unlike the previous implementation,
- The `retry_request` kwarg, allowing the user to disable auto-retries on ratelimit or Discord server errors.

Due to restructuring, there HAS been some breaking changes.
- Documentation and comments will likely need to be updated across the board.
- `MaybeUnlock` has been removed and replaced with a completely different `RateLimit`.
- `HTTPClient.request()` will no longer overwrite specific provided headers, allowing you, for example, to override the auth and user-agent.
- Due to `HTTPClient` no longer being centered around a single token, the `HTTPClient.token` attribute has been removed and `Client._token` has been added.
  - This can be added back of course, but hmm.
- While `HTTPClient` may technically not be documented, almost all of its attributes were not `_`'d, and thus were technically public. Since basically all of those attributes are gone/renamed, definitely a breaking change there,
- `HTTPClient.static_login()` was modified to better fit with the rework of `HTTPClient`.

Notes:
- `Webhook` and `HTTPClient` had their own rate limiting implementations before this PR, and with this PR still do. To avoid more delays, migrating webhooks to use `HTTPClient` is not in this PR.

## Checklist

- [ ] This PR fixes an issue.
- [X] This PR is a breaking change (e.g. methods or parameters removed/renamed)


alentoghostflame:
oh shoot I knew I forgot something

nextcord/http.py
```diff
@@ -155,25 +154,255 @@ def bucket(self) -> str:
...
-        self._unlock: bool = True
+class RateLimitMigrating(BaseException):
+    ...
+
+
+class RateLimit:
+    def __init__(self, time_offset: float = 0.3) -> None:
+        """Used to time gate a large batch of requests to only occur X every Y seconds. Used via ``async with``
+
+        NOT THREAD SAFE.
```
Skelmis, line 165: How does this work with voice things? Iirc that's thread based in some aspects or at-least it used to be

nextcord/http.py
```diff
@@ -155,25 +154,255 @@ def bucket(self) -> str:
...
+        """When this RateLimit is being deprecated and acquiring requests need to migrate to a different RateLimit, this
+        variable should be set to the different RateLimit/buckets string name.
+        """
+
+    async def update(self, response: aiohttp.ClientResponse) -> None:
+        """Updates the rate limit with information found in the response. Specifically the headers."""
+
+        if (
+            response.headers.get("X-RateLimit-Global") or response.headers.get("x-ratelimit-global")
+        ) in (True, "true"):
```
Skelmis, line 204: Possibly want to `.lower()` this?

nextcord/http.py
```diff
@@ -155,25 +154,255 @@ def bucket(self) -> str:
...
+            self.remaining,
+            self.reset,
+            self.reset_after,
+        )
+
+    def start_reset_task(self) -> None:
+        """Starts the reset task, non-blocking."""
+        loop = asyncio.get_running_loop()
+        _log.debug("Bucket %s will reset after %s seconds.", self.bucket, self.reset_after)
+        self._reset_remaining_task = loop.create_task(self.reset_remaining(self.reset_after))
```
Skelmis, line 278: possibly want to check for existing tasks? might overwrite an already queued task is my thoughts

nextcord/http.py
```diff
@@ -155,25 +154,255 @@ def bucket(self) -> str:
...
+    def release(self) -> None:
+        # Basically a placeholder, could probably be removed ;)
+        pass
+
+
+class GlobalRateLimit(RateLimit):
+    """
+    Represents the global rate limit, and thus has to have slightly modified behavior.
+
+    Still not thread safe.
```
Skelmis, line 367: see prior

nextcord/http.py
```diff
@@ -155,25 +154,255 @@ def bucket(self) -> str:
...
+        # As updates are little weird, it's best to start the reset task as soon as the first request has acquired.
+        if not self._reset_remaining_task or self._reset_remaining_task.done():
+            self.start_reset_task()
+
+        return ret
+
+    async def update(self, response: aiohttp.ClientResponse) -> None:
+        if (
+            response.headers.get("X-RateLimit-Global") or response.headers.get("x-ratelimit-global")
+        ) not in (True, "true"):
```
Skelmis, line 381: see prior

nextcord/http.py
```diff
@@ -269,152 +574,185 @@ async def request(
...
+                            #     "http_ratelimit",  # TODO: Uhhhhhhhhhhh
+                            #     int(limit),
+                            #     int(remaining),
+                            #     delta,
+                            #     bucket,
+                            #     response.headers.get("X-RateLimit-Scope"),
+                            # )
+
+                            if response.status >= 400:
+                                if response.status in {500, 502, 504}:
```
Skelmis, line 681: Couldnt this just check if its 500 or higher

nextcord/http.py
alentoghostflame: does voice use HTTP rate limits? I thought that was a websocket/gateway sort of thing.

nextcord/http.py
alentoghostflame: The `if response.status in {500, 502, 504}` was lifted from the original `HTTPClient.request()` code, changing it to `>= 500` is a good idea.

nextcord/http.py
alentoghostflame: I actually did this while fixing another issue I found, lol. Will push once I get some of the other issues I found worked out.

nextcord/http.py
```diff
@@ -155,25 +155,286 @@ def bucket(self) -> str:
...
+    ...
+
+
+class IncorrectBucket(DiscordException):
+    ...
+
+
+class RateLimit:
+    def __init__(self, time_offset: float = 0.3) -> None:
+        """Used to time gate a large batch of requests to only occur X every Y seconds. Used via ``async with``
```
ooliver1, line 168: Usually we have docstrings under `class` instead of under `__init__`

nextcord/http.py
```diff
@@ -155,25 +155,286 @@ def bucket(self) -> str:
...
+
+        NOT THREAD SAFE.
+
+        Parameters
+        ----------
+        time_offset: :class:`float`
+            Number in seconds to increase all timers by. Used for lag compensation.
```
ooliver1, line 175: This seems to default to quite an arbitrary value and be a bit unnecessary. `unsync_clock` meant that `datetime.now()` was not trustable (out of sync, so it would send requests too early or late), and so `X-RateLimit-Reset` was also not able to be used.

`x_reset_after = float(x_reset_after) + self._time_offset`
If there was a weird latency difference one way to the other, `reset_after` would be adjusted to be longer, but surely it has already decreased by the time the header was received.

If Discord sends `4.5` as a reset-after, and it takes `0.3`s to get back to the client, the reset after is actually `4.2` now since `0.3`s has already elapsed. So it is safe to not add an "offset".

nextcord/http.py
```diff
@@ -155,25 +155,286 @@ def bucket(self) -> str:
...
+    @property
+    def resetting(self) -> bool:
+        return self._reset_remaining_task is not None and not self._reset_remaining_task.done()
+
+    async def update(self, response: aiohttp.ClientResponse) -> None:
+        """Updates the rate limit with information found in the response. Specifically the headers."""
+
+        if (
+            response.headers.get("X-RateLimit-Global") or response.headers.get("x-ratelimit-global")
+        ) in (True, "true"):
```
ooliver1, line 213: Headers should be strings already.

nextcord/http.py
```diff
@@ -155,25 +155,286 @@ def bucket(self) -> str:
...
+
+        # Updates the reset-after count, being pessimistic.
+        x_reset_after = response.headers.get("X-RateLimit-Reset-After")
+        if x_reset_after is not None:
+            x_reset_after = float(x_reset_after) + self._time_offset
+            if self.reset_after is None:
+                self.reset_after = x_reset_after
+            else:
+                if self.reset_after < x_reset_after:
+                    _log.info(
```
ooliver1, line 264: Seems more like a `DEBUG` log to me.

nextcord/http.py
alentoghostflame: When initially tested this implementation, before I decided to move it into NC, I kept running into 429's seemingly caused by Discord servers not being properly time synced with one another, but perhaps I need to retest that.

nextcord/http.py
Skelmis: Hmm thats true, fair enough

nextcord/http.py
alentoghostflame: After setting it to 0.0 and running a quick test, it seems fine. Perhaps I'm misremembering what happened in the past. I'll default it to 0.0 and keep it in for now unless something happens in testing or you want it removed entirely.

nextcord/http.py
```diff
@@ -182,49 +440,144 @@ def __exit__(
...
+        self._url_deny_list: set[tuple[str, str, str | None]] = set()
+        """{("METHOD", "route", "auth string"), ...}
+
+        Paths that result in a 404. These will immediately error if a request is made to them.
+        """
+
+    def _make_global_rate_limit(self, auth: str | None, max_per_second: int) -> GlobalRateLimit:
+        _log.info(
+            "Creating global ratelimit for auth %s with max per second %s.", auth, max_per_second
+        )
```
Skelmis, line 523: Yea, no. I don't approve of bot tokens being in logs, especially at info level

nextcord/http.py
```diff
@@ -182,49 +440,144 @@ def __exit__(
...
+        rate_limit.limit = max_per_second
+        rate_limit.remaining = max_per_second
+        rate_limit.reset_after = 1 + self._time_offset
+        rate_limit.bucket = f"Global {auth if auth else 'Unauthorized'}"
+
+        self._global_rate_limits[auth] = rate_limit
+        return rate_limit
+
+    def _make_url_rate_limit(self, method: str, route: Route, auth: str | None) -> RateLimit:
+        _log.info("Making URL rate limit for %s %s %s", method, route.bucket, auth)
```
Skelmis, line 534: Yea, no. I don't approve of bot tokens being in logs, especially at info level

Skelmis (changes requested):

Skelmis:
Regardless I also think that kind of information is too high level / frequent for `info`

nextcord/http.py
alentoghostflame: Fair, the latest commit should be an adequate fix?

nextcord/http.py
alentoghostflame: Fair, the latest commit should be an adequate fix?

nextcord/http.py
```diff
@@ -269,152 +624,232 @@ async def request(
...
+                                        rate_limit_path,
+                                    )
+                                    await asyncio.sleep(1 + retry_count * 2)
+                                    should_retry = True
+                                elif response.status == 401:
+                                    _log.warning(
+                                        "Path %s resulted in error 401, rejected authorization?",
+                                        rate_limit_path,
+                                    )
+                                    raise HTTPException(response, ret)
```
EmreTech, line 777: Any plans to make a specific Unauthorized exception class?

nextcord/http.py
```diff
@@ -155,25 +164,283 @@ def bucket(self) -> str:
...
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
     ) -> None:
-        if self._unlock:
-            self.lock.release()
+        self.release()
+
+    def locked(self) -> bool:
+        return self.remaining <= 0
```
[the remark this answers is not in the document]
alentoghostflame: Theoretically, never.
Hypothetically, if multiple `self.remaining -= 1`'s are called.
Practically, if there's a bug that causes it to go below zero I really hope it sees it as locked.
It's never intended to go below zero, but I still let it check below zero as a backup.

nextcord/http.py
```diff
@@ -45,9 +47,7 @@
 _log = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
-    from types import TracebackType
-
-    from typing_extensions import Self
+    pass
```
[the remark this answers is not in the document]
alentoghostflame: No clue. Removing.

nextcord/http.py
```diff
@@ -155,25 +164,283 @@ def bucket(self) -> str:
...
+        """When this RateLimit is being deprecated and acquiring requests need to migrate to a different RateLimit, this
+        variable should be set to the different RateLimit/buckets string name.
+        """
+
+    @property
+    def resetting(self) -> bool:
+        return self._reset_remaining_task is not None and not self._reset_remaining_task.done()
+
+    async def update(self, response: aiohttp.ClientResponse) -> None:
+        """Updates the rate limit with information found in the response. Specifically the headers."""
```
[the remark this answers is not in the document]
alentoghostflame: Updating the docstring to be more generic, as `update()` uses the response headers and status.

nextcord/http.py
```diff
@@ -155,25 +164,283 @@ def bucket(self) -> str:
...
+            # The response is intended for the regular rate limit, not a global rate limit.
+            return
+
+        if response.status == 429:
+            # Oh dear, we hit the rate limit.
+            _log.warning("Global rate limit 429 encountered, setting remaining to 0.")
+            self.remaining = 0
+            data = await response.json()
+            _log.warning(data)
+            if response.headers.get("X-RateLimit-Scope") == "global":
```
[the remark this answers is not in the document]
alentoghostflame: Not really? This specifically checks if the rate limit is in the global scope. I'm checking it because I don't trust Discord to not do some funky thing where it's a global rate limit, but it's not in the global scope for some reason?
Also adjusted where `await response.json()` is ran to be inside of the `if` statement.

nextcord/http.py
alentoghostflame: I was going to say no, but then I remembered that `request()` actually throwing a status `401` related error was never specifically handled, or suppose to happen in the old implementation. Adding and fixing!

nextcord/http.py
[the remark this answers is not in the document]
EmreTech: I guess that could be possible, depending on how the value of `X-RateLimit-Global` is calculated...

nextcord/http.py
```diff
@@ -269,152 +623,245 @@ async def request(
...
+                                        "Path %s resulted in error 401, rejected authorization?",
+                                        rate_limit_path,
+                                    )
+                                    raise HTTPException(response, ret)
+                                elif response.status == 403:
+                                    _log.warning(
+                                        "Path %s resulted in error 403, check your permissions?",
+                                        rate_limit_path,
+                                    )
+                                    raise Unauthorized(response, ret)
```
EmreTech, line 793: You accidentally replaced the type of this exception with `Unauthorized`.

alentoghostflame: thank you, fixing.

TheOriginalSoni:
Commenting to follow PR - Waiting for resolution on 

"Nextcord waits out 429 which causes time delays for things like (Change channel name multiple times in a row). Should throw an error instead"

nextcord/http.py
```diff
@@ -155,25 +164,280 @@ def bucket(self) -> str:
...
+            pass  # Don't need to set it again.
+        elif self.bucket is None:
+            self.bucket = x_bucket
+        else:
+            raise IncorrectBucket(
+                f"Update given for bucket {x_bucket}, but this RateLimit is for bucket {self.bucket}!"
+            )
+
+        if response.status == 404:
+            self._deny = True
```
EmreTech, line 239: I don't like this automatic denial system. If a bot repeatedly calls the same request with the same parameters that already 404ed, then that's the fault of the bot for not properly handling those 404s. Nextcord shouldn't be responsible for that.

nextcord/http.py
```diff
@@ -268,152 +663,250 @@ async def request(
...
+        #  too much, 429s keep getting hit, or something is internally wrong.
+        for retry_count in range(max_retry_count):  # To prevent infinite loops.
+            should_retry = False
+            try:
+                async with global_rate_limit:
+                    async with url_rate_limit:
+                        # This check is for asyncio.gather()'d requests where the rate limit can change.
+                        if (
+                            temp := self._get_url_rate_limit(route.method, route, auth)
+                        ) is not url_rate_limit and not None:
```
EmreTech, line 698:
```suggestion
                        ) not in (url_rate_limit, None):
```

nextcord/http.py
```diff
@@ -155,25 +164,280 @@ def bucket(self) -> str:
...
+class RateLimitMigrating(DiscordException):
+    ...
+
+
+class IncorrectBucket(DiscordException):
+    ...
+
+
+class RateLimit:
+    """Used to time gate a large batch of requests to only occur X every Y seconds. Used via ``async with``
```
EmreTech, line 176:
```suggestion
    """Used to time gate a large batch of requests to only occur X every Y seconds. Used via an async context manager.
```

EmreTech (changes requested):

nextcord/http.py
alentoghostflame: IIRC, this was added with `asyncio.gather()` in mind. In the event that a large batch of requests are gathered up and a mispelling in the route was made, it would cancel almost all of them to avoid Discord getting upset and Cloudflare possibly taking out the ban hammer. It seemed to me that it was a relatively free check that could save a major headache for some people. I can remove if though if you wish.

nextcord/http.py
ooliver1: Not too sure about that one, identity checks are much faster so are better when you can.

nextcord/http.py
```diff
@@ -182,49 +446,164 @@ def __exit__(
...
-        self.bot_token: bool = False
-        self.proxy: Optional[str] = proxy
-        self.proxy_auth: Optional[aiohttp.BasicAuth] = proxy_auth
-        self.use_clock: bool = not unsync_clock
-        self._dispatch: Callable = dispatch
+        self._connector = connector
+        self._default_max_per_second = default_max_per_second
+        """Maximum amount of requests per second per authorization."""
+        self._time_offset = time_offset
+        """Amount of seconds added to all ratelimit timers for lag compensation."""
```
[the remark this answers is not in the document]
alentoghostflame: That's a fair point, duplicate documentation is asking for it to be out of sync.

nextcord/http.py
```diff
@@ -182,49 +446,164 @@ def __exit__(
...
+
+    def _set_url_rate_limit(
+        self, method: str, route: Route, auth: str | None, rate_limit: RateLimit
+    ) -> None:
+        self._url_rate_limits[(method, route.bucket, auth)] = rate_limit
+
+    def _get_url_rate_limit(self, method: str, route: Route, auth: str | None) -> RateLimit | None:
+        return self._url_rate_limits.get((method, route.bucket, auth), None)
+
+    def _set_default_auth(self, auth: str | None) -> None:
```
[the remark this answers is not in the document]
alentoghostflame: _set_default_auth had different behavior in the past, and I wanted to make setting the default auth not be a casual `self._default_auth = x` thing. I'll think about removing it unless additional people chime in that it should stay or be removed.

nextcord/http.py
```diff
@@ -237,28 +616,44 @@ async def request(
...
+        """|coro|
+
+        Makes an API request to Discord, handling authorization (if needed), rate limits, and limited error handling.
+
+        Parameters
+        ----------
+        route: :class:`Route`
+            The Discord Route to make the API request for.
+        files: Optional[Sequence[:class:`File`]]
+            pass
```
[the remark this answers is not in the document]
alentoghostflame: Thank you!

nextcord/http.py
```diff
@@ -237,28 +616,44 @@ async def request(
...
+        auth: :class:`str` | `None` | `MISSING`
+            Authorization string to use. Will not auto-format given tokens. For example, a bot token must be provided
+            as "Bot <bot token>". If set to `None`, no authorization will be used. If left unset, the default
+            auth will be used. (if the default is not set, no auth will be used.)
+        retry_request: :class:`bool`
+            If the request should be retried in specific cases. This mainly concerns 500 errors (Discord server issues)
+            or 429s. (ratelimit issues)
+        kwargs
+            This is purposefully undocumented. Behavior of extra kwargs may change in a breaking way at any point, and
```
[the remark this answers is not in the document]
alentoghostflame: Currently yes, but I don't want behavior of additional kwargs to be defined or relied on in any way. Mentioning that additional kwargs are simply routed to `ClientSession.request` implies that it may stay that way.

nextcord/http.py
```diff
@@ -237,28 +616,44 @@ async def request(
...
+            extra kwargs may not be allowed in the future.
+
+        Returns
+        -------
+        pass
```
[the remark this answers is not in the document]
alentoghostflame: Thank you!

nextcord/http.py
```diff
@@ -268,152 +663,250 @@ async def request(
...
-                        if remaining == "0" and response.status != 429:
-                            # we've depleted our current bucket
-                            delta = utils.parse_ratelimit_header(response, use_clock=self.use_clock)
+        auth = headers.get("Authorization")
+
+        # If a global rate limit for this authorization doesn't exist yet, make it.
+        if (global_rate_limit := self._global_rate_limits.get(auth)) is None:
+            global_rate_limit = self._make_global_rate_limit(auth, self._default_max_per_second)
+
+        global_rate_limit = cast(GlobalRateLimit, global_rate_limit)
```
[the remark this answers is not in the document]
alentoghostflame: Pyright may have gotten upset about it, I'll comment it out in my next commit and remove it in the one after that.

nextcord/http.py
alentoghostflame: *SIGH*, why do I occasionally do this. Fixed.

nextcord/http.py
```diff
@@ -268,152 +663,250 @@ async def request(
...
+                                kwargs.get("data"),
+                                response.status,
                             )
-                            maybe_lock.defer()
-                            self.loop.call_later(delta, lock.release)
-
-                        # the request was successful so just return the text/json
-                        if 300 > response.status >= 200:
-                            _log.debug("%s %s has received %s", method, url, data)
-                            return data
```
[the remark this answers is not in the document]
alentoghostflame: `should_retry` is `False` by default. If no errors are raised and it's never set to `True`, the if statement checking `should_retry` will break execution out of the loop and it will hit the `return` at the end of the function. Different styles of coding, I suppose.

nextcord/http.py
```diff
@@ -268,152 +663,250 @@ async def request(
...
-                                    retry_after,
+
+                            await global_rate_limit.update(response)
+                            try:
+                                await url_rate_limit.update(response)
+                            except IncorrectBucket as e:
+                                # This condition can be met when doing asyncio.gather()'d requests.
+                                if (
+                                    temp := self._buckets.get(
+                                        # The empty string default makes pyright happy. (hopefully)
```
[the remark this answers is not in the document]
alentoghostflame: That makes sense, updated the comment slightly.

nextcord/http.py
```diff
@@ -268,152 +663,250 @@ async def request(
...
+                                    url_rate_limit = temp
+                                    self._set_url_rate_limit(
+                                        route.method, route, auth, url_rate_limit
+                                    )
+                                    await url_rate_limit.update(response)
+                                else:
+                                    _log.debug(
+                                        "Route %s was given a different bucket, making a new one: %s",
+                                        rate_limit_path,
+                                        e,
```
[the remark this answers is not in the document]
alentoghostflame: Yes, but the error fully includes the full name of the old bucket and the migrated-to bucket, so I deemed it useful.

nextcord/http.py
```diff
@@ -268,152 +663,250 @@ async def request(
...
-                            _log.warning(fmt, retry_after, bucket)
-
-                            # check if it's a global rate limit
-                            if is_global:
-                                _log.warning(
-                                    "Global rate limit has been hit. Retrying in %.2f seconds.",
-                                    retry_after,
-                                )
-                                self._dispatch(
-                                    "global_http_ratelimit",
```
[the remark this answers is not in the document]
alentoghostflame: I must have missed that, added.

nextcord/http.py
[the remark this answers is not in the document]
EmreTech: Fair point. In that case, I think some of the important kwargs that should (hopefully) stay unchanged like `json` and `query` should be documented.

nextcord/http.py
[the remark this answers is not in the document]
EmreTech: Oh, I missed that part of the code. My bad.

nextcord/http.py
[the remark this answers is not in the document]
EmreTech: The format for the logged message doesn't say anything about the exception. It infers that only the name of the new bucket is given, not the entire exception message.

nextcord/http.py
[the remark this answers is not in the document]
alentoghostflame: okay so pyright is no longer upset about it, removed.

nextcord/http.py
```diff
@@ -155,25 +164,280 @@ def bucket(self) -> str:
...
-        traceback: Optional[TracebackType],
+        exc_type: Optional[Type[BaseException]],
+        exc_val: Optional[BaseException],
+        exc_tb: Optional[TracebackType],
     ) -> None:
-        if self._unlock:
-            self.lock.release()
+        self.release()
+
+    def locked(self) -> bool:
```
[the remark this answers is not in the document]
alentoghostflame: Did I not make this a property? Will fix when I get home.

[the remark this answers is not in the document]
EmreTech: You did make it a property in 3931e230fc0d9dc54489dc4f0ea7129e8c6ccc30.

nextcord/http.py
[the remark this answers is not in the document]
alentoghostflame: In hindsight, I'm dumb. Will change `_set_default_auth` to `set_default_auth`

nextcord/http.py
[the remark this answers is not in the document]
alentoghostflame: When webhooks get moved over to using HTTPClient, that would probably be a good time to expand on that.

nextcord/http.py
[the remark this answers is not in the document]
alentoghostflame: Fixing by creating the rate limit, updating it, THEN logging it with the rate limit name given.

nextcord/http.py
alentoghostflame: Updated it so a better, less generic error is raised.

EmreTech: That's not the problem I was addressing. I think this feature should be toggleable, and it should be off by default. Not everyone is going to be using `asyncio.gather()`, after all.

alentoghostflame: After a small talk on the Discord, it was decided that removing the denial system was a better idea than adding a toggle for it, with it toggled off by default.

nextcord/client.py
```diff
@@ -329,6 +328,8 @@ def __init__(
         self._ready: asyncio.Event = asyncio.Event()
         self._connection._get_websocket = self._get_websocket
         self._connection._get_client = lambda: self
+        self._token: str | None = None
```
Skelmis, line 331: As we still target a version lower then 3.10 this will need to change (I think, futures may fix it)
```suggestion
        self._token: Optional[str] = None
```
Note: Check optional is also imported

nextcord/errors.py
```diff
@@ -78,6 +80,10 @@ def _flatten_error_dict(d: Dict[str, Any], key: str = "") -> Dict[str, str]:
     return dict(items)
 
 
+class HTTPCancelled(DiscordException):
+    """Exception that's raised when an HTTP request is internally cancelled."""
```
Skelmis, line 84: Would just love maybe a `#` comment explaining if an end user should ever see this or not? More so my curiosity as to expectations here then anything

nextcord/http.py
```diff
@@ -122,25 +138,331 @@ def bucket(self) -> str:
...
+    """Minimum reset - current time diff threshold. Any diff lower will be ignored.
+    Only used when use_reset_timestamp is enabled.
+    """
+
+    def __init__(self, time_offset: float = 0.0, *, use_reset_timestamp: bool = False) -> None:
+        self.limit: int = 1
+        """Maximum amount of requests before requests have to wait for the rate limit to reset."""
+        self.remaining: int = 1
+        """Remaining amount of requests before requests have to wait for the rate limit to reset."""
+        self.reset: datetime | None = None
```
Skelmis, line 170: See earlier comments
```suggestion
        self.reset: Optional[datetime] = None
```

nextcord/http.py
```diff
@@ -122,25 +138,331 @@ def bucket(self) -> str:
...
+        """Datetime that the bucket roughly will be reset at."""
+        self._tracked_reset_time: float = 1.0
+        """The estimate time between bucket resets. Found via reset if use_reset_timestamp is True, found with
+        reset_after if False.
+        """
+        self.reset_after: float = 1.0
+        """Amount of seconds roughly until the rate limit will be reset."""
+        self.bucket: str | None = None
```
Skelmis, line 178: See earlier comment 
```suggestion
        self.bucket: Optional[str] = None
```

nextcord/http.py
```diff
@@ -122,25 +138,331 @@ def bucket(self) -> str:
...
+        """Name of the bucket, if it has one."""
+
+        self._time_offset: float = time_offset
+        """Number in seconds to increase all timers by. Used for lag compensation."""
+        self._use_reset_timestamp: bool = use_reset_timestamp
+        """If the reset timestamp should be used for bucket resets. If False, the float reset_after is used."""
+        self._first_update: bool = True
+        """If the next update to be ran will be the first."""
+        self._reset_remaining_task: asyncio.Task | None = None
```
Skelmis, line 187:
```suggestion
        self._reset_remaining_task: Optional[asyncio.Task] = None
```

nextcord/http.py
```diff
@@ -122,25 +138,331 @@ def bucket(self) -> str:
...
+        """Holds the task object for resetting the remaining count."""
+        self._on_reset_event: asyncio.Event = asyncio.Event()
+        """Used to indicate when the rate limit is ready to be acquired."""
+        self._migrating: str | None = None
```
Skelmis, line 191:
```suggestion
        self._migrating: Optional[str] = None
```

nextcord/http.py
```diff
@@ -122,25 +138,331 @@ def bucket(self) -> str:
...
+        """When this RateLimit is being deprecated and acquiring requests need to migrate to a different RateLimit, this
+        variable should be set to the different RateLimit/buckets string name.
+        """
+
+        self._on_reset_event.set()
+
+    @property
+    def resetting(self) -> bool:
+        return self._reset_remaining_task is not None and not self._reset_remaining_task.done()
```
Skelmis, line 200: Can we document this? Tad confused as to when it fires

nextcord/http.py
```diff
@@ -122,25 +138,331 @@ def bucket(self) -> str:
...
+            self.reset_after = x_reset_after
+            # Once we figure out what the true reset delay is, for example 5 seconds, we want to keep it at that.
+            if not self._use_reset_timestamp and self._tracked_reset_time < x_reset_after:
+                self._tracked_reset_time = x_reset_after
+                _log.debug(
+                    "Bucket %s: Reset after time increased to %s seconds, adapting reset time.",
+                    self.bucket,
+                    self._tracked_reset_time,
+                )
+                self.start_reset_task()
```
Skelmis, line 277: Given from my understanding this is adaptive to possible future, changes seems fine as is?

nextcord/http.py
```diff
@@ -149,83 +471,234 @@ def __exit__(
...
         proxy_auth: Optional[aiohttp.BasicAuth] = None,
-        loop: Optional[asyncio.AbstractEventLoop] = None,
-        unsync_clock: bool = True,
         dispatch: Callable,
     ) -> None:
-        self.loop: asyncio.AbstractEventLoop = asyncio.get_event_loop() if loop is None else loop
-        self.connector = connector
+        # TODO: Think about adding ratelimit_multiplier? Would reduce the internal RateLimit.limit by that
+        #  float (such as 0.7) and could allow people to run multiple NC bots/processes on the same token while avoiding
+        #  ratelimit issues. Could also help with replit-style scenarios.
```
Skelmis, line 532: Note with this approach, it assumes each process makes the seem amount of requests towards a bucket. However the following conditions are present:
- Seperate processes are unlikely to contribute equally
- Based on my ratelimit understanding and multiple processes, outside of the global ratelimit a single process shouldn't be sharing ratelimits with another process due to how sharding currently works

While I love the feature idea, a simple number is likely a bit too naive to get things right nicely and the complexity required to do it nicely would be too large imo

nextcord/http.py
```diff
@@ -149,83 +471,234 @@ def __exit__(
...
 
         user_agent = "DiscordBot (https://github.com/nextcord/nextcord/ {0}) Python/{1[0]}.{1[1]} aiohttp/{2}"
-        self.user_agent: str = user_agent.format(__version__, sys.version_info, aiohttp.__version__)
+        self._user_agent: str = user_agent.format(
+            __version__, sys.version_info, aiohttp.__version__
+        )
+
+        self._buckets: dict[str, RateLimit] = {}
+        """{"Discord bucket name": RateLimit}"""
+        self._global_rate_limits: dict[str | None, RateLimit] = {}
```
Skelmis, line 552:
```suggestion
        self._global_rate_limits: dict[Optional[str], RateLimit] = {}
```

nextcord/http.py
```diff
@@ -149,83 +471,234 @@ def __exit__(
...
+        """{"Auth string": RateLimit}, None for auth-less ratelimit."""
+        self._url_rate_limits: dict[tuple[str, str, str | None], RateLimit] = {}
```
Skelmis, line 554:
```suggestion
        self._url_rate_limits: dict[tuple[str, str, Optional[str]], RateLimit] = {}
```

nextcord/http.py
```diff
@@ -235,151 +708,266 @@ async def request(
...
+                        "error and should be reported!"
+                    ) from e
+
+                url_rate_limit = self._buckets.get(url_rate_limit.migrating)
+                if url_rate_limit is None:
+                    # This means we have an internal issue that we need to fix.
+                    raise ValueError(
+                        "RateLimit said to migrate, but the RateLimit to migrate was not found? This is an "
+                        "internal Nextcord error and should be reported!"
+                    ) from e
```
Skelmis, line 951: Plausibly we also want to do a `_log.error` within this statement with as much context as the library has regarding the situation to help us debug? Cos let's be real, it'll happen sooner or later

Skelmis (changes requested):
If futures imports resolve the union comments feel free to resolve them. I've also failed to highlight all instances of the pipe being used so you'll have to go through yourself and find them haha.

I've been running this PR in prod for a few months now. I've updated to latest as of now and come next review will see how it continues to hold up.


Consider this an approval outside of the few comments I've made, good stuff honestly 💯

nextcord/errors.py
alentoghostflame: As it stands right now, the end user can see it. Currently it's raised when someone does a request with `retry_request=False` and the ratelimit is already capped out, but I can make a more specific error that subclasses `HTTPCancelled` if that's desired.

nextcord/http.py
alentoghostflame: that's fair, I'll probably remove that TODO when I get home.

nextcord/http.py
alentoghostflame: Agreed. Removing TODO.

nextcord/errors.py
Skelmis: My thinking here is just that as an end user, seeing an error with "internally cancelled" doesn't really tell me if I need to handle the error, or if its fine to ignore if that makes sense?

Maybe just another line saying essentially what you said would be nice. Something like: "... This exception indicates that a request was made to a route which exceeded the ratelimited. As a result, the request was cancelled."

alentoghostflame: Made some changes and added a new exception

nextcord/http.py
```diff
@@ -124,25 +140,329 @@ def bucket(self) -> str:
...
+            # The response is intended for the regular rate limit, not a global rate limit.
+            return
+
+        if response.status == 429:
+            # Oh dear, we hit the rate limit.
+            _log.warning("Global rate limit 429 encountered, setting remaining to 0.")
+            self.remaining = 0
+            if response.headers.get("X-RateLimit-Scope") == "global":
+                data = await response.json()
+                _log.debug(data)
```
Skelmis, line 448: This is a sizeable slowdown for bots without logging enabled at the debug level. I assume this is a log leftover from development?

If it's not, please port it to a lazy logging setup
```suggestion
                _log.debug("%s", data)
```

nextcord/http.py
```diff
@@ -151,48 +471,171 @@ def __exit__(
...
+    """
 
     def __init__(
         self,
         connector: Optional[aiohttp.BaseConnector] = None,
         *,
+        max_global_requests: int = 50,
+        time_offset: float = 0.0,
+        default_auth: Optional[str] = None,
+        assume_unsync_clock: bool = False,  # TODO: Think about renaming this?
```
Skelmis, line 525: Is there a resolution to this todo? Seems fine to me given its documented well

nextcord/http.py
```diff
@@ -151,48 +471,171 @@ def __exit__(
...
+        _log.debug(
+            "Creating global ratelimit for auth %s with max per second %s.",
+            _get_logging_auth(auth),
+            max_per_second,
+        )
+        rate_limit = GlobalRateLimit(time_offset=self._time_offset)
+        rate_limit.limit = max_per_second
+        rate_limit.remaining = max_per_second
+        rate_limit.reset_after = 1 + self._time_offset
+        rate_limit.bucket = f"Global {_get_logging_auth(auth) if auth else 'Unauthorized'}"
```
Skelmis, line 562: TBF not fussed on this, feel free to resolve
```suggestion
        log_auth = _get_logging_auth(auth)
        _log.debug(
            "Creating global ratelimit for auth %s with max per second %s.",
            log_auth,
            max_per_second,
        )
        rate_limit = GlobalRateLimit(time_offset=self._time_offset)
        rate_limit.limit = max_per_second
        rate_limit.remaining = max_per_second
        rate_limit.reset_after = 1 + self._time_offset
        rate_limit.bucket = f"Global {log_auth if auth else 'Unauthorized'}"
```

nextcord/http.py
```diff
@@ -404,55 +1001,125 @@ async def close(self) -> None:
...
+        client_id: int,
+        client_secret: str,
+        code: str,
+        redirect_uri: str,
+        auth: Optional[str] = MISSING,
+        retry_request: bool = True,
+    ):
+        # TODO: Look into how viable this function is here.
+        # This doesn't actually have hard ratelimits it seems? Not in the headers at least. The default bucket should
+        #  keep it at 1 every 1 second.
```
Skelmis, line 1030: Any resolution to this?

Skelmis (changes requested):
Been running the version from last review with no issues, assuming you've also tested all the minor method changes yourself (which i havent) I'm keen as for this outside of the minor stuff raised here

nextcord/http.py
alentoghostflame: Not sure. On one hand it doesn't seem to have public ratelimits, on the other this IS the HTTP handler class and thus should handle HTTP requests to Discord. I think I'm going to keep it, having it at the Nextcord default of 1/1s seems fine to me.

nextcord/http.py
alentoghostflame: It was left over at development. I was going to remove it, but hitting the global rate limit is rare and debug logs in such a case would likely be useful.

nextcord/http.py
alentoghostflame: Fair, removed.

nextcord/errors.py
```diff
@@ -78,6 +81,17 @@ def _flatten_error_dict(d: Dict[str, Any], key: str = "") -> Dict[str, str]:
     return dict(items)
 
 
+class HTTPInternalCancelled(DiscordException):
```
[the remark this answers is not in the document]
Skelmis: Fair nuff

nextcord/http.py
```diff
@@ -124,25 +140,329 @@ def bucket(self) -> str:
...
+            self.remaining = int(x_remaining)
+        else:
+            # If requests come back out of order, it's possible that we could get a wrong amount remaining.
+            # It's best to be pessimistic and assume it cannot go back up unless the reset task occurs.
+            self.remaining = (
+                int(x_remaining) if int(x_remaining) < self.remaining else self.remaining
+            )
+
+        # Updates the datetime of the reset.
+        x_reset = response.headers.get("X-RateLimit-Reset")
```
TAG-Epic, line 244: I would love to see leaky buckets on this as well via Retry-After.

nextcord/http.py
```diff
@@ -124,25 +140,329 @@ def bucket(self) -> str:
...
+        if x_reset is not None:
+            x_reset = datetime.fromtimestamp(float(x_reset) + self._time_offset, tz=timezone.utc)
```
TAG-Epic, line 246: time.time is always in UTC which is a lot more readable than using datetimes and getting to weird dates close to epoch.

nextcord/http.py
```diff
@@ -124,25 +140,329 @@ def bucket(self) -> str:
...
+            "Bucket %s: Updated with limit %s, remaining %s, reset %s, and reset_after %s seconds.",
+            self.bucket,
+            self.limit,
+            self.remaining,
+            self.reset,
+            self.reset_after,
+        )
+
+    def start_reset_task(self) -> None:
+        """Starts the reset task, non-blocking."""
```
TAG-Epic, line 305: This should be private.
Also no need to comment it is non blocking? Document just the blocking parts imo

nextcord/http.py
```diff
@@ -124,25 +140,329 @@ def bucket(self) -> str:
...
+                    seconds_until_reset = reset_delta
+
+            else:
+                seconds_until_reset = self._tracked_reset_time
+
+        else:  # In reset_after seconds mode.
+            seconds_until_reset = self._tracked_reset_time
+
+        loop = asyncio.get_running_loop()
+        self._reset_remaining_task = loop.create_task(self.reset_remaining(seconds_until_reset))
```
TAG-Epic, line 328: why not use asyncio.create_task?

nextcord/http.py
```diff
@@ -124,25 +140,329 @@ def bucket(self) -> str:
...
+
+    async def reset_remaining(self, time: float) -> None:
```
TAG-Epic, line 330: Should be private?

nextcord/http.py
```diff
@@ -124,25 +140,329 @@ def bucket(self) -> str:
...
+                    self.limit,
+                )
+                self._on_reset_event.clear()
+                if not self.resetting:
+                    self.start_reset_task()
+
+        if self.migrating:
+            raise RateLimitMigrating(
+                f"This RateLimit is deprecated, you need to migrate to bucket {self.migrating}"
+            )
```
TAG-Epic, line 405: Now I know its easy to get via RateLimit.migrating, but would make sense to have the new bucket hash on the error?

nextcord/http.py
```diff
@@ -124,25 +140,329 @@ def bucket(self) -> str:
...
     ) -> None:
-        if self._unlock:
-            self.lock.release()
+        self.release()
+
+    @property
+    def locked(self) -> bool:
+        return self.remaining <= 0
+
+    async def acquire(self) -> bool:
```
TAG-Epic, line 375: Requests which are in-progress don't count towards the rate limit. This may cause issues when the request is close to the reset time

nextcord/http.py
```diff
@@ -124,25 +140,329 @@ def bucket(self) -> str:
...
+        _log.debug("Bucket %s: Continuing with request.", self.bucket)
+        self.remaining -= 1
+        return True
+
+    def release(self) -> None:
+        # Basically a placeholder, could probably be removed ;)
+        pass
+
+
+class GlobalRateLimit(RateLimit):
```
TAG-Epic, line 416: These imo are too different to be subclasses of eachother.

nextcord/http.py
```diff
@@ -151,48 +471,172 @@ def __exit__(
...
+
+    Parameters
+    ----------
+    connector: Optional[:class:`aiohttp.BaseConnector`]
+        The connector object to use for the client session.
+    max_global_requests: :class:`int`
+        Maximum amount of requests per second per authorization.
+
+        Discord by default only allows 50 requests per second, but if your bot has had its maximum increased, then
+        increase this parameter.
```
TAG-Epic, line 486: Since you have access to the gateway parts, why not make this estimate based on the guild/shard count?
Might not be accurate, but worst that can happen is they get less perf than what they could if they aren't manually setting it.

nextcord/http.py
```diff
@@ -151,48 +471,172 @@ def __exit__(
...
+        Amount of seconds added to all ratelimit timers for lag compensation.
+
+        Due to latency and Discord servers not perfectly time synced, having no offset can cause 429's to occur even
+        with us following the reported X-RateLimit-Reset-After.
+
+        Increasing will protect from erroneous 429s but will slow bucket resets, lowering max theoretical speed.
+
+        Decreasing will hasten bucket resets and increase max theoretical speed but may cause 429s.
+    default_auth: Optional[:class:`str`]
+        Default string to use in the Authorization header if it's not manually provided.
```
TAG-Epic, line 497: dpy only has one bot per http client iirc and has seperate RL implementations for webhooks? 
Why is this not required?

nextcord/http.py
```diff
@@ -151,48 +471,172 @@ def __exit__(
...
+        If ``False``, rate limits will use a combination of server-given bucket reset timestamps compared with local
+        time to accurately reset when the bucket resets and discovered static-ish time delays to predict the next reset
+        when needed.
+        Generally results in more accurate bucket reset times, especially with high-latency connections, but relies
+        heavily on host time being accurate with Discords time.
+    proxy: Optional[:class:`str`]
+        The proxy url to connect to the Discord API with, if any.
+    proxy_auth: Optional[:class:`aiohttp.BasicAuth`]
+        The authentication to use in order to make a request to the proxy url. Not to be confused with ``default_auth``.
+    dispatch: :class:`Callable`
```
TAG-Epic, line 514: Generics

nextcord/http.py
```diff
@@ -151,48 +471,172 @@ def __exit__(
...
+
+        Due to latency and Discord servers not perfectly time synced, having no offset can cause 429's to occur even
+        with us following the reported X-RateLimit-Reset-After.
+
+        Increasing will protect from erroneous 429s but will slow bucket resets, lowering max theoretical speed.
+
+        Decreasing will hasten bucket resets and increase max theoretical speed but may cause 429s.
+    default_auth: Optional[:class:`str`]
+        Default string to use in the Authorization header if it's not manually provided.
+    assume_unsync_clock: :class:`bool`
```
TAG-Epic, line 498: Is there even any logic for this in the rate limiters? Can't find any

nextcord/http.py
```diff
@@ -151,48 +471,172 @@ def __exit__(
...
+        # For consistency with possible future changes to set_default_auth.
+        self.set_default_auth(default_auth)
+        self._ratelimit_use_timestamp = not assume_unsync_clock
+        self._proxy = proxy
+        self._proxy_auth: Optional[aiohttp.BasicAuth] = proxy_auth
+        self._dispatch = dispatch
 
         # to mitigate breaking changes
-        self.user_agent: str = _USER_AGENT
+        self._user_agent: str = _USER_AGENT
```
TAG-Epic, line 543: If it is to mitigate breaking changes maybe don't change the name?

nextcord/http.py
```diff
@@ -151,48 +471,172 @@ def __exit__(
...
+        log_auth = _get_logging_auth(auth)
+        _log.debug(
+            "Creating global ratelimit for auth %s with max per second %s.",
+            log_auth,
+            max_per_second,
+        )
+        rate_limit = GlobalRateLimit(time_offset=self._time_offset)
+        rate_limit.limit = max_per_second
+        rate_limit.remaining = max_per_second
+        rate_limit.reset_after = 1 + self._time_offset
```
TAG-Epic, line 562: Isn't this double added?

nextcord/http.py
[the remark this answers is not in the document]
TAG-Epic: This needs to be somewhat documented (for example the json param) at least.

nextcord/http.py
```diff
@@ -240,151 +707,280 @@ async def request(
...
+        if (url_rate_limit := self._get_url_rate_limit(route.method, route, auth)) is None:
+            url_rate_limit = self._make_url_rate_limit(route.method, route, auth)
 
+        max_retry_count = 5
+        rate_limit_path = (
+            route.method,
+            route.bucket,
+            _get_logging_auth(auth),
+        )  # Only use this for logging.
+        ret: Union[Optional[str], dict] = None
```
TAG-Epic, line 726: Generics. Also please make the variable name sensible.

nextcord/http.py
```diff
@@ -240,151 +707,280 @@ async def request(
...
-                    form_data = aiohttp.FormData(quote_fields=False)
-                    for params in form:
-                        form_data.add_field(**params)
-                    kwargs["data"] = form_data
-
-                try:
-                    async with self.__session.request(method, url, **kwargs) as response:
+
+        # If retry_request is False and any of the rate limits are locked, don't continue and raise immediately.
+        if retry_request is False:
```
TAG-Epic, line 730: I feel like this is unexpected behavior. Waiting for the rate limit to reset isn't retrying the request?

nextcord/http.py
```diff
@@ -240,151 +707,280 @@ async def request(
...
+                    "Request would exceed the route ratelimit and retries are disabled."
+                )
+
+        # The loop is to allow migration to a different RateLimit if needed.
+        # If we hit this loop max_retry_count times, something is wrong. Either we're migrating buckets way
+        #  too much, 429s keep getting hit, or something is internally wrong.
+        for retry_count in range(max_retry_count):  # To prevent infinite loops.
+            should_retry = False
+            try:
+                async with global_rate_limit, url_rate_limit:
```
TAG-Epic, line 755: This causes slowdowns.
Quick fix: swap them around
Long fix: wait for both to be ready before acquiring. Will happily send you a example of how to do this from my rate limit library.

nextcord/http.py
```diff
@@ -240,151 +707,280 @@ async def request(
...
+                    # This check is for asyncio.gather()'d requests where the rate limit can change.
+                    if (
+                        temp := self._get_url_rate_limit(route.method, route, auth)
```
TAG-Epic, line 758: var name is awful

nextcord/http.py
```diff
@@ -240,151 +707,280 @@ async def request(
...
+                        _log.debug(
+                            "Route %s had the rate limit changed, resetting and retrying.",
+                            rate_limit_path,
+                        )
+                        url_rate_limit = temp
+                        continue
+
+                    if files:
+                        for f in files:
+                            f.reset(seek=retry_count)
```
TAG-Epic, line 769: Wouldn't make this so files are offset by one byte for every retry?

nextcord/http.py
```diff
@@ -240,151 +707,280 @@ async def request(
...
+
+                raise
+
+            except RateLimitMigrating as e:
+                if url_rate_limit.migrating is None:
+                    raise ValueError(
+                        "RateLimitMigrating raised, but RateLimit.migrating is None. This is an internal Nextcord "
+                        "error and should be reported!"
+                    ) from e
+
```
TAG-Epic, line 945: An assert sounds better here?

nextcord/http.py
```diff
@@ -404,55 +1000,124 @@ async def close(self) -> None:
...
+        *,
+        client_id: int,
+        client_secret: str,
+        code: str,
+        redirect_uri: str,
+        auth: Optional[str] = MISSING,
+        retry_request: bool = True,
+    ):
+        # This doesn't actually have hard ratelimits it seems? Not in the headers at least. The default bucket should
+        #  keep it at 1 every 1 second.
```
TAG-Epic, line 1028: This seems really slow. Maybe keep it at one at a time? Or hard-code oauth endpoints to not have RLs?

TAG-Epic (changes requested):
Since this is breaking changes I assume this is gonna be in next major release?

nextcord/http.py
alentoghostflame: idk, I just took it from the original HTTPClient code.

nextcord/http.py
```diff
@@ -151,48 +471,172 @@ def __exit__(
...
         # to mitigate breaking changes
-        self.user_agent: str = _USER_AGENT
+        self._user_agent: str = _USER_AGENT
+
+        self._buckets: dict[str, RateLimit] = {}
+        """{"Discord bucket name": RateLimit}"""
+        self._global_rate_limits: dict[Optional[str], RateLimit] = {}
+        """{"Auth string": RateLimit}, None for auth-less ratelimit."""
+        self._url_rate_limits: dict[tuple[str, str, Optional[str]], RateLimit] = {}
+        """{("METHOD", "Route.bucket", "auth string"): RateLimit} auth string may be None to indicate auth-less."""
```
TAG-Epic, line 550: Mem leak? I don't think buckets and url rate limits ever gets cleared?

TAG-Epic (changes requested):

nextcord/http.py
EmreTech: That comment was apart of #1114, unrelated to this PR.

nextcord/http.py
alentoghostflame: I'm going to leave that for the OAuth PR.

nextcord/errors.py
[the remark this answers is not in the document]
alentoghostflame: They are exposed to the user, specifically when `retry_request=False` currently.

nextcord/http.py
```diff
@@ -124,25 +140,329 @@ def bucket(self) -> str:
...
+        self.bucket: Optional[str] = None
+        """Name of the bucket, if it has one."""
+
+        self._time_offset: float = time_offset
+        """Number in seconds to increase all timers by. Used for lag compensation."""
+        self._use_reset_timestamp: bool = use_reset_timestamp
+        """If the reset timestamp should be used for bucket resets. If False, the float reset_after is used."""
+        self._first_update: bool = True
+        """If the next update to be ran will be the first."""
+        self._reset_remaining_task: Optional[asyncio.Task] = None
```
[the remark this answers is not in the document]
EmreTech: `Task` is generic when used in type hints, basically meaning you need to use brackets to give more information about the task objects.
For example, `Task[None]` tells you that this is a task that will have the return type of `None`.

nextcord/http.py
```diff
@@ -92,13 +98,23 @@ async def json_or_text(response: aiohttp.ClientResponse) -> Union[Dict[str, Any]
...
+def _get_logging_auth(auth: Optional[str]) -> str:
+    if auth is None:
+        return "None"
+
+    if len(auth) < 12:  # This shouldn't ever occur, but whatever.
+        return "[redacted]"
+
+    return f"{auth[:12]}[redacted]"
+
+
```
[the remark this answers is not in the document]
ooliver1: I might recommend the last x characters instead of the first x, as the first portion of a bot token is the same for the same application.

nextcord/http.py
```diff
@@ -124,25 +140,329 @@ def bucket(self) -> str:
...
+        if x_remaining is None:
+            self.remaining = 1
+        elif self._first_update:
+            self.remaining = int(x_remaining)
+        else:
+            # If requests come back out of order, it's possible that we could get a wrong amount remaining.
+            # It's best to be pessimistic and assume it cannot go back up unless the reset task occurs.
+            self.remaining = (
+                int(x_remaining) if int(x_remaining) < self.remaining else self.remaining
+            )
```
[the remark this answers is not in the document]
teaishealthy:
```suggestion
            self.remaining = min(int(x_remaining), self.remaining)
```

nextcord/http.py
```diff
@@ -151,48 +476,172 @@ def __exit__(
...
     def recreate(self) -> None:
         if self.__session.closed:
             self.__session = aiohttp.ClientSession(
-                connector=self.connector,
+                connector=self._connector,
                 ws_response_class=DiscordClientWebSocketResponse,
-                headers={
-                    "User-Agent": _USER_AGENT,
-                },
             )
```
teaishealthy, line 639: Is this safe? (I know it's not your code)

nextcord/http.py
teaishealthy:
```suggestion
                            f.reset(seek=bool(retry_count))
```

Who came up with this 😭

The value of seek isn't interesting, it's whether it's truthy (≠0) or falsy (=0).
→ If we are retrying/retried, we seek back to the original position.

nextcord/http.py
```diff
@@ -240,151 +712,280 @@ async def request(
...
-                        remaining = response.headers.get("X-Ratelimit-Remaining", "")
+                        await global_rate_limit.update(response)
+                        try:
+                            await url_rate_limit.update(response)
+                        except IncorrectBucket:
+                            # This condition can be met when doing asyncio.gather()'d requests.
+                            if (
+                                temp := self._buckets.get(
+                                    # Defaulting to "" makes pyright happy because None is an invalid type of key.
+                                    response.headers.get("X-RateLimit-Bucket", "")
```
teaishealthy, line 806: cast to str or Any

teaishealthy (changes requested):

nextcord/http.py
```diff
@@ -204,33 +653,56 @@ async def ws_connect(self, url: str, *, compress: int = 0) -> Any:
...
+
+        Parameters
+        ----------
+        route: :class:`Route`
+            The Discord Route to make the API request for.
+        files: Optional[Sequence[:class:`File`]]
+            The files to provide with the JSON body of this request. If provided, the body of the request will be
+            repackaged into a multipart form.
+        form: Optional[Iterable[:class:`dict`[:class:`str`, `Any`]]]
+            pass
```
EmreTech, line 681: Remember to document this parameter

nextcord/http.py
```diff
@@ -204,33 +653,56 @@ async def ws_connect(self, url: str, *, compress: int = 0) -> Any:
...
+        auth: :class:`str` | `None` | `MISSING`
```
EmreTech, line 682:
```suggestion
        auth: :class:`str` | `None`
```
`MISSING` shouldn't be included in the type hint since it's not really a type (for now).

nextcord/http.py
EmreTech: It's not, it partially reverts #1114 (and thus #1002 becomes an issue again).

nextcord/http.py
```diff
@@ -204,33 +653,56 @@ async def ws_connect(self, url: str, *, compress: int = 0) -> Any:
 
         return await self.__session.ws_connect(url, **kwargs)
 
+    # ruff: noqa: PLR0915
+    # Ruff says this is too many statements in a single function, but I'm not sure how to break it down while
```
EmreTech, line 657: You could remove the parts of the code that make the actual HTTP request & handle the HTTP errors into an internal function then have this request function just be dedicated to handling ratelimits.

nextcord/http.py
alentoghostflame: Currently I have the user agent stuff handled in `HTTPClient._make_headers`, but I can add it back into the `ClientSession` itself if desired.

teaishealthy: I was more concerned about the creation of a ClientSession potentionally without a running loop / without an async context. (bot crashed, user calls recreate with a closed loop, ...)

nextcord/http.py
teaishealthy: Also this shouldn't be a 3.10 union.

nextcord/http.py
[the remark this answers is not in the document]
alentoghostflame: Dang it, forgot `min` existed. Adding that.

nextcord/http.py
alentoghostflame: Fixing, and making `Client.clear()` an async function as well.

nextcord/http.py
alentoghostflame: I'm not sure how to do that?

teaishealthy:
```suggestion
                                    cast(Any, response.headers.get("X-RateLimit-Bucket"))
```

make sure typing.Any and typing.Cast have been imported

EmreTech: What's the purpose of casting to Any? I think the current solution works fine, since it just defaults to an empty string instead of trying to trick the type checker.

nextcord/http.py
alentoghostflame: To be honest, I have little idea what a form is supposed to be.

nextcord/http.py
alentoghostflame: ok.

nextcord/http.py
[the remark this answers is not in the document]
alentoghostflame: Duh, fixing.

nextcord/http.py
```diff
@@ -45,9 +47,7 @@
 _log = logging.getLogger(__name__)
 
 if TYPE_CHECKING:
-    from types import TracebackType
-
-    from typing_extensions import Self
+    pass
```
teaishealthy:
```suggestion
```

nextcord/http.py
```diff
@@ -130,6 +130,15 @@ def _modify_api_version(version: Literal[9, 10]):
     Route.BASE = f"https://discord.com/api/v{version}"
 
 
+def _get_logging_auth(auth: str | None) -> str:
+    if auth is None:
+        return "None"
+    elif len(auth) < 12:  # This shouldn't ever occur, but whatever.
+        return "[redacted]"
+    else:
+        return f"{auth[:12]}[redacted]"
```
teaishealthy:
```suggestion
    else:
        return f"{auth[:12]}[redacted]"
```

out of bounds string slicing returns an empty string:
```py
>>> "hello"[:12]
'hello'
>>> ""[:12]
''
```

teaishealthy (dismissed):

docs/api.rst
```diff
@@ -3465,6 +3469,8 @@ Exception Hierarchy
                 - :exc:`Forbidden`
```
EmreTech, line 3469:
```suggestion
                - :exc:`Unauthorized`
                - :exc:`Forbidden`
```

nextcord/http.py
```diff
@@ -240,151 +705,326 @@ async def request(
...
+        retry_request: bool,
+        retry_count: int,
+        url_rate_limit: RateLimit,
+        global_rate_limit: GlobalRateLimit,
+    ) -> bool:
+        """|coro|
+        This exists to make the main request function smaller and easier to work with.
+
+        I'm on a 1440p 21:9 monitor and even with PyCharm split into multiple panes, the request method with all of
+        this in it is absolutely huge.
```
EmreTech, line 931: I think this would be better off as a comment above the function instead of in the documentation.
```suggestion

```

nextcord/http.py
```diff
@@ -151,48 +472,172 @@ def __exit__(
 
 
 class HTTPClient:
```
EmreTech, line 474: Will this be in the actual documentation? If so, remember to add it to `api.rst`.

EmreTech (commented):
Few more suggestions left, sorry 😭 
It's all documentation stuff at this point tho

nextcord/http.py
alentoghostflame: ...not at this time. Once Nextcord/Client/Bot becomes less of a single monolithic structure and more build-a-bot with separation between the various parts sure, but until then I'd like to keep it off the docs.

ooliver1 (approved):
Tested by others is good :+1:, merging since 3.0 is in development anyway and this is a large change.

PR #1071, opened 2023-06-17
refactor: update to pyright 1.1.317

ooliver1:
## Summary

Update to [Pyright 1.1.315](https://github.com/microsoft/pyright/releases/tag/1.1.315)

## Checklist

### DIFF
--- a/docs/extensions/builder.py
+++ b/docs/extensions/builder.py
@@ -26,9 +26,6 @@ class DPYStandaloneHTMLBuilder(StandaloneHTMLBuilder):
     def write_genindex(self) -> None:
         # the total count of lines for each index letter, used to distribute
         # the entries into two columns
-        if self.env is None:
-            raise ValueError("No environment was found.")
-
         genindex = IndexEntries(self.env).create_index(self, group_entries=False)
         indexcounts = []
         for _k, entries in genindex:
--- a/docs/requirements.txt
+++ b/docs/requirements.txt
@@ -1,4 +1,4 @@
-sphinx==5.0.1
+sphinx==5.2.3
 sphinxcontrib_trio==1.1.2
 sphinxcontrib-websupport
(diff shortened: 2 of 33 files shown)


Skelmis (commented):
Are all the guard removals due to pyright saying it *should* be fine?

ooliver1:
Yes, I left some in where I was not sure why they would even exist in the first place (audit logs for example).

Skelmis:
> Yes, I left some in where I was not sure why they would even exist in the first place (audit logs for example).

However in some cases is that not just because you `cast` the line before the check? If the check were before the `cast` then pyright would likely still like it? Or is it also a case of "api should never send this"

ooliver1:
I assume you're referring to this?
```py
channel_id = cast(Optional[Union[str, int]], data["channel_id"])
``
Voice state updates for the bot itself happen to have `channel_id` nullable, but setting it in the underlying type means that has to be checked everywhere a user voice state is referenced. I could change it to do that but then other assertions/casts would be done elsewhere

**Comment by Skelmis** [comment]
Okay, cool as

**Review by Skelmis — APPROVED** [review]
Seems fine, couple areas behavior may change but I am assuming you've looked into those to ensure it's consistent with the API

File: docs/extensions/builder.py [hunk 1]
```diff
@@ -27,6 +27,7 @@ def write_genindex(self) -> None:
         # the total count of lines for each index letter, used to distribute
         # the entries into two columns
         if self.env is None:
```
teaishealthy, line 29: This is where `Builder`s are created and env is not guaranteed to be BuildEnvironment (see typehints). https://github.com/sphinx-doc/sphinx/blob/master/sphinx/registry.py#L157

But in the flow of a Sphinx app this shouldn't be None ever?

teaishealthy (approved):
🦅

PR #1067, opened 2023-06-16
feat(guild): add premium_progress_bar_enabled

ooliver1:
## Summary

discord/discord-api-docs#4001

## Checklist

- [x] I have tested my changes.
- [x] I have updated the documentation to reflect the changes.
- [x] I have run `task pyright` and fixed the relevant issues.

PR #1074, opened 2023-06-26
fix(guild): premium_progress_bar_enabled can be null

spifory:
## Summary

The documentation specifies that this field will always be avaliable, but apparently not.
This PR no longer uses key access to get the field in case it happens to be null when there is an unavaliable guild payload.

Continuation of #1067

## Checklist

- [x] This PR fixes an issue.

### DIFF
--- a/nextcord/guild.py
+++ b/nextcord/guild.py
@@ -516,7 +516,7 @@ class Guild(Hashable):
         for event in guild.get("guild_scheduled_events") or []:
             self._store_scheduled_event(event)
 
-        self._premium_progress_bar_enabled: bool = guild["premium_progress_bar_enabled"]
+        self._premium_progress_bar_enabled: Optional[bool] = guild.get("premium_progress_bar_enabled")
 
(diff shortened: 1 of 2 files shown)


Skelmis (approved):

nextcord/guild.py
```diff
@@ -654,7 +656,7 @@ def scheduled_events(self) -> List[ScheduledEvent]:
         return list(self._scheduled_events.values())
 
     @property
-    def premium_progress_bar_enabled(self) -> bool:
+    def premium_progress_bar_enabled(self) -> Optional[bool]:
         """:class:`bool`: Whether the premium boost progress bar is enabled.
```
EmreTech, line 660:
```suggestion
        """Optional[:class:`bool`:] Whether the premium boost progress bar is enabled.
```

EmreTech (changes requested):

EmreTech (approved):

PR #1061, opened 2023-06-08
feat: implement new username system

spifory:
## Summary
This adds the new Discord [username](https://dis.gd/usernames) update.
Implements discord/discord-api-docs#6218 discord/discord-api-docs#6130
Closes #1060

## Checklist

- [x] If code changes were made then they have been tested.
  - [x] I have updated the documentation to reflect the changes.
  - [x] I have run `task pyright` and fixed the relevant issues.
- [ ] This PR fixes an issue.
- [x] This PR adds something new (e.g. new method or parameters).
- [x] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

PR #1096, opened 2023-07-20
refactor(user): use key access for `BaseUser.discriminator` 

spifory:
## Summary

This makes `BaseUser.discriminator` use key access instead of `.get`.

The issue was initially caused by #1061 where at first I had planned for the discriminator to be None, but forgot it would be a breaking change.

## This is a **Code Change**


EmreTech (approved):
Doesn't actually fix typing, just more of a good practice. The typing was already correct.

teaishealthy (approved):

PR #1099, opened 2023-07-26
feat(utils): add format_ts

### LINKED ISSUE #1085: Let nextcord.utils.format_dt also take a timestamp instead of only datetime
### Summary

Let the `format_dt` function from `nextcord.utils` also be able to work with unix timestamps 

The core library

### The Problem

I have to either manually format them or convert something to a python `datetime` object just for the sake of using the `format_dt` function

### The Ideal Solution

Let `nextcord.utils.format_dt(...)` also take unix timestamps instead of only `datetime` objects, especially when the `format_dt` function just calls `.timestamp()` on the supplied `datetime` object anyways

### The Current Solution

Manually formatting it when not converting it to a `datetime` object just to be able to use this function

### Additional Context

_No response_

teaishealthy:
## Summary

Create a new function `format_ts` that can format a unix timestamp, format_dt can use this function too to reduce code duplication.

Closes #1085 

## This is a **Code Change**

### DIFF
--- a/nextcord/utils.py
+++ b/nextcord/utils.py
@@ -97,6 +97,7 @@ __all__ = (
     "parse_raw_channel_mentions",
     "as_chunks",
     "format_dt",
+    "format_ts",
 )
 
 DISCORD_EPOCH = 1420070400000
--- a/nextcord/utils.py
+++ b/nextcord/utils.py
@@ -1129,9 +1129,54 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None) 
     """
     if not isinstance(dt, datetime.datetime):  # pyright: ignore[reportUnnecessaryIsInstance]
         raise InvalidArgument("'dt' must be of type 'datetime.datetime'")
+    return format_ts(int(dt.timestamp()), style=style)
+
+
+def format_ts(ts: int, /, style: Optional[TimestampStyle] = None) -> str:
+    """A helper function to format an :class:`int` unix timestamp for presentation within Discord.
+
+    This allows for a locale-independent way of presenting data using Discord specific Markdown.
+
+    +-------------+----------------------------+-----------------+
+    |    Style    |       Example Output       |   Description   |
+    +=============+============================+=================+
+    | t           | 22:57                      | Short Time      |
+    +-------------+----------------------------+-----------------+
+    | T           | 22:57:58                   | Long Time       |
+    +-------------+----------------------------+-----------------+
+    | d           | 17/05/2016                 | Short Date      |
+    +-------------+----------------------------+-----------------+
+    | D           | 17 May 2016                | Long Date       |
+    +-------------+----------------------------+-----------------+
+    | f (default) | 17 May 2016 22:57          | Short Date Time |
+    +-------------+----------------------------+-----------------+
+    | F           | Tuesday, 17 May 2016 22:57 | Long Date Time  |
+    +-------------+----------------------------+-----------------+
+    | R           | 5 years ago                | Relative Time   |
+    +-------------+----------------------------+-----------------+
+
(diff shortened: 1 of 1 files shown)


nextcord/utils.py
```diff
@@ -1129,9 +1129,54 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None)
     """
     if not isinstance(dt, datetime.datetime):  # pyright: ignore[reportUnnecessaryIsInstance]
         raise InvalidArgument("'dt' must be of type 'datetime.datetime'")
+    return format_ts(int(dt.timestamp()), style=style)
```
spifory, line 1132: No need to re-int an integer
```suggestion
    return format_ts(dt.timestamp(), style=style)
```

nextcord/utils.py
```diff
@@ -1129,9 +1129,54 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None)
...
+
+
+def format_ts(ts: int, /, style: Optional[TimestampStyle] = None) -> str:
+    """A helper function to format an :class:`int` unix timestamp for presentation within Discord.
```
spifory, line 1136:
```suggestion
    """A helper function to format a Unix timestamp as an :class:`int` for presentation within Discord.
```

nextcord/utils.py
```diff
@@ -1129,9 +1129,54 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None)
...
+
+    Note that the exact output depends on the user's locale setting in the client. The example output
+    presented is using the ``en-GB`` locale.
+
+    .. versionadded:: 2.6
+
+    Parameters
+    ----------
+    ts: :class:`int`
+        The unix timestamp to format.
```
spifory, line 1166:
```suggestion
        The Unix timestamp to format.
```

nextcord/utils.py
```diff
@@ -1129,9 +1129,54 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None)
...
+    style: :class:`str`
+        The style to format the datetime with.
```
spifory, line 1168:
```suggestion
        The style to format the timestamp with.
```

nextcord/utils.py
```diff
@@ -1129,9 +1129,54 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None)
...
+
+    Returns
+    -------
+    :class:`str`
+        The formatted string.
+    """
+    if not isinstance(ts, int):  # pyright: ignore[reportUnnecessaryIsInstance]
```
spifory, line 1175: While I do understand that `format_dt` initially did this, _is_ this actually needed? `ts` is already typed as an integer so is there really a need to check if it's an integer a second time?

nextcord/utils.py
```diff
@@ -1129,9 +1129,54 @@ def format_dt(dt: datetime.datetime, /, style: Optional[TimestampStyle] = None)
     """
     if not isinstance(dt, datetime.datetime):  # pyright: ignore[reportUnnecessaryIsInstance]
```
spifory, line 1130: Same question about checking the type again

spifory (changes requested):
And also it's not too important but you may want to add a `versionchanged` to `format_dt`.

nextcord/utils.py
teaishealthy: Its a float

teaishealthy:
> And also it's not too important but you may want to add a `versionchanged` to `format_dt`.
@spifory 
The external api for format_dt did not change, no reason to.

EmreTech (approved):

spifory (approved):

ooliver1 (approved):

PR #1113, opened 2023-08-17
feat: voice recording

TrapDrap:
## Summary

This PR aims to implement voice recording to nextcord.
This is mostly fresh code, with some decoding & decrypting parts taken from the previous dead PR on voice recording.
Note that this is still a work-in-progress, it is not yet fully finalized, most features are working but may be unstable, and some options may be missing.
Next steps would be:
- Timings calculations (done) ✅
- Make stop_recording method (done) ✅
- Create export methods & ffmpeg stuff (done) ✅
- Refine cache cleanups (done) ✅
- Examples (mostly done, requires some refining) ✔🧹
- Documentation (mostly done) ✔🧹
- Audio Merging implementation (done, needs to be moved) ✔🧹
- Potential single-track direct recording support? (have an unstable prototype) 🚧

<!--
## This is a **Code Change**

- [] I have updated the documentation to reflect the changes.
- [] I have run `task pyright` and fixed the relevant issues.


TrapDrap:
The file structure and placements are not final! 
Comments and documentation have not been done.
I preferred working on the voice recorder feature standalone, touching as little of nextcord's main code as possible.
Once everything is working, I will work on splitting them into different files and merging them with nextcord's main code.
As for copyright, I'm not sure what I'm doing so I kinda just added what I think is right.

nextcord/cache.py
```diff
@@ -0,0 +1,102 @@
...
+class BetterTTLCache(dict):
+    __slots__ = ("maxsize", "ttl", "timelink", "working")
+
+    def __init__(
+        self, maxsize: Optional[int] = None, ttl: Optional[Union[int, float]] = None
+    ) -> None:
+        self.maxsize = maxsize or float("inf")
+        self.ttl = ttl
+        self.timelink = {}
+        self.working = False
```
teaishealthy, line 39:
```suggestion
        self.maxsize: float = float(maxsize) or float("inf")
        self.ttl: float = float(ttl)
        self.timelink = {}
        self.working: bool = False

```

maxsize will be inf even if 0 is passed, not sure if this matters.

nextcord/voicerec/core.py
```diff
@@ -0,0 +1,338 @@
...
+SILENCE_STRUCT_10 = SILENCE_STRUCT * 10
+SILENCE_STRUCT_100 = SILENCE_STRUCT * 100
+SILENCE_STRUCT_1000 = SILENCE_STRUCT * 1000  # 3.4mb in size
+
+
+class Silence:
+    def __init__(self, frames: int) -> None:
+        self.frames: int = frames
+
+    def write_to(self, buffer: Union[BufferedWriter, BytesIO]) -> None:
```
teaishealthy, line 62: buffer should be typed as `io.BufferedIOBase`

nextcord/voicerec/core.py
```diff
@@ -0,0 +1,338 @@
...
+        write = buffer.write
+        frames = self.frames
```
teaishealthy, line 64: why?

nextcord/voicerec/core.py
```diff
@@ -0,0 +1,338 @@
...
+        if temp_type == TempType.File:
+            self.buffer = open_tmp_file(guild_id, user_id, "wb+")
+        elif temp_type == TempType.Memory:
+            self.buffer = BytesIO()
+        else:
+            raise InvalidTempType(
+                f"Arg `temp_type` must be of type TempType not `{type(temp_type)}`"
+            )
+
+    def write(self, bytes) -> None:
```
teaishealthy, line 143: bytes should be `io.BufferedIOBase`

teaishealthy (changes requested):

nextcord/cache.py
```diff
@@ -0,0 +1,102 @@
+from time import time as now
+from typing import Optional, Union
+
+__set_item__ = dict.__setitem__
+__get_item__ = dict.__getitem__
+__get__ = dict.get
+__pop__ = dict.pop
```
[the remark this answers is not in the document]
TrapDrap: Yes, although this avoids dot narration which I tested was faster

nextcord/cache.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+class SizedDict(dict):
+    __slots__ = "maxsize"
+
+    def __init__(self, maxsize: Optional[Union[float, int]] = None) -> None:
+        if not maxsize:
+            maxsize = float("inf")
+        else:
+            self.maxsize = maxsize
```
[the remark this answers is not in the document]
TrapDrap: Yes thank you, although that isn't used (I copied code from my bot lol) so it shouldn't matter. Will commit either way

nextcord/voicerec/core.py
TrapDrap: Thank you, will get to that when I work on it again

nextcord/voicerec/core.py
TrapDrap: Avoiding dot notation when the variable could be used many times, local variables are faster.

nextcord/voicerec/core.py
TrapDrap: (And I also shouldn't overwrite the type bytes, will resolve that next update

nextcord/cache.py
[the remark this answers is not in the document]
EmreTech: Does it really have that big of a speed difference? Using the base classes' methods in this manner makes the code less readable.

TrapDrap:
I was also wondering if there's a better way of doing my silence writer, right now I'm relying on calling write multiple times in a while loop with bigger to smaller silence frames, I was wondering whether there's a way to create a bytes-like object that can just write the silence frame x times without having to call write x times. It would also not have to cache the 3.5mb or so of silence frames for it to work. The silence is always the same data, so it would be a lot simpler if there's a better way I can write x amount of this data without creating it in memory (multiplication does that, a silence for a minute or two will result in like 100mb of ram spike)

nextcord/voicerec/core.py
```diff
@@ -0,0 +1,338 @@
...
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
```
EmreTech, line 23:
```suggestion
# SPDX-License-Identifier: MIT
```
Nextcord does not use this module level docstring anymore. Instead, we use a SPDX license identifier comment.

nextcord/voicerec/decrypter.py
```diff
@@ -0,0 +1,64 @@
...
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
```
EmreTech, line 24:
```suggestion
# SPDX-License-Identifier: MIT
```

nextcord/voicerec/errors.py
```diff
@@ -0,0 +1,45 @@
+"""
+The MIT License (MIT)
+
+:copyright: (c) 2021-present Nextcord Developers
+
+:license: See license for more info
+"""
```
EmreTech, line 7:
```suggestion
# SPDX-License-Identifier: MIT
```

nextcord/voicerec/exporters.py
```diff
@@ -0,0 +1,193 @@
+import os
```
EmreTech, line 1:
```suggestion
# SPDX-License-Identifier: MIT

import os
```

nextcord/voicerec/opus.py
```diff
@@ -0,0 +1,92 @@
...
+all copies or substantial portions of the Software.
+
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS
+OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER
+DEALINGS IN THE SOFTWARE.
+"""
```
EmreTech, line 25:
```suggestion
# SPDX-License-Identifier: MIT
```

nextcord/voicerec/shared.py
```diff
@@ -0,0 +1,47 @@
+"""
+"""
```
EmreTech, line 2:
```suggestion
# SPDX-License-Identifier: MIT
```

EmmmaTech (changes requested):

nextcord/voicerec/shared.py
```diff
@@ -0,0 +1,46 @@
...
+format_exports = {
+    Formats.MP3: "export_with_ffmpeg",
+    Formats.MP4: "export_with_ffmpeg",
+    Formats.M4A: "export_with_ffmpeg",
+    Formats.MKA: "export_with_ffmpeg",
+    Formats.MKV: "export_with_ffmpeg",
+    Formats.OGG: "export_with_ffmpeg",
+    Formats.PCM: "export_as_PCM",
+    Formats.WAV: "export_as_WAV",
+}
```
teaishealthy, line 34: why not map directly to functions?

TrapDrap: > why not map directly to functions?

I don't exactly remember rn, I although know that I considered mapping directly to the functions, I must've had a reason.

TrapDrap:
no way!! it worked 💀

nextcord/voicerec/core.py
TrapDrap: closing, reversed change due to incorrect typing (missing methods used in io.BufferedIOBase

nextcord/voicerec/core.py
TrapDrap: closing, reversed change due to incorrect typing (missing methods used in io.BufferedIOBase

TrapDrap:
I added periodic silence writes which relieves a huge issue with writing silences long term. This defaults to be on, but the delay can be adjusted. Without this feature, if a user hasn't spoken in 3 minutes, it will try to write all 3 minutes of silence the next time they speak, which could result in massive blockages.

nextcord/voicerec/shared.py
TrapDrap: changed to address issue

nextcord/cache.py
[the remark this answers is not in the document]
TrapDrap: Resolving as this results in a 2x+ significance in speedup. Micro-optimization of cache is beneficial as this will be reused in future PRs on cache reworks, these can be called millions of times per second, twice the speed is a major bump well worth slightly less readability in this case.
A detailed comment has been added to describe the reasoning.

nextcord/cache.py
TrapDrap: forgot to mention, this is intentional

nextcord/cache.py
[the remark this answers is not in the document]
TrapDrap: This has been reverted. Max size is supposed to be int, the float type was included as support for infinity

nextcord/voicerec/core.py
TrapDrap: Resolved due to no longer relevant, implement changed

nextcord/cache.py
```diff
@@ -0,0 +1,149 @@
...
+__get_item__ = dict.__getitem__
+__get__ = dict.get
+__pop__ = dict.pop
+
+
+class CacheMissing(Enum):
+    MISSING = 0
+
+
+_MISSING = CacheMissing.MISSING
```
alentoghostflame, line 12: Why not `MISSING` from `nextcord.utils`?

nextcord/cache.py
```diff
@@ -0,0 +1,149 @@
...
+        """
+        Takes care of expiring keys in the timelink that have passed their ttl lifespan.
+        Uses a rolling motion for expiring, all items at the start of the timelink
+        are assumed to be the earliest created items to avoid checking every item.
+        Timelink order must be managed properly in order for expiration to work.
+        """
+        if not self.ttl or self.working:
+            return
+
+        self.working = True
```
alentoghostflame, line 87: I'm confused what this is for? Threading support?

nextcord/cache.py
```diff
@@ -0,0 +1,149 @@
...
+        if key not in self.timelink:
+            self.timelink[key] = now()
+            return True
+        return False
+
+    def update_time(self, key) -> None:
+        """
+        Update the timelink for a key.
+        Creates the key anew if it does not exist in the timelink.
+        Make sure the item exists in self, could result in memory leak if used improperly!
```
alentoghostflame, line 139: How significantly does a `if key not in self: raise KeyError` slow this down?

nextcord/cache.py
```diff
@@ -0,0 +1,141 @@
...
+    def __init__(self, maxsize: Optional[Union[float, int]] = None) -> None:
+        self.maxsize: Union[float, int] = maxsize or float("inf")
+
+    def __setitem__(self, key, val) -> None:
+        """Ensures doesn't exceed max size before adding new item."""
+        if len(self) >= self.maxsize:
+            for item in self:  # iter to delete first item
+                del self[item]
+                break
+        super().__setitem__(key, val)
```
alentoghostflame, line 36: I know you just changed it, but if speed is an issue would doing `dict.__setitem__` adequately speed things up?

nextcord/cache.py
TrapDrap: > Why not `MISSING` from `nextcord.utils`?
Because in this case the cache is expected to be able to store nextcord's MISSING object, without its own MISSING, it could cause issues with ttl & returning content

nextcord/cache.py
TrapDrap: I'm not sure whether it's necessary to be honest, I had it as an extra layer of insurance. Ensures the method doesn't get called twice at once, if that happens it can cause an error during iteration.

nextcord/cache.py
TrapDrap: It wouldn't although it's intentionally designed this way, this method is supposed to be a helper method, and the checks depends on whichever method implements it.

nextcord/cache.py
alentoghostflame: Due to no async calls being made and ignoring threading (we discourage threading), `__expire__` shouldn't ever be able to run while it's already running.

nextcord/cache.py
TrapDrap: > I know you just changed it, but if speed is an issue would doing `dict.__setitem__` adequately speed things up?

Yes, `dict.__setitem__` is significantly faster than `super()`, only very slightly behind a relay variable due to attribute lookup, it would be a preferred approach if I could.

nextcord/cache.py
TrapDrap: Then I could remove this for a tiny speedup, would it be worth it in case the user decides to access the cache from a thread and it completely breaks the ttl structure?

alentoghostflame: If threading support was important, we'd want to use something like `threading.Lock` to properly make sure we don't have multiple `__expire__`'s running.
Also as is, iterating through `BetterTTLCache().items()` and triggering `__expire__` would cause a Changed Size During Iteration error if `__expire__` hit the `del self[item]` stuff. I think you'd need to have everything use the same lock to prevent that, and tbh throwing a "Does not support threading" docstring in the class seems easier than doing all of that.

TrapDrap: Fair, I will add that next commit

nextcord/voice_recording/opus.py
```diff
@@ -0,0 +1,65 @@
...
+    def decode(self, opus_frame) -> None:
+        self.decode_queue.append(opus_frame)
+
+    def run(self) -> None:
+        while not self._end.is_set():
+            try:
+                opus_frame: OpusFrame = self.decode_queue.pop(0)
+
+            except IndexError:
+                sleep(0.001)
```
teaishealthy, line 39: What and why?

nextcord/voice_recording/shared.py
```diff
@@ -0,0 +1,36 @@
...
+def open_tmp_file(guild_id, user_id, mode):
+    path = Path(f"{getcwd()}/{TMP_DIR}/{guild_id}.{user_id}.tmp")
+
+    try:
+        return open(path, mode)  # noqa: SIM115
+
+    except FileNotFoundError:
+        makedirs(dirname(path), exist_ok=True)  # Create missing dirs
+        path.touch(exist_ok=True)  # create missing file
+        return open(path, mode)  # noqa: SIM115
```
teaishealthy, line 36: https://docs.python.org/3/library/tempfile.html

nextcord/voice_recording/opus.py
```diff
@@ -0,0 +1,65 @@
...
+from typing import TYPE_CHECKING, Dict
+
+import nextcord as nc
+
+if TYPE_CHECKING:
+    from .core import OpusFrame, RecorderClient
+
+
+class DecoderThread(Thread, nc.opus._OpusStruct):
+    def __init__(self, recorder) -> None:
```
teaishealthy, line 18: types

nextcord/voice_recording/errors.py
```diff
@@ -0,0 +1,45 @@
+# SPDX-License-Identifier: MIT
+
+import nextcord.errors as nc_errors
```
teaishealthy, line 3: why?

teaishealthy (changes requested):
a lot of types are missing, reviewing again when fixed 🦅🦅🦅🦅

nextcord/voice_recording/opus.py
TrapDrap: It is a while loop after all, sleeping when waiting for an item in the queue since the queue isn't expected to receive items more than that often. A frame only arrives every 20ms per user, sleeping during the wait can lower CPU usage.

nextcord/voice_recording/shared.py
TrapDrap: Thank you this, I will use this, don't have to worry about handling deletion pog

nextcord/voice_recording/errors.py
TrapDrap: import as prevents circular imports because it waits for the module to be ready before continuing

nextcord/voice_recording/opus.py
TrapDrap: Also, the most of opus.py is copied from the previous PR

TrapDrap: Although there is probably a more elegant approach to waiting for an item in queue

nextcord/voice_recording/shared.py
TrapDrap: Okay so I tried to implement this but it will not work as I rely on the names of temp files to access and overwrite them. Using python's built-in tempfile does not allow for naming the file on creation.

TrapDrap:
> a lot of types are missing, reviewing again when fixed 🦅🦅🦅🦅

Fixed

nextcord/cache.py
TrapDrap: Added a warning to be more implicit

nextcord/voice_recording/opus.py
teaishealthy: make the decode_queue an actual `queue.Queue` and use it's methods

teaishealthy (changes requested):
I think the way you're using (reimplementing) temp files is fundamentally flawed. Instead of this whole `tmp_type` junk and needing to overwrite filenames, keep a lookup table of buffers (!) and let the user bring their own buffer. If it walks like a buffer, it is a buffer. When you actually need to encode with a muxer that needs to be able to seek back, check if you got a file buffer, if not write your data to real temp file and then call ffmpeg.

nextcord/voice_recording/opus.py
```diff
@@ -0,0 +1,65 @@
...
+from time import sleep
+from typing import TYPE_CHECKING, Dict
+
+import nextcord as nc  # `import as` to prevent circular import
+
+if TYPE_CHECKING:
+    from .core import OpusFrame, RecorderClient
+
+
+class DecoderThread(Thread, nc.opus._OpusStruct):
```
EmreTech, line 17: Could we avoid using threads here?

Admiralmatt:
Would this feature be able to save to a file as it's recording or does it save in RAM as its recording? if the bot crashes partway through a (6 hour) recording the recording would be lost if it is not saving a temp file to disk.

nextcord/cache.py
```diff
@@ -0,0 +1,151 @@
+# SPDX-License-Identifier: MIT
```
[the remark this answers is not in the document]
EmreTech: Now conflicts with #1185 because of this.

nextcord/voice_recording/core.py
```diff
@@ -0,0 +1,1174 @@
...
+            raise TypeError(f"audio_format must be of type `Formats` not {type(audio_format)}")
+
+        return await export_methods[audio_format](
+            self, audio_format, tmp_type, filters=filters  # individual files
+        )
+
+
+# slots only work on python 3.10 with dataclasses
+@dataclass(**{"slots": True} if PYTHON_VERSION >= (3, 10) else {})
+class OpusFrame:
```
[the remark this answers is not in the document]
TrapDrap: Can't hardcode slots, causes an error if we want to support python versions earlier than 3.10

nextcord/voice_recording/core.py
```diff
@@ -0,0 +1,1174 @@
...
+        # data
+        self.time_tracker: Optional[TimeTracker] = None
+        self.audio_data: Optional[AudioData] = None
+        self.recording_paused: bool = False
+        self.filters: RecordingFilter = filters or RecordingFilter()
+
+        # processes
+        self.process: Optional[Thread] = None
+        self.auto_deaf: bool = auto_deaf if isinstance(auto_deaf, bool) else True
+        self.tmp_type: TmpType = tmp_type or TmpType.File
```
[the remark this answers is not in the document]
TrapDrap: tmp_type can be `None` in which case, default to using `TmpType.File`

nextcord/voice_recording/exporters.py
```diff
@@ -0,0 +1,371 @@
...
+                stdin=subprocess.PIPE,
+                stdout=subprocess.PIPE,
+                creationflags=FLAG,
+            )
+            process.communicate(data)
+        except FileNotFoundError as e:
+            raise NoFFmpeg(
+                "FFmpeg is not installed or aliased improperly. Unable to launch `ffmpeg` command."
+            ) from e
+
```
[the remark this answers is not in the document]
TrapDrap: That's true, I guess it could be a module. Will move into a module instead.

nextcord/abc.py
```diff
@@ -1771,15 +1807,19 @@ async def connect(
...
 
         if state._get_voice_client(key_id):
             raise ClientException("Already connected to a voice channel.")
 
         client = state._get_client()
-        voice = cls(client, self)
+        if recordable:
+            voice = RecorderClient(
+                client, self, auto_deaf, tmp_type, filters, periodic_write_seconds, prevent_leakage
+            )
```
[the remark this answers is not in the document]
TrapDrap: Users should be able to import RecorderClass and superclass it, the implementation in abc.py is just for native recording support like the VoiceClient natively supporting play.

I will look into your suggestion though, see if that's plausible without causing any breaking changes.

nextcord/cache.py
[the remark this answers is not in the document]
TrapDrap: I feel both modules should be merged. #1185 is still a draft after all, I don't feel this is a change that should be made in this PR.

nextcord/voice_recording/opus.py
```diff
@@ -0,0 +1,65 @@
...
+
+from __future__ import annotations
+
+import gc
+import logging
+from threading import Event, Thread
+from time import sleep
+from typing import TYPE_CHECKING, Dict
+
+import nextcord as nc  # `import as` to prevent circular import
```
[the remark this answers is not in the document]
TrapDrap: > I don't get the logic behind this? It makes sense in the errors module, but not here...

To be honest I don't know lol. There was a circular import issue and this solved it.

TrapDrap:
> Would this feature be able to save to a file as it's recording or does it save in RAM as its recording? if the bot crashes partway through a (6 hour) recording the recording would be lost if it is not saving a temp file to disk.

This allows you to save to file or ram as it's recording. You can also decide to intercept and process the audio packets yourself in your own manner. If say you recorded using RAM tmp files, the bot crashing will cause the files to be lost. If you record using File tmp files, the bot crashing will keep all the recorded pcm files, but you will need extra functions yourself to process the data into usable recording files.

nextcord/voice_recording/core.py
```diff
@@ -0,0 +1,1174 @@
...
+    ------
+    TypeError
+        The argument passed for any `user` was not :class:`int`, :class:`User` or :class:`Member`
+    """
+
+    __slots__ = ("blocklist", "allowlist", "client")
+
+    def __init__(
+        self,
+        client=None,
```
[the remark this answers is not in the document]
TrapDrap: (Has been a long time) But if I remember correctly, I wasn't able to typehint this because the class `client` takes (i.e. RecorderClient) was defined after, while the class also needs to call it in its functions so I can't just move it. If you know how I could typehint it, let me know.

nextcord/voice_recording/shared.py
```diff
@@ -0,0 +1,57 @@
...
+    Formats.M4A: ("m4a", "ipod"),
+    Formats.MKA: ("mka", "matroska"),
+    Formats.MKV: ("mkv", "matroska"),
+    Formats.OGG: ("ogg", "ogg"),
+    "mp3": ("mp3", "mp3"),
+    "mp4": ("mp4", "mp4"),
+    "m4a": ("m4a", "ipod"),
+    "mka": ("mka", "matroska"),
+    "mkv": ("mkv", "matroska"),
+    "ogg": ("ogg", "ogg"),
```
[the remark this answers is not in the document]
TrapDrap: This is used in the FFMPEG export. It's used to translate the enum into the string used by FFMPEG to dictate the output format.

ooliver1:
As per the description, this is in-progress so will be marked as draft.

PR #1115, opened 2023-08-26
refactor: remove _modify_api_version

EmreTech:
## Summary

This PR removes `http._modify_api_version`. It was a temporary workaround in the past around the message content intent changes that were made in API v10. Nowadays, it has no use and it's not recommended to use it, according to its own documentation.

<!--
## This is a **Code Change**

### DIFF
--- a/nextcord/http.py
+++ b/nextcord/http.py
@@ -98,38 +98,6 @@ _DEFAULT_API_VERSION = 10
 _API_VERSION: Literal[9, 10] = _DEFAULT_API_VERSION
 
 
-class UnsupportedAPIVersion(UserWarning):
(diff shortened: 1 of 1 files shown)


nextcord/http.py
```diff
@@ -98,38 +97,6 @@ async def json_or_text(response: aiohttp.ClientResponse) -> Union[Dict[str, Any]
 _API_VERSION: Literal[9, 10] = _DEFAULT_API_VERSION
```
ooliver1, line 97: Should this be just `Literal[10]` too?

ooliver1 (approved):

teaishealthy (approved):

PR #1120, opened 2023-09-19
fixed view=None

HumanBot000:
fixed 

```
AttributeError: 'NoneType' object has no attribute 'to_components'
```
if view=None


HumanBot000:
yes but if the user enters None as paramtere

PR #1121, opened 2023-09-19
view=None user warning

HumanBot000:
if the user decides to use view=None instead of leaving it out


ooliver1:
`abc.Messageable.send`'a view parameter is not documented as allowing `None` as you can see by the type errors. This has not followed the contributing guidelines at `CONTRIBUTING.md` and not used the pull request template.

Soheab:
You can use `nextcord.utils.MISSING` fyi.

PR #1123, opened 2023-09-23
feat(User & Member): update display_name method to include global_name

ToasterUwU:
## Summary

Without this change the display_name property returns a nickname if present, if not just the username (which is the hidden away username). It doesn't consider the "Display Name" also known as global_name at all.

In my opinion, that is bad behavior, because the global_name takes priority in the Discord UI over the username if it exists, so it should here as well. After all, when people use this property, they want the name the person is displayed as, not the hidden away unique username.

## This is a **Code Change**

### DIFF
--- a/nextcord/member.py
+++ b/nextcord/member.py
@@ -505,11 +505,14 @@ class Member(abc.Messageable, _UserTag):
     def display_name(self) -> str:
         """:class:`str`: Returns the user's display name.
 
-        For regular users this is just their username, but
-        if they have a guild specific nickname then that
-        is returned instead.
+        Returns the users displayed name.
+        So whatever is the first thing to exist from the following priority list:
+
+        1. Guild specific nickname
+        2. Global Name (also known as 'Display Name' in the Discord UI)
+        3. Unique username
         """
-        return self.nick or self.name
+        return self.nick or self.global_name or self.name
 
     @property
     def display_avatar(self) -> Asset:
--- a/nextcord/member.py
+++ b/nextcord/member.py
@@ -506,7 +506,7 @@ class Member(abc.Messageable, _UserTag):
         """:class:`str`: Returns the user's display name.
 
         Returns the users displayed name.
-        So whatever is the first thing to exist from the following priority list:
+        This will return the name using the following hierachy:
 
         1. Guild specific nickname
         2. Global Name (also known as 'Display Name' in the Discord UI)
--- a/nextcord/member.py
+++ b/nextcord/member.py
@@ -505,7 +505,6 @@ class Member(abc.Messageable, _UserTag):
     def display_name(self) -> str:
         """:class:`str`: Returns the user's display name.
 
-        Returns the users displayed name.
         This will return the name using the following hierachy:
 
         1. Guild specific nickname
--- a/nextcord/user.py
+++ b/nextcord/user.py
@@ -247,12 +247,10 @@ class BaseUser(_UserTag):
     def display_name(self) -> str:
         """:class:`str`: Returns the user's display name.
 
-        Returns the users displayed name.
         This will return the name using the following hierachy:
 
-        1. Guild specific nickname
-        2. Global Name (also known as 'Display Name' in the Discord UI)
-        3. Unique username
+        1. Global Name (also known as 'Display Name' in the Discord UI)
+        2. Unique username
         """
         return self.global_name or self.name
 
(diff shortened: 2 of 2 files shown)


nextcord/user.py
```diff
@@ -247,11 +247,14 @@ def created_at(self) -> datetime:
     def display_name(self) -> str:
         """:class:`str`: Returns the user's display name.
 
-        For regular users this is just their username, but
-        if they have a guild specific nickname then that
-        is returned instead.
+        Returns the users displayed name.
+        So whatever is the first thing to exist from the following priority list:
+
+        1. Guild specific nickname
```
spifory, line 253: This doesn't apply since this is for `User`, so you probably want to just mention that it returns the global name, or username

nextcord/member.py
```diff
@@ -505,11 +505,14 @@ def mention(self) -> str:
     def display_name(self) -> str:
         """:class:`str`: Returns the user's display name.
 
-        For regular users this is just their username, but
-        if they have a guild specific nickname then that
-        is returned instead.
+        Returns the users displayed name.
+        So whatever is the first thing to exist from the following priority list:
```
spifory, line 509: Something along the lines of
```suggestion
        This will return the name using the following hierachy:
```

nextcord/user.py
```diff
@@ -247,11 +247,14 @@ def created_at(self) -> datetime:
     def display_name(self) -> str:
         """:class:`str`: Returns the user's display name.
 
-        For regular users this is just their username, but
-        if they have a guild specific nickname then that
-        is returned instead.
+        Returns the users displayed name.
+        So whatever is the first thing to exist from the following priority list:
```
spifory, line 251: Same here
```suggestion
        This will return the name using the following hierachy:
```
Alternatively, something along the lines of `Returns the user's global name if any, else the username will be returned`

nextcord/member.py
```diff
@@ -505,11 +505,14 @@ def mention(self) -> str:
     def display_name(self) -> str:
         """:class:`str`: Returns the user's display name.
 
-        For regular users this is just their username, but
-        if they have a guild specific nickname then that
-        is returned instead.
+        Returns the users displayed name.
```
spifory, line 508: This part should be removed. It's already explaining what this function does on line 2 lines above.

nextcord/user.py
```diff
@@ -247,11 +247,14 @@ def created_at(self) -> datetime:
     def display_name(self) -> str:
         """:class:`str`: Returns the user's display name.
 
-        For regular users this is just their username, but
-        if they have a guild specific nickname then that
-        is returned instead.
+        Returns the users displayed name.
```
spifory, line 250: Same thing here (regarding the duplication).

spifory (changes requested):

spifory (approved):

Kevin-OVI:
Shouldn’t this change also affect nextcord.WidgetMember ?

ToasterUwU:
> Shouldn’t this change also affect nextcord.WidgetMember ?

I had no idea that this exists lol. But since it inherits from User, which got the change, it will have the same thing happen to it as well.

ooliver1:
> I had no idea that this exists lol. But since it inherits from User, which got the change, it will have the same thing happen to it as well.

Except it doesn't. It inherits from `BaseUser` and has it's own `display_name` impl anyway.

Skelmis (approved):

Skelmis (commented):
Care to also do this for `WidgetMember` as per comments?

ToasterUwU:
I can, what is that one for anyways? I never heard of it before

Skelmis:
Honestly a bit unsure myself

ToasterUwU:
Ok, just in case i will sit this out until someone tells me what it is exactly and I can be sure this behavior is wanted

Astroyo:
i have been waiting for some change like this for ages 😂

Skelmis:
> Ok, just in case i will sit this out until someone tells me what it is exactly it is and I can be sure this behavior is wanted

I would just copy the user functionality over to the widget member class

ToasterUwU:
So, why are we waiting here?

EmreTech:
Development is pretty inactive right now, main devs are just busy with personal stuff.

EmreTech:
> I can, what is that one for anyways? I never heard of it before

According to the Discord API Docs, it's a partial user object that's included with Guild Widgets. As Oliver said though, that object has its own `.display_name` implementation.

EmreTech (approved):

PR #1095, opened 2023-07-18
refactor!: remove discord module alias

EmreTech:
## Summary

This PR removes the `discord` alias that Nextcord has included since 2.0.

<!--
## This is a **Code Change**

- [ ] I have tested my changes.
- [ ] I have updated the documentation to reflect the changes.
- [ ] I have run `task pyright` and fixed the relevant issues.

PR #1125, opened 2023-09-27
refactor(health_check): replace pkg_resources with importlib.metadata

cdce8p:
## Summary
`pkg_resources` is deprecated and emits a warning when used. Replace it with `importlib.metadata`.
https://docs.python.org/3/library/importlib.metadata.html

```
/.../nextcord/health_check.py:1: DeprecationWarning: pkg_resources is deprecated as an API. See https://setuptools.pypa.io/en/latest/pkg_resources.html
    from pkg_resources import get_distribution, DistributionNotFound
```

### DIFF
--- a/nextcord/health_check.py
+++ b/nextcord/health_check.py
@@ -1,8 +1,7 @@
 # SPDX-License-Identifier: MIT
 
+from importlib.metadata import PackageNotFoundError, metadata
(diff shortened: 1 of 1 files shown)


ooliver1:
This is already going to be removed with #1095

PR #1001, opened 2023-02-24
chore: migrate to poetry

teaishealthy:
**Summary**

Migrate to poetry as the setup provider

- Migrated all setup.py metadata and dependencies into pyproject.toml
- Deleted setup.py, requirements.txt, requirements-dev.txt, MANIFEST.in (poetry handles all of this)
- Created **an optional** plugin for poetry that will add the git hash to the version number on development builds.

> **Note**
> To install the plugin, you need to find the poetry installation directory (see below) and run `<poetry-install-dir>/venv/bin/pip install ./scripts/versioner` (one time process), hopefully this process will get simplified in the future 

**Install directories**
- ~/Library/Application Support/pypoetry on MacOS.
- ~/.local/share/pypoetry on Linux/Unix.
- %APPDATA%\pypoetry on Windows.

**TODO:**

- [ ] Unaliased setup
- [ ] Actions

PR #1126, opened 2023-09-27
build: use poetry over pip+setuptools

ooliver1:
## Summary

Extension of #1001, relies on #1095.

This makes use of (a fork of) [`poetry-dynamic-versioning`](https://pypi.org/project/poetry-dynamic-versioning/) to support git metadata in the version string. This also means that the version strings and tuples no longer need to be manually changed, it is all done via git tags.

`poetry-dynamic-versioning` works by modifying `__version__` (and `version_info` with the fork) when installing or `poetry build`ing a package. This means that the version string is not constantly modifed, but kept up to date, and modified on every installation from git.

The revision part of pre-releases via git (`3.0.0a4+abcdef1`, the "4" part) is the distance in commits from when the tag was made, instead of the original implementation using the absolute commit count.

`aiohttp[speedups]` is not a valid version requirement with poetry, which apparently will be better with [PEP 621 support](https://github.com/python-poetry/poetry-core/pull/613#pullrequestreview-1553260196) or that PR being merged, so unfortunately we have to manually specify aiodns and brotli here. I did not add cchardet as that is unmaintained and not included in latest aiohttp dependencies.

<!--
## This is a **Code Change**


ooliver1:
And there we are, actions passing.

teaishealthy (approved):
tested

PR #1133, opened 2023-10-01
refactor: bump pyright to 1.1.333

ooliver1:
## Summary

Bump to [Pyright 1.1.333](https://github.com/microsoft/pyright/releases/tag/1.1.333). This makes changes to `application_commands.py` to support the fact that `ellipsis` is `@type_check_only`. `Range` and `String` need to support type checkers properly with `Annotated` and such, but this is a patch for now to bump to latest Pyright.

## This is a **Code Change**


nextcord/application_command.py
```diff
@@ -68,6 +68,11 @@
     _CustomTypingMetaBase = Any
 else:
     _CustomTypingMetaBase = object
+    # `ellipsis` is a type-checking only variable. This assignment avoids ruff `F821`
+    # as this is defined at runtime now.
+    ellipsis = ...
+
+EllipsisType = type(...)
```
Skelmis, line 75: Is this meant to be `Ellipsis`?

Skelmis (approved):

nextcord/audit_logs.py
```diff
@@ -462,11 +462,10 @@ def _from_data(self, data: AuditLogEntryPayload) -> None:
 
         # this key is technically not usually present
         self.reason = data.get("reason")
-        self.extra = data.get("options", {})  # type: ignore
-        # I gave up trying to fix this
+        self.extra = extra = data.get("options", {})  # type: ignore
```
Skelmis, line 465: Is this meant to be changed to `extra = ...`? Otherwise the rest of the code changes don't work

Skelmis (changes requested):

nextcord/audit_logs.py
ooliver1: Why wouldn't they? This double assignment is kinda a patch for pyright inference constantly changing every update. With this inference and at least one use of it, it works, which is so irritating.

Skelmis: Sorry, didn't see the double assign here when reviewing

Skelmis:
Care to resolve conflicts with the base branch

Skelmis (approved):

ooliver1:
This may be increased to 340 as that is out, but it is producing some wild errors (thinking `TextChannel` is a `TypeVar`?, unnecessary type ignore which complains when I remove it, ...) on my laptop, so I will try again on desktop.

EmreTech:
Any updates? Latest version is now [1.1.369](https://github.com/microsoft/pyright/releases/tag/1.1.369).

ooliver1:
Closing this to get it out of the list, there are so many versions that basically make all these changes irrelevant.

PR #1160, opened 2024-01-03
feat(webhooks): allow editing webhook messages that were sent in threads

Ednaordinary:
## Summary

Nextcord doesn't currently allow editing webhook messages that were sent in threads. This aims to change that. I tested it preliminarily but it definitely needs more tests. Will sort it out soon. Most of it is copied straight from the implementation for sending webhook messages in threads.

## This is a **Code Change**

### DIFF
--- a/nextcord/webhook/async_.py
+++ b/nextcord/webhook/async_.py
@@ -1684,7 +1684,7 @@ class Webhook(BaseWebhook):
         thread: :class:`~nextcord.abc.Snowflake`
             The thread that the message to be edited is in.
 
-            .. versionadded:: ?.?
+            .. versionadded:: 3.0
 
         Raises
         ------
--- a/nextcord/webhook/async_.py
+++ b/nextcord/webhook/async_.py
@@ -332,7 +332,9 @@ class AsyncWebhookAdapter:
             webhook_token=token,
             message_id=message_id,
         )
-        return self.request(route, session, payload=payload, multipart=multipart, files=files, params=params)
+        return self.request(
+            route, session, payload=payload, multipart=multipart, files=files, params=params
+        )
 
     def delete_webhook_message(
         self,
@@ -1677,11 +1679,11 @@ class Webhook(BaseWebhook):
             The updated view to update this message with. If ``None`` is passed then
             the view is removed. The webhook must have state attached, similar to
             :meth:`send`.
-            
+
             .. versionadded:: 2.0
         thread: :class:`~nextcord.abc.Snowflake`
             The thread that the message to be edited is in.
-            
+
             .. versionadded:: ?.?
 
         Raises
@@ -1731,7 +1733,7 @@ class Webhook(BaseWebhook):
         thread_id: Optional[int] = None
         if thread is not MISSING:
             thread_id = thread.id
-            
+
         data = await adapter.edit_webhook_message(
             self.id,
             self.token,
--- a/nextcord/webhook/async_.py
+++ b/nextcord/webhook/async_.py
@@ -320,7 +320,11 @@ class AsyncWebhookAdapter:
         payload: Optional[Dict[str, Any]] = None,
         multipart: Optional[List[Dict[str, Any]]] = None,
         files: Optional[List[File]] = None,
+        thread_id: Optional[int] = None,
     ) -> Response[Message]:
+        params = {}
+        if thread_id:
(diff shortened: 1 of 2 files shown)


Ednaordinary:
meant to mark this as ready for review earlier. I've merged this into enough environments and used it enough that I consider it "tested"

nextcord/webhook/async_.py
```diff
@@ -1674,6 +1681,10 @@ async def edit_message(
             :meth:`send`.
 
             .. versionadded:: 2.0
+        thread: :class:`~nextcord.abc.Snowflake`
+            The thread that the message to be edited is in.
+
+            .. versionadded:: ?.?
```
xXenvy, line 1687: The next version is `3.0`

nextcord/webhook/sync.py
```diff
@@ -1048,7 +1055,8 @@ def edit_message(
         allowed_mentions: :class:`AllowedMentions`
             Controls the mentions being processed in this message.
             See :meth:`.abc.Messageable.send` for more information.
-
+        thread: :class:`~nextcord.abc.Snowflake`
+            The thread that the message to be edited is in.
         Raises
```
xXenvy, line 1060: Add new line so it doesn't overlap + versionadded is missing

nextcord/webhook/sync.py
```diff
@@ -1019,6 +1025,7 @@ def edit_message(
         files: List[File] = MISSING,
         attachments: List[Attachment] = MISSING,
         allowed_mentions: Optional[AllowedMentions] = None,
+        thread: Snowflake = MISSING,
```
xXenvy, line 1028: Here it is named `thread`, and in other methods it is `thread_id` ? I think it is better to stick to one name.

xXenvy (changes requested):

nextcord/webhook/sync.py
xXenvy: No, it's not a thread object.

xXenvy: Aaah, I see what you mean. Why is it not just accepting `Thread` here?

Ednaordinary: > Here it is named `thread`, and in other methods it is `thread_id` ? I think it is better to stick to one name.

I'll have to double check this but I think I just kept the same inconsistency that sending in threads has. I thought it was weird too

Ednaordinary: > In this thread is a thread object, and in the other method thread_id is a thread id (which is an int ot a str, not sure) so the usage of var names is correct

thread_id is an int, btw

xXenvy: > So you just need to pass sth with an .id attribute (which is constructible using nextcord.Object(id=1234) ) instead of fetching the full thread

So why not just use the actual id? xD

xXenvy: Well, but I meant not to take the object itself. Rename it thread_id give typehint as int, because we don't need the WHOLE object, just its id.

Ednaordinary: Should I do that? If so, send() should also be changed for consistency

EmmmaTech: > Well, but I meant not to take the object itself. Rename it thread_id (or smth) give typehint as int, because we don't need the WHOLE object, just its id.

I don't like that approach. From the end user perspective, it seems more natural to just provide the thread object if you have one. And if you don't have one, then we have the `Object` placeholder to help out so developers don't have to write for unions of objects and ints.

EmmmaTech (approved):
Looks good :D

Skelmis (approved):

Skelmis:
@ooliver1 I'd almost say docs failure is un-related, care to merge or provide some more insight if its not?

nextcord/webhook/sync.py
Skelmis: This pattern of differing items based on the area they are present makes sense when looking at the library as it currently stands

ooliver1:
> @ooliver1 I'd almost say docs failure is un-related, care to merge or provide some more insight if its not?

@Skelmis I am unable to push a merge commit to fix it, I believe it was some changes to include 3.12 made all old PRs not be quite compatible. It builds locally though so it is okay.

PR #1181, opened 2024-04-17
feat(app-cmds): Implement User-Installed Apps

Kevin-OVI:
## Summary

This PR implements the "integration_types" and "contexts" fields on app commands to support User-Installed Apps. It also deprecates the "dm_permission" parameter in all application commands, as it has been replaced by "contexts".
The PR is marked as draft because user apps are still in preview.

## This is a **Code Change**

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -41,7 +41,14 @@ from .channel import (
     TextChannel,
     VoiceChannel,
 )
-from .enums import ApplicationCommandOptionType, ApplicationCommandType, ChannelType, Locale
+from .enums import (
+    ApplicationCommandOptionType,
+    ApplicationCommandType,
+    ChannelType,
+    Locale,
+    IntegrationType,
+    InteractionContextType
+)
 from .errors import (
     ApplicationCheckFailure,
     ApplicationCommandOptionMissing,
@@ -1953,6 +1960,8 @@ class BaseApplicationCommand(CallbackMixin, CallbackWrapperMixin):
         dm_permission: Optional[bool] = None,
         default_member_permissions: Optional[Union[Permissions, int]] = None,
         nsfw: bool = False,
+        integration_types: Optional[Iterable[IntegrationType]] = None,
+        contexts: Optional[Iterable[InteractionContextType]] = None,
         parent_cog: Optional[ClientCog] = None,
         force_global: bool = False,
     ) -> None:
@@ -1982,6 +1991,10 @@ class BaseApplicationCommand(CallbackMixin, CallbackWrapperMixin):
         dm_permission: :class:`bool`
             If the command should be usable in DMs or not. Setting to ``False`` will disable the command from being
             usable in DMs. Only for global commands, but will not error on guild.
+
+            .. warning::
+                This field is deprecated, use ``contexts`` instead.
+            .. deprecated:: 3.0
         default_member_permissions: Optional[Union[:class:`Permissions`, :class:`int`]]
             Permission(s) required to use the command. Inputting ``8`` or ``Permissions(administrator=True)`` for
             example will only allow Administrators to use the command. If set to 0, nobody will be able to use it by
@@ -1990,6 +2003,14 @@ class BaseApplicationCommand(CallbackMixin, CallbackWrapperMixin):
             Whether the command can only be used in age-restricted channels. Defaults to ``False``.
 
             .. versionadded:: 2.4
+        integration_types: Optional[Iterable[:class:`IntegrationType`]]
+            Where the command is available, only for globally-scoped commands. Defaults to ``guild_install``.
+
+            .. versionadded:: 3.0
+        contexts: Optional[Iterable[:class:`InteractionContextType`]]
+            Where the command can be used, only for globally-scoped commands. By default, all interaction context types included for new commands.
+
+            .. versionadded:: 3.0
         parent_cog: Optional[:class:`ClientCog`]
             ``ClientCog`` to forward to the callback as the ``self`` argument.
         force_global: :class:`bool`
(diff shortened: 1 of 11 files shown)


nextcord/enums.py
```diff
@@ -2037,6 +2039,32 @@ class ForumLayoutType(IntEnum):
     """Display posts as a collection of posts with images, this is more image focused."""
 
 
+class IntegrationType(IntEnum):
```
spifory, line 2042: Both class docstrings (and their parameters' docstrings) added in this file are missing periods at the ends.

spifory (changes requested):
You likely want to take a look at https://github.com/discord/discord-api-docs/pull/6723 to implement as it seems a lot is missing in this PR (I do understand this is still a draft though)

Kevin-OVI:
> You likely want to take a look at [discord/discord-api-docs#6723](https://github.com/discord/discord-api-docs/pull/6723) to implement as it seems a lot is missing in this PR (I do understand this is still a draft though)

I only had implemented what I had seen as necessary to enable user installed apps, without checking what had actually changed in the change-log. I think everything from there is now implemented.

Kevin-OVI:
I see that the pre-commit.ci check is failing. Is there anyway I can fix the rust compiler not being installed ? I ran `task lint` on my side before committing so there shouldn't be any issues.

alentoghostflame:
> I see that the pre-commit.ci check is failing. Is there anyway I can fix the rust compiler not being installed ? I ran `task lint` on my side before committing so there shouldn't be any issues.

I don't believe so, we have to wait for a `libcst` wheel to be published for whatever platform + version the pre-commit.ci check uses.

EmreTech (changes requested):
Looks good otherwise!

spifory (changes requested):
I seem to have made this review a month ago and forgot, but anyways just some nitpicks in the code

nextcord/application_command.py
```diff
@@ -1990,6 +2003,14 @@ def __init__(
             Whether the command can only be used in age-restricted channels. Defaults to ``False``.
 
             .. versionadded:: 2.4
+        integration_types: Optional[Iterable[Union[:class:`IntegrationType`, :class:`int`]]]
+            Where the command is available, only for globally-scoped commands. Defaults to ``guild_install``.
+
+            .. versionadded:: 3.0
+        contexts: Optional[Iterable[Union[:class:`InteractionContextType`, :class:`int`]]]
+            Where the command can be used, only for globally-scoped commands. By default, all interaction context types included for new commands.
```
teaishealthy, line 2011:
```suggestion
            Where the command can be used, only for globally-scoped commands. By default, all interaction context types are included for new commands.
```

nextcord/application_command.py
```diff
@@ -2007,11 +2028,21 @@ def __init__(
...
-        self.dm_permission: Optional[bool] = dm_permission
         self.default_member_permissions: Optional[
             Union[Permissions, int]
         ] = default_member_permissions
         self.nsfw: bool = nsfw
+        self.integration_types = (
+            None if integration_types is None else [IntegrationType(x) for x in integration_types]
+        )
+
+        if dm_permission is not None:
```
teaishealthy, line 2039: DeprecationWarning?

teaishealthy (changes requested):

nextcord/message.py
```diff
@@ -620,6 +628,132 @@ def created_at(self) -> datetime.datetime:
...
+        self.triggering_interaction_metadata: Optional[MessageInteractionMetadata] = (
+            MessageInteractionMetadata(
+                data=data["triggering_interaction_metadata"], guild=guild, state=state
+            )
+            if "triggering_interaction_metadata" in data
+            else None
+        )
+
+    def __repr__(self) -> str:
+        return f"<{self.__class__.__name__} id={self.id} type={self.type} user={self.user!r} name={self.name!r}"
```
EmreTech, line 739:
```suggestion
        return f"<{self.__class__.__name__} id={self.id} type={self.type} user={self.user!r} name={self.name!r}>"
```

EmreTech (commented):
Tested! 🎉

nextcord/application_command.py
Kevin-OVI: I was thinking this argument could be removed as it's a breaking change and the next version would be 3.0 anyway.

EmreTech:
This may need updates from the new [changelog](https://discord.com/developers/docs/change-log#june-27-2024-api-updates).

tizu69:
I'm not too familiar with this, is thre some way to use this already?

xXenvy:
> I'm not too familiar with this, is thre some way to use this already?

If you want to play with it, go ahead, but if you want to use it for ‘production’ - I would strongly suggest that you don't.

To install: `pip install -U git+https://github.com/Kevin-OVI/nextcord.git@user-apps`
U may also need docs: https://nextcord--1181.org.readthedocs.build/en/1181/

alentoghostflame:
@Kevin-OVI, are you still working on this? Once you get the conflicts resolved, I'll take a serious look at getting this merged in.

alentoghostflame (approved):
`dm_permission` either needs to be removed or a `DeprectionWarning` added (IMO it should be removed) but overall this PR looks really really good! While I haven't scrutinized every bit of the code, it seems like all newly added bits are in the correct files. Application command payloads compare correctly, and (after merging #1204) even lazy loading is working correctly!

Overall, this works nicely and everything seems to function as expected. A decision with `dm_permission` needs to be made and I would like to have at least one more solid approval before merging this in.

nextcord/application_command.py
Kevin-OVI: I've removed the `dm_permission` parameter. The PR should be ready to merge.

EmmmaTech (approved):
Looks good :3

alentoghostflame:
Alright, if anything goes badly @EmmmaTech is to blame obv.

EmmmaTech:
> Alright, if anything goes badly @EmmmaTech is to blame obv.

😭

PR #1183, opened 2024-04-25
feat(widget): update display_name method to include global_name

Kevin-OVI:
## Summary

This is a follow-up to #1123 . I had suggested to also apply that change to WidgetMember as it also has a display_name property, but that PR got merged without it.
This PR applies that same change as Member to WidgetMember : the display_name property will return the name using the following hierarchy:
1. Guild specific nickname
2. Global Name (also known as 'Display Name' in the Discord UI)
3. Unique username

## This is a **Code Change**

### DIFF
--- a/nextcord/widget.py
+++ b/nextcord/widget.py
@@ -174,8 +174,15 @@ class WidgetMember(BaseUser):
 
     @property
     def display_name(self) -> str:
-        """:class:`str`: Returns the member's display name."""
-        return self.nick or self.name
+        """:class:`str`: Returns the user's display name.
+
+        This will return the name using the following hierarchy:
+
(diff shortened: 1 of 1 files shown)


Skelmis:
I feel like 2 & 3 should maybe be swapped at first look, but this is a choice for others to agree or disagree with. Consider this comment an approval if others like the current order.

Kevin-OVI:
I don't think so, this is the order used by User and Member in the original PR, and also the one used in the Discord UI.

EmreTech (approved):
Yeah, looking back at it, this was actually needed in the original PR. Sorry about that!

Skelmis (approved):

PR #1185, opened 2024-05-20
feat!: async cache and object rewrite

alentoghostflame:
## Summary

This PR aims to
* make accesses to the cache asynchronous,
* allow the cache to be modified and replaced by users,
* refine how Discord objects are created,
* centralize the creation of Discord objects in a way that is modifiable and replaceable by users,
* reduce the dependency on guild intents,
* and allow bots to be closer to stateless.

To say this is in progress is an understatement. It will likely be quite a while before anything here is relatively usable, let alone useful. This is being published so people can monitor this and raise issues about the architecture before I dive too far into the wrong direction.

This is a **Code Change**

### DIFF
--- a/nextcord/cache.py
+++ b/nextcord/cache.py
@@ -105,7 +105,7 @@ class MemoryCache(BaseCache):
         self._members[(int(member_data["user"]["id"]), guild_id)] = member_data
 
     async def remove_member(self, member_id: int, guild_id: int) -> bool:
-        return self._members.pop((member_id, guild_id)) is not None
+        return self._members.pop((member_id, guild_id), None) is not None
 
     async def get_member(self, member_id: int, guild_id: int) -> Optional[MemberData]:
         return self._members.get((member_id, guild_id), None)
--- a/nextcord/cache.py
+++ b/nextcord/cache.py
@@ -1,0 +1,109 @@
+from __future__ import annotations
+
+from typing import Dict, Iterable, Optional
+from collections.abc import Collection
+from .types.guild import Guild as GuildData
+from .types.message import Message as MessageData
+from .types.user import User as UserData
+
+
+__all__ = (
+    "BaseCache",
+    "MemoryCache",
+    "NullCache",
+)
+
+
+class BaseCache:
+    def __init__(self, **kwargs):
+        pass
+
+    """A non-functional, NotImplemented cache class that any Nextcord and user made cache should subclass."""
+
+    # Guilds
+    async def add_guild(self, guild_data: GuildData):
+        raise NotImplementedError
+
+    async def get_guild(self, guild_id: int) -> Optional[GuildData]:
+        raise NotImplementedError
+
+    async def get_guild_ids(self) -> Iterable[int]:
+        # TODO: Should this return an async interator?
+        raise NotImplementedError
+
+    # Messages
+    async def add_message(self, message_data: MessageData):
+        raise NotImplementedError
+
+    async def remove_message(self, message_id: int) -> bool:
+        """Must return True if the message corresponding to the ID was found and removed, False if not found."""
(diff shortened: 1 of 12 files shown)


nextcord/cache.py
```diff
@@ -0,0 +1,103 @@
...
+from .types.user import User as UserData
+
+__all__ = (
+    "BaseCache",
+    "MemoryCache",
+    "NullCache",
+)
+
+
+class BaseCache:
```
Skelmis, line 16: Consider making this inherit `typing.Protocol` as well to help with tooling stuff

nextcord/cache.py
```diff
@@ -0,0 +1,103 @@
...
+        return self._guilds.get(guild_id, None)
+
+    async def get_guild_ids(self) -> Iterable[int]:
+        return tuple(self._guilds.keys())
+
+    async def add_message(self, message_data: MessageData) -> None:
+        self._messages[int(message_data["id"])] = message_data
+
+    async def remove_message(self, message_id: int) -> bool:
+        return True if self._messages.pop(message_id, None) is not None else False
```
spifory, line 88:
```suggestion
        return bool(self._messages.pop(message_id, None))
```

nextcord/message.py
```diff
@@ -77,6 +79,33 @@
...
+        ret.channel_id = int(payload["channel_id"])
+        ret.content = payload["content"]
+        ret.id = int(payload["id"])
+        ret.type = try_enum(MessageType, payload["type"])
+
+        return ret
+
+    def __repr__(self) -> str:
+        name = self.__class__.__name__
+        return f"<{name} id={self.id} channel_id={self.channel_id} type={self.type}>"
```
spifory, line 106:
```suggestion
        return f"<{name} id={self.id!r} channel_id={self.channel_id!r} type={self.type!r}>"
```

spifory (changes requested):

nextcord/cache.py
teaishealthy: Could also just be renamed to just `Cache`, the subclasses are alredy named specific enough imo

nextcord/gateway.py
```diff
@@ -547,7 +548,10 @@ async def received_message(self, msg: Union[str, bytes], /) -> None:
         except KeyError:
             _log.debug("Unknown event %s.", event)
         else:
-            func(data)
+            if inspect.iscoroutinefunction(func):
+                await func(data)
+            else:
+                func(data)
```
teaishealthy, line 554: isn't this just `utils.maybe_coro`?

nextcord/cache.py
```diff
@@ -64,46 +79,57 @@ class NullCache(BaseCache):
...
-    async def get_guild_ids(self) -> Iterable[int]:
+    async def get_guild_ids(self) -> Sequence[int]:
         return tuple(self._guilds.keys())
 
-    async def add_message(self, message_data: MessageData):
+    async def add_member(self, member_data: MemberData, guild_id: int) -> None:
+        self._members[(int(member_data["user"]["id"]), guild_id)] = member_data
+
+    async def remove_member(self, member_id: int, guild_id: int) -> bool:
+        return self._members.pop((member_id, guild_id)) is not None
```
EmreTech, line 109: `dict.pop` actually raises `KeyError` if the key provided isn't found in the dictionary.

nextcord/cache.py
```diff
@@ -64,46 +79,57 @@ class NullCache(BaseCache):
...
+
+    async def get_member(self, member_id: int, guild_id: int) -> Optional[MemberData]:
+        return self._members.get((member_id, guild_id), None)
```
EmreTech, line 112: Usually `None` is the default for `dict.get`.

PR #1188, opened 2024-06-01
docs(audit-logs): document missing attribute

### LINKED ISSUE #1159: `AuditLogEntry` documentation is missing attributes
### Summary

The documentation for the `AuditLogEntry` object does not accurately describe all attributes on the class.

### Reproduction Steps

Compare the documentation found [here](https://docs.nextcord.dev/en/stable/api.html#nextcord.AuditLogEntry) with the class itself found [here](https://github.com/nextcord/nextcord/blob/master/nextcord/audit_logs.py#L390)

### Minimal Reproducible Code

_No response_

### Expected Results

All attributes within the class are documented

### Actual Results

Attributes such as `guild` are missing from the documentation

### Intents

N/A

### System Information

Latest

### Checklist

### Additional Context

_No response_

xXenvy:
## Summary
Closes #1159

## This is a **Code Change**

### DIFF
--- a/nextcord/audit_logs.py
+++ b/nextcord/audit_logs.py
@@ -411,6 +411,8 @@ class AuditLogEntry(Hashable):
 
     Attributes
     ----------
+    guild: :class:`Guild`
+        The guild to which the audit log belongs.
     action: :class:`AuditLogAction`
         The action that was done.
     user: :class:`abc.User`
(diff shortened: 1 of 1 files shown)


EmreTech (approved):

nextcord/audit_logs.py
```diff
@@ -549,6 +548,9 @@ def created_at(self) -> datetime.datetime:
 
     @utils.cached_property
     def target(self) -> AuditTarget:
+        """:class:`Any`: The target that got changed.
```
Skelmis, line 551: Can you modify this from `Any` to the actual return type please?

Skelmis (changes requested):

nextcord/audit_logs.py
xXenvy: Yes, but this union is quite long, I don't think it is necessary.
AuditTarget = `Union[Guild, abc.GuildChannel, Member, User, Role, Invite, Emoji, StageInstance, GuildSticker, Thread, Object, AutoModerationRule, None]`

Skelmis: Can you not just :class:`AuditTarget`

Skelmis: I see, hrmmmm

xXenvy: Yes, so I would leave it as it is

Skelmis (approved):

PR #1189, opened 2024-06-03
feat: poll support

### LINKED ISSUE #1179: Polls integration
### Summary

Integration with discord's new polls feature

The core library

### The Problem

Bots cannot create polls, or determine if a message is a poll.

### The Ideal Solution

Have a `nextcord.Poll` object that repesents a poll, including:
- Poll option objects (with list of Members who voted for that option)
- Ending time
- Ended (bool)
- Result (Optional, which poll object"won")

### The Current Solution

None.

### Additional Context

[Polls FAQ](https://support.discord.com/hc/en-us/articles/22163184112407-Polls-FAQ)
There are currently no docs on Polls.

MaskDuck:
#### please notify me on Discord when you give a review, `@maskduck`, preferably in the bikeshedding channel of the support server, it's unlikely I'll ever notice GitHub, thank you for reading
add poll support
fix #1179

still in draft. however you can start reviewing code if you want :)
code untested. this will work, or won't. depends on whether I can write good code or not, generally no

add `Poll`, `PollMedia`, `PollAnswer`, `PollAnswerCount`, `PollResults` in polls.py
add types/poll.py
add AnswerVotersIterator built on top of a similar iterator named reactioniterator

regarding #950 : after that PR is merged there may need to be some sort of work regarding AnswerVotersIterator

## This is a **Code Change**


nextcord/abc.py
```diff
@@ -1430,6 +1436,8 @@ async def send(
             Whether to suppress embeds on this message.
 
             .. versionadded:: 2.4
+        poll: Optional[:class:`PollCreateRequest`]
+            The poll to send with this message.
```
xXenvy, line 1441: Missing `versionadded`

nextcord/message.py
```diff
@@ -737,6 +738,8 @@ class Message(Hashable):
         The guild that the message belongs to, if applicable.
     interaction: Optional[:class:`MessageInteraction`]
         The interaction data of a message, if applicable.
+    poll: Optional[:class:`Poll`]
+        The poll included in a message, if applicable.
```
xXenvy, line 742: Missing `versionadded`

nextcord/polls.py
```diff
@@ -0,0 +1,318 @@
...
+
+    def __init__(self, text: str, emoji: Optional[EmojiInputType] = None) -> None:
+        self.text: str = text
+        if emoji:
+            if isinstance(emoji, str):
+                self.emoji: Optional[PartialEmoji | Emoji] = PartialEmoji.from_str(emoji)
+            else:
+                self.emoji: Optional[PartialEmoji | Emoji] = emoji
+        else:
+            self.emoji: Optional[PartialEmoji | Emoji] = emoji  # type: ignore -- pyright is weird?
```
xXenvy, line 69: This can be simplified to
```py
self.emoji: Optional[PartialEmoji | Emoji] = emoji
if isinstance(emoji, str):
    self.emoji:  Optional[PartialEmoji | Emoji] = PartialEmoji.from_str(emoji)
```

nextcord/polls.py
```diff
@@ -0,0 +1,318 @@
...
+
+    def to_dict(self) -> PollAnswerPayload:
+        payload: PollAnswerPayload = {"poll_media": self.poll_media.to_dict()}
+        if self.answer_id:
+            payload["answer_id"] = self.answer_id
+
+        return payload
+
+    @classmethod
+    def from_dict(cls, data: PollAnswerPayload) -> PollAnswer:
```
xXenvy, line 117: Use `Self` typehint
```py
from typing_extensions import Self
def from_dict(cls, data: PollAnswerPayload) -> Self
```

nextcord/polls.py
```diff
@@ -0,0 +1,318 @@
...
+
+    def to_dict(self) -> PollMediaPayload:
+        payload: PollMediaPayload = {"text": self.text}
+        if self.emoji:
+            payload["emoji"] = resolve_emoji(self.emoji)
+
+        return payload
+
+    @classmethod
+    def from_dict(cls, data: PollMediaPayload) -> PollMedia:
```
xXenvy, line 79: Use `Self` typehint

nextcord/polls.py
```diff
@@ -0,0 +1,370 @@
...
+
+class PollAnswer:
+    """
+    A choice to answer in a poll.
+
+    ..versionadded:: 3.0
+
+    Attributes
+    ----------
+    answer_id: Optional[:class:`str`]
```
xXenvy, line 98: Here it is `Optional[str]`, but in the constructor you pass `Optional[int]` ?

nextcord/polls.py
```diff
@@ -0,0 +1,370 @@
...
+    @classmethod
+    def from_dict(cls, data: PollAnswerPayload) -> Self:
+        # data['answer_id'] is always sent as part of Discord's HTTP/Gateway
+        return cls(answer_id=data["answer_id"], poll_media=PollMedia.from_dict(data["poll_media"]))  # type: ignore[reportTypedDictNotRequiredAccess]
+
+
+class PollAnswerCount:
+    """The answer count for an answer in a poll.
+
+    .. versionadded:: 3.0
```
xXenvy, line 129: Same here. Add space so that it does not overlap.

nextcord/polls.py
```diff
@@ -0,0 +1,370 @@
...
+    Attributes
+    ----------
+    id: :class:`int`
+        The ID of the answer in question.
+    me_voted: :class:`int`
```
xXenvy, line 134: `me_voted` should be :class: bool.

nextcord/polls.py
```diff
@@ -0,0 +1,370 @@
...
+    def to_dict(self) -> PollResultsPayload:
+        payload: PollResultsPayload = {
+            "is_finalized": self.is_finalized,
+            "answer_counts": [answer_count.to_dict() for answer_count in self.answer_counts],
+        }
+        return payload
+
+
+class PollCreateRequest:
+    __slots__ = (
```
xXenvy, line 237: Slots should be located under docstrings.

nextcord/polls.py
```diff
@@ -0,0 +1,370 @@
...
+        This function returns the class instance to allow for fluent-style chaining.
+
+        Parameters
+        ----------
+        text: :class:`str`
+            The text description of the choice.
+        emoji: Optional[Union[:class:`PartialEmoji`, :class:`Emoji`, :class:`str`]]
+            The emoji description of the choice.
+
+        Returns
```
xXenvy, line 293: This `Returns` is unnecessary, as far as I know.

nextcord/polls.py
```diff
@@ -0,0 +1,370 @@
...
+        -------
+        The poll create request object.
+        """
+        self.answers.append(PollAnswer(poll_media=PollMedia(text=text, emoji=emoji)))
+        return self
+
+
+class Poll:
+    """A poll object.
+
```
xXenvy, line 303: Missing `versionadded` + too much space

nextcord/polls.py
```diff
@@ -0,0 +1,370 @@
...
+        "_state",
+        "answers",
+        "expiry",
+        "allow_multiselect",
+        "layout_type",
+        "results",
+    )
+
+    def __eq__(self, other) -> bool:
+        return self.message.id == other.message.id
```
xXenvy, line 335: I suggest to compare it like this: https://github.com/MaskDuck/nextcord/blob/01f1850d34fab582f3bcc1b8602a172930da417b/nextcord/mixins.py#L23-L29
Additionally, you can add `__ne__`.

xXenvy (changes requested):
I did not include this in the review, but please add the `__repr__` method to all new classes as well.

nextcord/polls.py
```diff
@@ -0,0 +1,370 @@
...
+    ----------
+    question: :class:`PollMedia`
+        The question of the poll. Current only `text` is supported.
+
+    """
+
+    def __init__(
+        self,
+        question: PollMedia,
+        answers: List[PollAnswer],
```
acco4321, line 263: Wondering if this should default to an empty list, then add options via `.add_answer`? Feels more consistent with other areas (like Embed) to me - happy either way though.

acco4321 (commented):
Just my inexperienced thoughts. Tested changes, they are functional, just feels a bit clunky - especially around the use of PollMedia objects. Not sure how/if to improve.

Skelmis (commented):
Quick look before I eventually test it. Can we also add an example poll bot to `examples`?

nextcord/types/polls.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class PollResults(TypedDict):
+    is_finalized: bool
+    answer_counts: List[PollAnswerCount]
+
+
+class Poll(TypedDict):
+    question: PollMedia
+    answers: List[PollAnswer]
```
xXenvy, line 41: Hmm, wouldn't it be better to use a dataclass with the ``to_dict`` method for http stuff instead of TypedDict?
It would provide a way to retrieve attributes through `.` + Additionally, you can add `__repr__`.

nextcord/polls.py
```diff
@@ -0,0 +1,437 @@
...
+    # https://github.com/nextcord/nextcord/blob/master/nextcord/message.py#L68
+    EmojiInputType = Union[Emoji, PartialEmoji, str]
+
+
+def resolve_emoji(emoji: Union[Emoji, PartialEmoji]) -> PartialEmojiPayload:
+    if isinstance(emoji, Emoji):
+        return emoji._to_partial().to_dict()  # type: ignore
+
+    if isinstance(emoji, PartialEmoji):
+        return emoji.to_dict()  # type: ignore
```
xXenvy, line 41: There is a `cast` for these things.
```suggestion
        return typing.cast(PartialEmojiPayload, emoji.to_dict())
```

nextcord/polls.py
```diff
@@ -0,0 +1,437 @@
...
+    )
+
+    # for reviewers: this typehint type is precedent in the nextcord codebase i.e.
+    # https://github.com/nextcord/nextcord/blob/master/nextcord/message.py#L68
+    EmojiInputType = Union[Emoji, PartialEmoji, str]
+
+
+def resolve_emoji(emoji: Union[Emoji, PartialEmoji]) -> PartialEmojiPayload:
+    if isinstance(emoji, Emoji):
+        return emoji._to_partial().to_dict()  # type: ignore
```
xXenvy, line 38:
```suggestion
        return typing.cast(PartialEmojiPayload, emoji._to_partial().to_dict())
```

nextcord/polls.py
```diff
@@ -0,0 +1,437 @@
...
+        The poll media of the answer.
+    """
+
+    __slots__ = (
+        "answer_id",
+        "poll_media",
+    )
+
+    def __repr__(self) -> str:
+        return f"<PollAnswer answer_id={self.answer_id!r} poll_media={self.poll_media!r}>"
```
xXenvy, line 109: The `__repr__` should be under the constructor. For the PollMedia class too.

nextcord/polls.py
```diff
@@ -0,0 +1,437 @@
...
+        Whether the bot has voted for this option.
+    count: :class:`int`
+        The number of users who have voted for this option.
+    """
+
+    __slots__ = ("id", "me_voted", "count", "poll")
+
+    def __repr__(self) -> str:
+        return (
+            "<PollAnswerCount "
```
xXenvy, line 147: The `__repr__` should be under the constructor.

nextcord/polls.py
```diff
@@ -0,0 +1,437 @@
...
+
+    Attributes
+    ----------
+    id: :class:`int`
+        The ID of the answer in question.
+    me_voted: :class:`bool`
+        Whether the bot has voted for this option.
+    count: :class:`int`
+        The number of users who have voted for this option.
+    """
```
xXenvy, line 141: It seems that you forgot to document the `poll` attribute.

nextcord/polls.py
```diff
@@ -0,0 +1,437 @@
...
+    def __init__(
+        self,
+        question: PollMedia,
+        duration: int,
+        allow_multiselect: bool,
+        layout_type: Optional[PollLayoutType] = MISSING,
+        answers: Optional[List[PollAnswer]] = None,
+    ) -> None:
+        self.question: PollMedia = question
+        if answers:
```
xXenvy, line 306: It may be onelined
```suggestion
        self.answers: List[PollAnswer] = answers or []
```

nextcord/polls.py
```diff
@@ -0,0 +1,437 @@
...
+    __slots__ = (
+        "question",
+        "answers",
+        "duration",
+        "allow_multiselect",
+        "layout_type",
+    )
+
+    def __repr__(self) -> str:
+        return (
```
xXenvy, line 287: Move `__repr__` under the constructor.

nextcord/polls.py
```diff
@@ -0,0 +1,437 @@
...
+            f"allow_multiselect={self.allow_multiselect!r} "
+            f"layout_type={self.layout_type!r} "
+            f"results={self.results!r}"
+            ">"
+        )
+
+    def __eq__(self, other) -> bool:
+        return isinstance(other, self.__class__) and other.message.id == self.message.id
+
+    def __ne__(self, other) -> bool:
```
xXenvy, line 390: Move all special methods to under the constuctor.

nextcord/polls.py
```diff
@@ -0,0 +1,437 @@
...
+            "question": self.question.to_dict(),
+            "answers": [answer.to_dict() for answer in self.answers],
+            "duration": self.duration,
+            "allow_multiselect": self.allow_multiselect,
+        }
+        if self.layout_type:
+            payload["layout_type"] = self.layout_type.value
+        return payload
+
+    def add_answer(self, text: str, emoji: Optional[EmojiInputType] = None) -> Self:
```
xXenvy, line 325: Im not sure, why this returns `Self`? It may be `None` here.

nextcord/polls.py
```diff
@@ -0,0 +1,437 @@
...
+            async for voter in count.voters():
+                await channel.send(f"{voter} has voted with choice number {count.id}!")
+
+        Flattening into a list ::
+
+            voters = await count.voters().flatten()
+            # voters is now a list of User...
+            await channel.send("Those people have voted with choice number {count.id}: {', '.join(voters)}")
+
+        .. versionadded:: 3.0
```
xXenvy, line 183: This `versionadded` is unnecessary since the whole object is new since version 3.0, and you've already specified it above.

nextcord/types/polls.py
```diff
@@ -0,0 +1,45 @@
+from typing import List, Literal, TypedDict
```
xXenvy, line 1:
```suggestion
# SPDX-License-Identifier: MIT

from typing import List, Literal, TypedDict
```

nextcord/polls.py
```diff
@@ -0,0 +1,437 @@
...
+
+    """
+
+    __slots__ = (
+        "text",
+        "emoji",
+    )
+
+    def __repr__(self) -> str:
+        return f"<PollMedia text={self.text} emoji={self.emoji}>"
```
xXenvy, line 67: Move `__repr__`  to be under the constructor.

xXenvy (changes requested):

nextcord/polls.py
xXenvy: Yes, I am confused about why you are doing this. Why does the method that adds `Answer` return itself? You may return None.

xXenvy: Aaah, I didn't notice, although I don't like it anyway xD

xXenvy (approved):
Untested

docs/api.rst
```diff
@@ -2815,6 +2843,54 @@ AutoModerationActionExecution
...
+
+.. autoclass:: PollAnswer()
+   :members:
+
+PollCreateRequest
+~~~~~~~~~~~~~~~~~
+
+.. attributetable:: PollCreateRequest
+
+.. autoclass:: PollCreateRequest()
```
ooliver1, line 2875: Since the below example shows you constructing a `PollCreateRequest`, this shouldn't have the trailing `()` in order to show the signature in the docs. This shouldn't be included for any user-instantiated class.

nextcord/abc.py
```diff
@@ -1508,6 +1518,8 @@ async def send(
         if file is not None and files is not None:
             raise InvalidArgument("Cannot pass both file and files parameter to send()")
 
+        poll_payload = poll.to_dict() if poll else None
```
ooliver1, line 1521: Could this instead be done like the other `*_payload`s? Like `sticker_payload`/`embed_payload`.

nextcord/polls.py
```diff
@@ -0,0 +1,439 @@
...
+
+    @classmethod
+    def from_dict(cls, data: PollMediaPayload) -> Self:
+        emoji = PartialEmoji.from_dict(data["emoji"]) if "emoji" in data else None
+
+        return cls(text=data["text"], emoji=emoji)
+
+
+class PollAnswer:
+    """
```
ooliver1, line 90: Could these docstrings be like `"""So` just like `PollAnswerCount`, in line with the other multiline docstrings.

nextcord/polls.py
```diff
@@ -0,0 +1,439 @@
...
+            "id": self.id,
+            "me_voted": self.me_voted,
+            "count": self.count,
+        }
+        return payload
+
+
+class PollResults:
+    """The result of the poll.
+    This contains the numbers of vote for each answer.
```
ooliver1, line 212: I know it's quite pedantic, but multiline docstrings have a blank line between the summary and description
```python
"""Summary goes here.

Description here.
"""
```

nextcord/state.py
```diff
@@ -2368,6 +2368,54 @@ def parse_guild_audit_log_entry_create(self, data) -> None:
...
+            if user:
+                self.dispatch(
+                    "message_poll_vote_add",
+                    message,
+                    answer_id,
+                    user,
+                )
+            else:
+                _log.debug(
+                    "message_poll_vote_add was not dispatched because the message (%r) and/or user (%r) are None ",
```
ooliver1, line 2385: Just wondering what the trailing whitespace in these logs are for

ooliver1 (changes requested):

xXenvy (changes requested):
I noticed now that you probably need to update this classmethod:
https://github.com/nextcord/nextcord/blob/ea6606db08e362ffd2df9fa74466cbaa333abd87/nextcord/permissions.py#L182-L195

nextcord/polls.py
```diff
@@ -0,0 +1,448 @@
...
+    def to_dict(self) -> PollAnswerPayload:
+        payload: PollAnswerPayload = {"poll_media": self.poll_media.to_dict()}
+        if self.answer_id:
+            payload["answer_id"] = self.answer_id
+
+        return payload
+
+    @classmethod
+    def from_dict(cls, data: PollAnswerPayload) -> Self:
+        # data['answer_id'] is always sent as part of Discord's HTTP/Gateway
```
EmreTech, line 131: I think an alternative to get around this problem is to make another typed dict specifically for poll answer payloads from Discord API requests.

EmmmaTech (changes requested):

nextcord/polls.py
```diff
@@ -0,0 +1,446 @@
...
+            f"answers={self.answers!r} "
+            f"expired={self.expired!r}"
+            ">"
+        )
+
+    def __eq__(self, other) -> bool:
+        return isinstance(other, self.__class__) and other.message.id == self.message.id
+
+    def __ne__(self, other) -> bool:
+        return isinstance(other, self.__class__) and other.message.id != self.message.id
```
xXenvy, line 412: I'm not sure if this method is necessary at all. I don't have time to test it, but it is mentioned in the [python docs](https://docs.python.org/3/reference/datamodel.html#object.__ne__)
**tldr:** For `__ne__()`, by default it delegates to `__eq__()` and inverts the result unless it is NotImplemented.

However, if we want to leave it for clarity, it can be simplified to:
```suggestion
        return not self.__eq__(other)
```

nextcord/polls.py
```diff
@@ -0,0 +1,446 @@
...
+        Flattening into a list ::
+
+            voters = await count.voters().flatten()
+            # voters is now a list of User...
+            await channel.send("Those people have voted with choice number {count.id}: {', '.join(voters)}")
+
+
+        Parameters
+        ----------
+
```
xXenvy, line 197: Unnecessary space
```suggestion

```

xXenvy (changes requested):
Otherwise looks good

nextcord/polls.py
ooliver1: The result of this method is incorrect too, it returns `False` for "not equal" -> equal, if it is a different type.

xXenvy: What? Returns `True` if the `equal` method returns `False`

ooliver1: I am not on about your suggestion, but about the code this is on. `        return isinstance(other, self.__class__) and other.message.id == self.message.id` is correct for `__eq__` as it must be the same type and same id, but `        return isinstance(other, self.__class__) and other.message.id != self.message.id` is incorrect for `__ne__` as it returns `False` if it is not the same type, meaning `poll != user` would be `False` despite them being different things.

xXenvy: That's right too, I hadn't noticed that

nextcord/polls.py
EmmmaTech: I don't really think there's an issue with that, since `to_dict` and `from_dict` are used in different situations.

nextcord/polls.py
```diff
@@ -0,0 +1,442 @@
...
+    """A poll create request.
+
+    You must use this to create a poll.
+
+    .. versionadded:: 3.0
+
+    Attributes
+    ----------
+    question: :class:`PollMedia`
+        The question of the poll. Current only `text` is supported.
```
EmmmaTech, line 276:
```suggestion
        The question of the poll. Currently only `text` is supported.
```

docs/api.rst
```diff
@@ -1258,6 +1258,32 @@ to handle it, which defaults to print a traceback and ignoring the exception.
...
+
+   .. versionadded:: 3.0
+
+    :param message: The message of the poll which has been voted on.
+    :type message: :class:`Message`
+    :param answer_id: The ID of the answer which had a vote removed.
+    :type answer_id: :class:`int`
+    :param user: The user who removed their vote on the poll.
+    :type user: :class:`User`
 .. _discord-api-utils:
```
EmmmaTech, line 1287:
```suggestion
    :type user: :class:`User`

.. _discord-api-utils:
```

docs/api.rst
```diff
@@ -1418,6 +1444,8 @@ of :class:`enum.Enum`.
 .. autoclass:: AuditLogAction
     :members:
 
+.. autoclass:: PollLayoutType
+    :members:
 Async Iterator
```
EmmmaTech, line 1449:
```suggestion
    :members:

Async Iterator
```

examples/polls.py
```diff
@@ -0,0 +1,22 @@
...
+@bot.command()
+async def send_poll(ctx: commands.Context):
+    poll = PollCreateRequest(
+        question=PollMedia(text="What is your favourite subject?"),
+        allow_multiselect=True,
+        duration=24,
+    )
+    poll.add_answer("Literature", "\U0001f4d6")
+    poll.add_answer("Maths", "\U0001f522")
+    poll.add_answer("Science", "\U0001f9ea")
```
EmmmaTech, line 18: Maybe add comments stating which emojis these are supposed to be?

nextcord/polls.py
```diff
@@ -0,0 +1,442 @@
...
+        Poll as PollData,
+        PollAnswer as PollAnswerPayload,
+        PollAnswerCount as PollAnswerCountPayload,
+        PollCreateRequest as PollCreateRequestPayload,
+        PollMedia as PollMediaPayload,
+        PollResults as PollResultsPayload,
+    )
+
+    # for reviewers: this typehint type is precedent in the nextcord codebase i.e.
+    # https://github.com/nextcord/nextcord/blob/master/nextcord/message.py#L68
```
EmmmaTech, line 32: Remember to remove this ;)

EmmmaTech (commented):
Few minor nitpicks, otherwise looks good 👍

nextcord/polls.py
```diff
@@ -0,0 +1,440 @@
...
+            f"question={self.question!r} "
+            f"answers={self.answers!r} "
+            f"expired={self.expired!r}"
+            ">"
+        )
+
+    def __eq__(self, other) -> bool:
+        return isinstance(other, self.__class__) and other.message.id == self.message.id
+
+    async def expire(self) -> Message:
```
alentoghostflame, line 408: I'm not entirely happy with this method name. It isn't expiring, you're manually ending it. Given that the endpoint is named "End Poll", I think `end` would be a better name for this?

alentoghostflame (changes requested):
Pretty good, only a few issues IMO.

nextcord/polls.py
```diff
@@ -0,0 +1,425 @@
...
+
+    def to_dict(self) -> PollResultsPayload:
+        payload: PollResultsPayload = {
+            "is_finalized": self.is_finalized,
+            "answer_counts": [answer_count.to_dict() for answer_count in self.answer_counts],
+        }
+        return payload
+
+
+class PollCreateRequest:
```
alentoghostflame, line 261: `PollCreateRequest` and `Poll` being seperate objects is kinda annoying, but I get it. If there is a clean way to merge them that would also be somewhat future proof that would be great, if not then ignore this.

nextcord/polls.py
```diff
@@ -0,0 +1,425 @@
...
+        "question",
+        "answers",
+        "duration",
+        "allow_multiselect",
+        "layout_type",
+    )
+
+    def __init__(
+        self,
+        question: PollMedia,
```
alentoghostflame, line 293: IMO, `question` should accept both a `str` and a `PollMedia`, and converts it to a `PollMedia` if it isn't already a `PollMedia` object.

alentoghostflame (changes requested):
Also, `interaction.response.send_message(poll=poll)` needs to be a thing, along with `interaction.followup.send()` if possible and `interaction.send()`.

I will more fully test and analyze the code once poll creation and expiration is fully possible to do within slash commands.

nextcord/types/polls.py
```diff
@@ -0,0 +1,51 @@
...
+
+class PollResults(TypedDict):
+    is_finalized: bool
+    answer_counts: List[PollAnswerCount]
+
+
+class Poll(TypedDict):
+    question: PollMedia
+    answers: List[DAPIPollAnswer]
+    expiry: str
```
sdmway, line 48: Should be optional
image

nextcord/state.py
```diff
@@ -2368,6 +2368,40 @@ def parse_guild_audit_log_entry_create(self, data) -> None:
...
+            user = self.get_user(int(data["user_id"]))
+            self.dispatch(
+                "message_poll_vote_remove",
+                message,
+                answer_id,
+                user,
+            )
+        else:
+            _log.debug(
+                "message_poll_vote_remove was not dispatched because the message (%r) if None ",
```
sdmway, line 2401: typo, if instead of is

nextcord/polls.py
```diff
@@ -0,0 +1,425 @@
...
+
+
+def resolve_emoji(emoji: Union[Emoji, PartialEmoji]) -> PartialEmojiPayload:
+    if isinstance(emoji, Emoji):
+        return cast(PartialEmojiPayload, emoji._to_partial().to_dict())
+
+    if isinstance(emoji, PartialEmoji):
+        return cast(PartialEmojiPayload, emoji.to_dict())
+
+    raise InvalidArgument(f"Cannot parse `{emoji}` into an emoji to send.")
```
teaishealthy, line 53: isn't this unreachable? type of emoji should be `Never` here

nextcord/polls.py
```diff
@@ -0,0 +1,425 @@
...
+        and meet the :class:`abc.Snowflake` abc.
+
+        Examples
+        --------
+        Usage ::
+
+            # I do not actually recommend doing this.
+
+            async for voter in count.voters():
+                await channel.send(f"{voter} has voted with choice number {count.id}!")
```
teaishealthy, line 185:
```suggestion
            async for voter in count.voters():
                print(f"{voter} has voted with choice number {count.id}!")
```

nextcord/polls.py
```diff
@@ -0,0 +1,425 @@
...
+
+        Parameters
+        ----------
+        limit: Optional[:class:`int`]
+            The maximum number of results to return.
+            If not provided, returns all the users who
+            voted to the choice.
```
teaishealthy, line 192:
```suggestion
            If not provided, return all users who voted for the choice.
```

nextcord/polls.py
```diff
@@ -0,0 +1,425 @@
...
+    .. versionadded:: 3.0
+
+    Attributes
+    ----------
+    is_finalized: :class:`bool`
+        Whether the votes has been precisely counted.
+    answer_counts: List[:class:`PollAnswerCount`]
+        The counts for each answer.
+    poll: :class:`Poll`
+        The poll this poll result represents.
```
teaishealthy, line 226:
```suggestion
        The poll associated with this result.
```
clearer imo

nextcord/polls.py
```diff
@@ -0,0 +1,425 @@
...
+
+    You must use this to create a poll.
+
+    .. versionadded:: 3.0
+
+    Attributes
+    ----------
+    question: Union[:class:`str`, :class:`PollMedia`]
+        The question of the poll. Currently only `text` is supported.
+        If a :class:`str` was passed, it automatically converts to a :class:`PollMedia`.
```
teaishealthy, line 269:
```suggestion
        If a :class:`str` was passed, it's automatically converted to a :class:`PollMedia`.
```

nextcord/polls.py
```diff
@@ -0,0 +1,425 @@
...
+    duration: Optional[:class:`int`]
+        The number of hours this poll should be open for. Defaults to ``24``. Max ``768`` (32 days).
+    allow_multiselect: Optional[:class:`bool`]
+        Whether voters can select more than one choice. Defaults to False.
+    answers: Optional[List[:class:`PollAnswer`]]
+        The answers to the poll.
+        If omitted then it must be added later through :func:`PollCreateRequest.add_answer`.
```
teaishealthy, line 276:
```suggestion
        If omitted, answers must be added later with :func:`PollCreateRequest.add_answer`.
```

nextcord/polls.py
```diff
@@ -0,0 +1,425 @@
...
+    answers: List[:class:`PollAnswer`]
+        The answers of the poll.
+    expiry: :class:`datetime.datetime`
+        The expiry of the poll.
+    allow_multiselect: :class:`bool`
+        Whether this poll allow people to select more than 1 option.
+    layout_type: :class:`PollLayoutType`
+        The layout type of the poll.
+    results: Optional[:class:`PollResults`]
+        The results of the poll.
```
teaishealthy, line 363:
```suggestion
    message: :class:`Message`
        The message this poll belongs to.
    question: :class:`PollMedia`
        The question being asked in this poll.
    answers: List[:class:`PollAnswer`]
        The answers for this poll.
    expiry: :class:`datetime.datetime`
        The expiry of this poll.
    allow_multiselect: :class:`bool`
        Whether this poll allows users to select more than one option.
    layout_type: :class:`PollLayoutType`
        The layout type of this poll.
    results: Optional[:class:`PollResults`]
        The results of this poll.
```

nextcord/polls.py
```diff
@@ -0,0 +1,425 @@
...
+
+    @classmethod
+    def from_dict(cls, data: PollMediaPayload) -> Self:
+        emoji = PartialEmoji.from_dict(data["emoji"]) if "emoji" in data else None
+
+        return cls(text=data["text"], emoji=emoji)
+
+
+class PollAnswer:
+    """A choice to answer in a poll.
```
teaishealthy, line 101:
```suggestion
    """An answer choice in a poll.
```

teaishealthy (changes requested):

nextcord/iterators.py
```diff
@@ -551,7 +552,6 @@ def get_retrieve():
             before=before.id if before else None,
             after=after.id if after else None,
         )
-
```
ooliver1, line 554: Seems a bit random too.

nextcord/types/polls.py
```diff
@@ -0,0 +1,51 @@
...
+
+class DAPIPollAnswer(TypedDict):
+    answer_id: int
+    poll_media: PollMedia
+
+
+class PollCreateRequest(TypedDict):
+    question: PollMedia
+    answers: List[PollAnswer]
+    duration: int
```
ooliver1, line 29: `duration` and `allow_multiselect` are both `NotRequired`

nextcord/types/polls.py
```diff
@@ -0,0 +1,51 @@
...
+
+from typing_extensions import NotRequired
+
+from .emoji import PartialEmoji
+
+LayoutType = Literal[1]
+
+
+class PollMedia(TypedDict):
+    text: str
```
ooliver1, line 13: This is also `NotRequired`

nextcord/types/polls.py
```diff
@@ -0,0 +1,51 @@
...
+    emoji: NotRequired[PartialEmoji]
+
+
+class PollAnswer(TypedDict):
+    poll_media: PollMedia
+
+
+class DAPIPollAnswer(TypedDict):
+    answer_id: int
+    poll_media: PollMedia
```
ooliver1, line 23: I'd personally have this inherit as they are both on the same table in the discord docs, but just a nitpick.

nextcord/state.py
```diff
@@ -2367,6 +2367,40 @@ def parse_guild_audit_log_entry_create(self, data) -> None:
...
+            user = self.get_user(int(data["user_id"]))
+            self.dispatch(
+                "message_poll_vote_add",
+                message,
+                answer_id,
+                user,
+            )
+        else:
+            _log.debug(
+                "message_poll_vote_add was not dispatched because the message (%r) is None",
```
ooliver1, line 2383: There should be a raw event in this case (both events), as messages aren't cached for too long.

nextcord/polls.py
```diff
@@ -0,0 +1,441 @@
...
+
+
+class PollMedia:
+    """A common object that backs both the question and answer.
+
+    .. versionadded:: 3.0
+
+    Attributes
+    ----------
+    text: Optional[:class:`str`]
```
ooliver1, line 63: This doesn't seem to be typed as `Optional` in code. It seems to be the case it is required for now but may not be in the future, so I think `Optional` here and let discord validation tell users otherwise.
image

nextcord/polls.py
```diff
@@ -0,0 +1,441 @@
...
+class PollAnswer:
+    """An answer choice in a poll.
+
+    .. versionadded:: 3.0
+
+    Attributes
+    ----------
+    answer_id: Optional[:class:`int`]
+        The answer ID of the answer.
+    media: Optional[:class:`PollMedia`]
```
ooliver1, line 109: This seems to be required, both in code and in Discord docs.

nextcord/polls.py
```diff
@@ -0,0 +1,441 @@
...
+
+
+class PollAnswer:
+    """An answer choice in a poll.
+
+    .. versionadded:: 3.0
+
+    Attributes
+    ----------
+    answer_id: Optional[:class:`int`]
```
ooliver1, line 107: There should be a mention as to why this would be `None`.

nextcord/polls.py
```diff
@@ -0,0 +1,441 @@
...
+        "layout_type",
+    )
+
+    def __init__(
+        self,
+        question: Union[str, PollMedia],
+        duration: int = 24,
+        allow_multiselect: bool = False,
+        layout_type: Optional[PollLayoutType] = MISSING,
+        answers: Optional[List[PollAnswer]] = None,
```
ooliver1, line 307: Creating a `PollAnswer` in a list of already having to make `PollAnswers` is quite tedious, here (or in a user instantiate-able `Poll`) there should be the possibility of using `str` for simple results.

```python
answers=[
    PollAnswer(poll_media=PollMedia(text="one")),
    PollAnswer(poll_media=PollMedia(text="two")),
    PollAnswer(poll_media=PollMedia(text="three")),
],
```

nextcord/polls.py
```diff
@@ -0,0 +1,441 @@
...
+        return cast(PartialEmojiPayload, emoji._to_partial().to_dict())
+
+    if isinstance(emoji, PartialEmoji):
+        return cast(PartialEmojiPayload, emoji.to_dict())
+
+    raise InvalidArgument(f"Cannot parse `{emoji}` into an emoji to send.")
+
+
+class PollMedia:
+    """A common object that backs both the question and answer.
```
ooliver1, line 57: I'm not too sure about this description, doesn't seem to describe the actual use very well.

nextcord/polls.py
```diff
@@ -0,0 +1,441 @@
...
+        payload: PollAnswerCountPayload = {
+            "id": self.id,
+            "me_voted": self.me_voted,
+            "count": self.count,
+        }
+        return payload
+
+
+class PollResults:
+    """The result of the poll.
```
ooliver1, line 225:
```suggestion
    """The result of a poll.
```
at least, or link to `Poll`

ooliver1 (changes requested):
I would advise improving the title documentation for the model classes, "A poll object" could be "Represents a poll in Discord." for example, as the poll class is not a poll object itself.

Please don't review about `versionadded` here, I am aware and it'll be updated by me, the author, or another maintainer, when it is actually going to be merged.

nextcord/polls.py
```diff
@@ -0,0 +1,370 @@
...
+    """
+    A choice to answer in a poll.
+
+    ..versionadded:: 3.0
+
+    Attributes
+    ----------
+    answer_id: Optional[:class:`str`]
+        The answer ID of the answer.
+    poll_media: Optional[:class:`PollMedia`]
```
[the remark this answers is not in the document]
ooliver1: I'd agree with this, the Discord API itself is a little silly and this is one where the field doesn't need to redundantly mention it is a poll.

nextcord/abc.py
```diff
@@ -1420,6 +1426,10 @@ async def send(
             Whether to suppress embeds on this message.
 
             .. versionadded:: 2.4
+        poll: Optional[:class:`PollCreateRequest`]
```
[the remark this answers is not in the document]
ooliver1: disnake seems to do it
https://github.com/DisnakeDev/disnake/blob/2ab0c535b021dd120b5a6aff67dc1c195b9913ae/disnake/poll.py#L291

nextcord/polls.py
ooliver1: "Represents the data for a poll's question and answers." perhaps, I just think the wording of "backs" is that friendly to understand

nextcord/polls.py
ooliver1: I imagine they're not going to bloat up polls for a while, I think at least accepting strings in the answer list would be the simplest whilst still giving the full capability of expansive polls with the current syntax.

PR #1199, opened 2024-06-29
feat: add burst (super) reactions

spifory:
## Summary

- Implements discord/discord-api-docs#6056 & discord/discord-api-docs#6375

Tested with: https://paste.nextcord.dev/?id=1725902791782971

image
image
image

I'm still uncertain whether or not bots *can* or cannot send burst reactions, however the docs don't document anything about it

## This is a **Code Change**


nextcord/reaction.py
```diff
@@ -2,18 +2,42 @@
...
-from typing import TYPE_CHECKING, Any, Optional, Union
+from typing import TYPE_CHECKING, Any, List, Optional, Union
 
 from .iterators import ReactionIterator
 
 __all__ = ("Reaction",)
 
 if TYPE_CHECKING:
     from .abc import Snowflake
+    from .colour import Color
```
ooliver1, line 13: From a glance, looks like this is in type checking but used below for `burst_colors` (don't we use british english). For that anyway, it could probably be a (cached) property to save memory for the majority of people that do not want the burst reaction colours.

spifory: Oh okay thanks, will do that
Should I replace `burst_colors` with colours to make an alias?

nextcord/reaction.py
```diff
@@ -2,18 +2,43 @@
...
+        The amount of normal reactions.
+
+    .. versionadded:: 3.0
+    """
+
+    __slots__ = ("burst", "normal")
+
+    def __init__(self, data: ReactionCountDetailsPayload) -> None:
+        self.burst = data.get("burst")
+        self.normal = data.get("normal")
```
teaishealthy, line 41: types?

teaishealthy:
```suggestion
        self.burst: int = data.get("burst")
        self.normal: int = data.get("normal")
```

teaishealthy (changes requested):

nextcord/raw_models.py
```diff
@@ -168,7 +169,17 @@ class RawReactionActionEvent(_RawReprMixin):
...
+    __slots__ = (
+        "message_id",
+        "user_id",
+        "channel_id",
+        "guild_id",
+        "emoji",
+        "event_type",
+        "member",
+        "burst",
+        "burst_colors",
```
EmmmaTech, line 181: There seems to be a missing [type attribute](https://discord.com/developers/docs/resources/message#get-reactions-reaction-types) here, although I'm not sure why it exists since the `burst` boolean is also there...

nextcord/reaction.py
```diff
@@ -92,6 +139,22 @@ def __str__(self) -> str:
     def __repr__(self) -> str:
         return f"<Reaction emoji={self.emoji!r} me={self.me} count={self.count}>"
 
+    @cached_property
+    def burst_colours(self) -> List[Colour]:
+        """List[:class:`Colour`]: The HEX colors used for a burst reaction.
```
EmmmaTech, line 144: I'm confused on whether we're sticking with American or British English here. Also why is hex in uppercase?

EmmmaTech (changes requested):

nextcord/reaction.py
alentoghostflame: Traditionally, there is a function with the British English name that holds the functionality and an American alias for it. An example would be [User.colour and User.color](https://github.com/nextcord/nextcord/blob/master/nextcord/user.py#L215-L231).

EmmmaTech: I'm aware of this, but the problem is American English spelling is used in this docstring. In the example you brought up, the proper spellings for the dialects are used.

spifory: My mistake British English it is. Also "HEX" is capitalised for no reason in particular really, AFAIK you can write it capitalised or uncapitalised

nextcord/reaction.py
```diff
@@ -13,7 +16,30 @@
...
+    """Represents a reaction's count details.
+
+    Attributes
+    ----------
+    burst: :class:`int`
+        The amount of burst reactions.
+    normal: :class:`int`
+        The amount of normal reactions.
+
+    .. versionadded:: 3.0
```
ooliver1, line 35: This goes before attributes, after any main description.

ooliver1 (changes requested):

EmmmaTech (changes requested):
tested, and for newly created messages with a new reaction, `count_details` seems to be missing when provided from the api. I was able to get my bot to crash a few times due to this.
here's my test code if necessary: https://paste.nextcord.dev/?id=1732326138949174
exception traceback if necessary: https://paste.nextcord.dev/?id=1732326348787843

nextcord/reaction.py
```diff
@@ -13,7 +16,30 @@
...
+        The amount of burst reactions.
+    normal: :class:`int`
+        The amount of normal reactions.
+    """
+
+    __slots__ = ("burst", "normal")
+
+    def __init__(self, data: ReactionCountDetailsPayload) -> None:
+        self.burst: int = data.get("burst")
+        self.normal: int = data.get("normal")
```
teaishealthy, line 42: doesn't have to be .get()

nextcord/reaction.py
```diff
@@ -69,6 +112,12 @@ def __init__(
         )
         self.count: int = data.get("count", 1)
         self.me: bool = data.get("me")
+        self.me_burst: bool = data.get("me_burst")
+        self.count_details: ReactionCountDetails = ReactionCountDetails(
+            data=data.get("count_details")
```
teaishealthy, line 117: same here

teaishealthy (commented):
okay™

teaishealthy (approved):
untested

nextcord/types/raw_models.py
```diff
@@ -35,6 +36,9 @@ class ReactionActionEvent(TypedDict):
     emoji: PartialEmoji
     guild_id: NotRequired[Snowflake]
     member: NotRequired[Member]
+    burst: bool
+    burst_colors: NotRequired[List[str]]
+    type: ReactionType
```
ooliver1, line 42: The type here isn't an enum, its a `Literal[0, 1]`, so needs `ReactionType = Literal[0, 1]` near the top of this file instead of an enums import

nextcord/iterators.py
```diff
@@ -9,17 +9,18 @@
...
+    "archived_thread_iterator",
     "audit_log_iterator",
+    "ban_iterator",
     "guild_iterator",
+    "history_iterator",
     "member_iterator",
-    "archived_thread_iterator",
+    "reaction_iterator",
     "scheduled_event_iterator",
     "scheduled_event_user_iterator",
```
ooliver1, line 25: I'd leave these alone and leave sorting to a future PR (enabling the ruff rule for it ofc)

nextcord/raw_models.py
```diff
@@ -148,6 +150,10 @@ class RawReactionActionEvent(_RawReprMixin):
         The member who added the reaction. Only available if ``event_type`` is ``REACTION_ADD`` and the reaction is inside a guild.
 
         .. versionadded:: 1.3
+    burst: :class:`bool`
+        If the reaction is a super/burst reaction.
+
+        .. versionadded:: 3.2
```
ooliver1, line 159: + document the `type` attribute

nextcord/http.py
```diff
@@ -1612,6 +1612,7 @@ def get_reaction_users(
         emoji: str,
         limit: int,
         after: Optional[Snowflake] = None,
+        type: Optional[ReactionType] = None,
```
ooliver1, line 1622: It seems like you can also pass `type` as a query param to `add_reaction`

ooliver1 (changes requested):
Otherwise, tested

nextcord/iterators.py
spifory: Oh my fault     I likely hit save on the file back then and it formatted that too

Issue #1076, opened 2023-07-02
Type nextcord.Embed method parameters more specifically


EmmmaTech:
### Summary

The methods for `nextcord.Embed` should have properly typed parameters.

### What is the feature request for?

The core library

### The Problem

Currently, many of the parameters in `nextcord.Embed` methods that are supposed to be `str` are typed with `Any`. This is not ideal because static type checkers will allow you to provide any object, whether it's appropriate for the method. It's like that so the parameters can support any object then be stringified later, but it could lead to weird problems if not used correctly.

Here's an example of this:
https://github.com/nextcord/nextcord/blob/dcadc7ea6d3e29c581cd49c2bbce3936a8804983/nextcord/embeds.py#L324

### The Ideal Solution

`nextcord.Embed` methods should be typed more precisely to only allow objects that support conversion to `str`. For example, in our example method `set_footer`, if an object that supports conversion to a string (`__str__`) is provided, then the type checker will be happy. Otherwise, it will throw an error.

This can be accomplished by typing the parameter with a protocol that lists the supported conversion method. Unfortunately, Python does not have a [built-in protocol](https://docs.python.org/3/library/typing.html#protocols) for conversion to a string because technically every object can be converted to a string. We will have to make our own. 

### Additional Context

As [discussed on the Discord Server](https://discord.com/channels/881118111967883295/881124672056467456/1125003163024502784), this issue will be for v3.

Jodenee:
I'm not sure if a protocol would fit in this situation, as you know every class in python can be converted into a str with or without manually implementing the __str__ magic method as python will just default to __repr__ if __str__ is not implemented, and that's where the issue is. It's impossible to tell if it is implemented or not with protocols as to protocols __str__ is always defined.

But I am aware of one way of achieving this, what we can do instead is make use of an ABC with just the __str__ magic method as an abstract method, but this comes with its own drawbacks such as, forcing the ABC to be inherited by the user's classes and it not supporting literal strings, so the typing would need to be a union of the ABC and str.

If you are fine with me using the above method, I wouldn't mind submitting a pr to resolve your issue.

Jodenee:
Forgot to mention this in my last message but the user wont be able to pass raw numbers and such as arguments. So they'll need to convert those values into strings using `str()`. This is because primitive types don't inherit the ABC so the type checker wont know if `__str__` is implemented or not, just thought that was worth mentioning in case this isn't the solution you wanted.

[...]

PR #1201, opened 2024-07-09
typing(embeds): Retype embed methods to accept str instead of any

Jodenee:
## Summary

Retypes embed methods to use `str` instead of `any` to improve typing and better reflect the docs.

Closes issue #1076.

## This is a **Code Change**

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -2222,6 +2222,13 @@ Connectable
 
 .. _discord_api_models:
 
+SupportsStr
+~~~~~~~~~~~
+
+.. attributetable:: nextcord.abc.SupportsStr
+
+.. autoclass:: nextcord.abc.SupportsStr()
+
 Discord Models
 --------------
 
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -2222,13 +2222,6 @@ Connectable
 
 .. _discord_api_models:
 
-SupportsStr
-~~~~~~~~~~~
-
-.. attributetable:: nextcord.abc.SupportsStr
-
-.. autoclass:: nextcord.abc.SupportsStr()
-
 Discord Models
 --------------
 
--- a/examples/custom_interaction.py
+++ b/examples/custom_interaction.py
@@ -8,7 +8,7 @@ class MyInteraction(nextcord.Interaction):
             title=title,
             description=description,
         )
-        embed.set_author(name=self.user.display_name, icon_url=self.user.display_avatar)
+        embed.set_author(name=self.user.display_name, icon_url=self.user.display_avatar.url)
         await self.send(embed=embed)
 
 
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -5,6 +5,7 @@ from __future__ import annotations
 import asyncio
 import contextlib
 import copy
+from abc import ABC, abstractmethod
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -1795,3 +1796,16 @@ class Connectable(Protocol):
             raise  # re-raise
 
         return voice
+
+
+class SupportsStr(ABC):
+    """An ABC with one abstract magic method ``__str__``.
+
+    This ABC is meant to be implemented by the user. This will allow the type checker to see if a class has implemented the ``__str__`` magic method.
+    """
+
+    __slots__ = ()
+
+    @abstractmethod
+    def __str__(self) -> str:
+        raise NotImplementedError
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -5,7 +5,6 @@ from __future__ import annotations
 import asyncio
 import contextlib
 import copy
-from abc import ABC, abstractmethod
 from typing import (
     TYPE_CHECKING,
     Any,
@@ -1796,16 +1795,3 @@ class Connectable(Protocol):
             raise  # re-raise
 
         return voice
-
-
-class SupportsStr(ABC):
-    """An ABC with one abstract magic method ``__str__``.
-
-    This ABC is meant to be implemented by the user. This will allow the type checker to see if a class has implemented the ``__str__`` magic method.
-    """
-
-    __slots__ = ()
-
-    @abstractmethod
-    def __str__(self) -> str:
-        raise NotImplementedError
--- a/nextcord/embeds.py
+++ b/nextcord/embeds.py
@@ -12,8 +12,6 @@ from .colour import Colour
 if TYPE_CHECKING:
     from typing_extensions import Self
 
-    from .abc import SupportsStr
-
 __all__ = ("Embed",)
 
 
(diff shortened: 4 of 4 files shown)


ooliver1 (changes requested):
I do not see much point in this. If the user is using a type checker, they would already see it only accepts `str`. If they are not using a type checker, not noticing it is `str` would also mean they do not notice `SupportsStr`. Having to inherit an ABC removes almost all uses of a non-str argument that should already be `str` anyway.

teaishealthy:
Because `object` implements `__str__`, all classes also implement `__str__`. `SupportsStr` is effectively the same as `Any` (actually `object`, but that doesn't matter here). I don't think there is anything to be gained from this change.

Jodenee:
> Because `object` implements `__str__`, all classes also implement `__str__`

But what about any objects the user creates? `SupportsStr` is meant to allow custom objects to be passed as args. 

Though looking back on it nextcord classes wont be accepted by the type checker as they don't inherit from `SupportsStr`, so I would need to go through every class that does implement `__str__` and make them inherit from `SupportsStr`, but I don't think that solution would be accepted. Unfortunately this is a flaw of this solution as `SupportsStr` needs to be inherited for the type checker to know.

> `SupportsStr` is effectively the same as `Any` (actually `object`, but that doesn't matter here).

I have to disagree with this, `SupportsStr` allows the type checker to see if a class has manually implemented `__str__`, while `Any` allows any value to be passed, which is the problem this pr aims to solve.

I get this change is odd, I myself do not like it as it isn't a very clean solution, but this is the only solution I could come up with to close this issue. If you or anyone else have any ideas on how I can improve this pr I'm all ears, if not then this pr can be closed.

ooliver1:
> For attributes yes, but the methods do not strictly only take strings as they are typed with `Any` unlike what the docs say, so this change is aimed at making the typing more concise.

Then I think they should be typed as `str` instead.,

> While I don't think of this as that big of an issue as the user can just convert any non-str arguments into strings using `str()`, unfortunately this is just how it is, as I'm not aware of any way to fix this flaw.

What I meant was that I could not think of any form of class which would satisfy `SupportsStr` or would be worth inheriting it for, instead of just using `str()` already.

Jodenee:
> Then I think they should be typed as str instead.

> What I meant was that I could not think of any form of class which would satisfy `SupportsStr` or would be worth inheriting it for, instead of just using `str()` already.

I agree, this is probably the best solution, but would doing this still satisfy the issue? If so I'll push those changes.

teaishealthy:
I mistook the `SupportsStr` class for a `typing.Protocol` (where my criticisms apply), my bad. 

But then I don't like this change because of programming ergonomics, à la "... if it quacks like a duck, then it probably is a duck":

Overriding `__str__` is not the only way to influence the output of `__str__`, changing the `__repr__` also changes `__str__`. So every class that *only* implements `__repr__` has to implement a dummy method something like this:
```py
def __str__(self) -> str:
    return repr(self)
```

I think we have 2-3 possible options instead:
1. accept all objects and just call `str()` on them (no changes)
2. to avoid breaking changes, change the typehint to `str` but keep the same logic
3. change the type to `str` and remove the call to `str()`

I think 2 or 3 is the right course of action here, while I'm personally leaning towards 2.

Jodenee:
> I think we have 2-3 possible options instead:
> 1. accept all objects and just call str() on them (no changes)
> 2. to avoid breaking changes, change the typehint to str but keep the same logic
> 3. change the type to str and remove the call to str()
>
> I think 2 or 3 is the right course of action here, while I'm personally leaning towards 2.

Same here, 2 is probably the way to go.

@EmreTech can I get your thoughts on which option would be best? (Apologies for the mention)

EmreTech:
> > I think we have 2-3 possible options instead:
> > 
> > 1. accept all objects and just call str() on them (no changes)
> > 2. to avoid breaking changes, change the typehint to str but keep the same logic
> > 3. change the type to str and remove the call to str()
> > 
> > I think 2 or 3 is the right course of action here, while I'm personally leaning towards 2.
> 
> Same here, 2 is probably the way to go.
> 
> @EmreTech can I get your thoughts on which option would be best? (Apologies for the mention)

I would go with option 3 since we're working on 3.0. Breaking changes don't matter all that much.

Jodenee:
> I would go with option 3 since we're working on 3.0. Breaking changes don't matter all that much.

Alright, I'll begin work on that then.

Jodenee:
Forgot to mention but option 3 was implemented in the above commits, feel free to review.

EmmmaTech (approved):
Looks good :D

alentoghostflame:
Has decent testing been done on all modified functions and things that call them? I wouldn't be surprised if NC internally relies on `str()` being used.

Jodenee:
> Has decent testing been done on all modified functions and things that call them? I wouldn't be surprised if NC internally relies on `str()` being used.

I wasn't aware of any internals calling these functions, I'll look into it though, thanks.

alentoghostflame:
I'm unsure if anything calls them internally, just wanted to check if you looked into that already or not.

Also, the PR summary probably should get updated? I'm not seeing a `SupportsStr` anywhere.

Jodenee:
> Also, the PR summary probably should get updated? I'm not seeing a `SupportsStr` anywhere.

Yeah, I'll get on that soon.

Jodenee:
> Has decent testing been done on all modified functions and things that call them? I wouldn't be surprised if NC internally relies on `str()` being used.

Looked into the internals as you requested, I didn't find any internal code calling these functions so this pr should be fine to merge. If i'm wrong and there are internals calling these functions, please do let me know.

alentoghostflame:
Alright, so I did find one *slight* issue. `/examples/custom_interaction.py` calls `embed.set_author()`, but gives the `icon_url` kwarg an `Asset` object, rather than a string. `Asset` just returns `self._url` when `str()`'d, but with the `str()` removed I don't think that would work anymore.

Super easy fix for people, just add `.url` to the asset being passed in, but that does make this a breaking change. I would like to see someone knowing that this is a breaking change approve this (@EmmmaTech) and the example file I mentioned above be fixed before merging this in.

Is passing an Asset object into `icon_url` common? I know I don't do it, but I also don't use Asset objects very often. Are there other parts of Nextcord with "image url" kwargs that accept Asset objects, and thus this PR would make Embeds non-standard?

EmmmaTech:
I'm aware such a typing change would be breaking

Jodenee:
> Alright, so I did find one _slight_ issue. `/examples/custom_interaction.py` calls `embed.set_author()`, but gives the `icon_url` kwarg an `Asset` object, rather than a string. `Asset` just returns `self._url` when `str()`'d, but with the `str()` removed I don't think that would work anymore.

Thanks for informing me, I'll commit the fix soon.

> Is passing an Asset object into `icon_url` common? I know I don't do it, but I also don't use Asset objects very often. Are there other parts of Nextcord with "image url" kwargs that accept Asset objects, and thus this PR would make Embeds inconsistent with the rest of the lib?

I don't think there are any methods that strictly accept an Asset object for a: "image_url", "url" parameter. From what I know they accept str, so this pr shouldn't make embeds inconsistent.

alentoghostflame (approved):
LGTM!

PR #1204, opened 2024-08-17
fix: NameError within slash command isinstance check

Skelmis:
## Summary

When lazy loading slash commands, `if not isinstance(app_cmd, BaseApplicationCommand):` was called without `BaseApplicationCommand` being defined, as such a `NameError` is raised when this code path is hit. This PR resolves this by doing a local import when the class is required, while still maintaining the type checking import.

## This is a **Code Change**


alentoghostflame:
Can't test right now, is there a reason why you can't move the typechecking import to a normal import so we don't need to do a local import?

Skelmis:
I hadn't tried tbh, just saw it was a typing import and figured i'd leave as is. I can try it out sometime this week hopefully

Skelmis:
@alentoghostflame yea seems to work fine

alentoghostflame (approved):
LGTM. It's a small enough change that I'll ignore the 2-approval requirement.

PR #1206, opened 2024-09-08
feat(guild): add `fetch_role`

spifory:
## Summary

- Implements discord/discord-api-docs#7074

Tested with
```py
from nextcord import Intents, Interaction, Role
from nextcord.ext.commands import Bot

bot = Bot(
    intents=Intents(guilds=True, message_content=True, messages=True),
    default_guild_ids=[...],
)

@bot.slash_command()
async def role(inter: Interaction, role: Role):
    assert inter.guild is not None
    fetched_role = await inter.guild.fetch_role(role.id)
    return await inter.send(repr(fetched_role))

bot.run(...)

```
image

## This is a **Code Change**

### DIFF
--- a/nextcord/guild.py
+++ b/nextcord/guild.py
@@ -2804,6 +2804,32 @@ class Guild(Hashable):
 
         return roles
 
+    async def fetch_role(self, role_id: int, /) -> Role:
+        """|coro|
+
+        Retrieves a :class:`Role` that a guild has.
+
+        .. versionadded:: 3.0
+
+        Parameters
+        ----------
+        role_id: int
+            The ID of the role to fetch.
+
+        Raises
+        ------
+        HTTPException
+            Retrieving the roles failed.
+
+        Returns
+        -------
+        :class:`Role`
+            The role in the guild.
+        """
+        data = await self._state.http.get_role(self.id, role_id)
+
+        return Role(guild=self, state=self._state, data=data)
+
     @overload
     async def create_role(
         self,
--- a/nextcord/guild.py
+++ b/nextcord/guild.py
@@ -2819,7 +2819,9 @@ class Guild(Hashable):
         Raises
         ------
         HTTPException
-            Retrieving the roles failed.
+            Retrieving the role failed.
+        NotFound
+            The role was not found.
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


nextcord/guild.py
```diff
@@ -2804,6 +2804,32 @@ async def fetch_roles(self, *, cache: bool = False) -> List[Role]:
...
+        Parameters
+        ----------
+        role_id: int
+            The ID of the role to fetch.
+
+        Raises
+        ------
+        HTTPException
+            Retrieving the roles failed.
+
```
xXenvy, line 2823: I've tested, it actually throws an exception like this
```suggestion
        Raises
        ------
        HTTPException
            Retrieving the roles failed.
        NotFound
            The role was not found.
```

nextcord/guild.py
```diff
@@ -2804,6 +2804,32 @@ async def fetch_roles(self, *, cache: bool = False) -> List[Role]:
...
+        Returns
+        -------
+        :class:`Role`
+            The role in the guild.
+        """
+        data = await self._state.http.get_role(self.id, role_id)
+
```
xXenvy, line 2830:
```suggestion
```

xXenvy (changes requested):
otherwise looks good

nextcord/guild.py
```diff
@@ -2804,6 +2804,32 @@ async def fetch_roles(self, *, cache: bool = False) -> List[Role]:
 
         return roles
 
+    async def fetch_role(self, role_id: int, /) -> Role:
+        """|coro|
+
+        Retrieves a :class:`Role` that a guild has.
```
ooliver1, line 2810: I understand you're following `fetch_roles` wording but to be picky I would prefer it to be similar to fetching single objects by ID, something like:

```suggestion
        Retrieve a :class:`Role` from this guild by its ID.
```
also matches the imperative form docstrings should have (which isn't true for all yet)

spifory: Since this function under the `Guild` class, is specifying "guild by its ID" not a bit redundant?

ooliver1: "A role the guild has" would also be redundant in that case, you can put what you like as there's no real convention yet, but imperative and descriptive is the idea

xXenvy (approved):

ooliver1 (approved):

PR #843, opened 2022-10-10
feat: track and expose non-persistent views

EmreTech:
## Summary

This PR adds functionality to track and expose non-persistent views to the user. The user can manually add views to be tracked via `Client.add_view` and can retrieve all non-persistent views via `Client.views`.

Closes #14.

## Checklist

- [x] If code changes were made then they have been tested.
  - [x] I have updated the documentation to reflect the changes.
  - [ ] I have run `task pyright` and fixed the relevant issues.
- [ ] This PR fixes an issue.
- [x] This PR adds something new (e.g. new method or parameters).
- [ ] This PR is a breaking change (e.g. methods or parameters removed/renamed)
- [ ] This PR is **not** a code change (e.g. documentation, README, ...)

PR #1228, opened 2024-11-24
fix(client): remove duplication of views in views/all_views

ooliver1:
## Summary

Introduced in #843 and discovered in [this help thread](https://discord.com/channels/881118111967883295/1310305439073828864), the new implementation for retrieving views via the client has duplicated views with multiple components.
This is due to removing the construction of a dictionary internally to make the list unqiue. This PR adds this method back and comments
it to ensure this isn't wrongly removed again.

With the following code, running `send_view` twice, and `all_views` once returns 4 views when it should be 2 unique, as seen in the new behaviour below.

```python
class MyView(View):
    @button(label="one")
    async def one(self, button, interaction: Interaction):
        await interaction.response.send_message("one")

    @button(label="two")
    async def two(self, button, interaction: Interaction):
        await interaction.response.send_message("two")

@bot.slash_command()
async def send_view(inter: Interaction):
    await inter.response.send_message("View", view=MyView())

@bot.slash_command()
async def all_views(inter: Interaction):
    views = bot.all_views
    views2 = bot.views(persistent=False)
    await inter.response.send_message(str(views) + "\n" + str(views2))
```

old and new behaviour

## This is a **Code Change**

### DIFF
--- a/nextcord/ui/view.py
+++ b/nextcord/ui/view.py
@@ -500,7 +500,13 @@ class ViewStore:
         self._state: ConnectionState = state
 
     def all_views(self) -> List[View]:
-        return [v for (v, _) in self._views.values()]
+        # Create a unique list of views, as _views stores the same view multiple times,
+        # one for each dispatchable item.
+        views = {
+            view.id: view
+            for view, _ in self._views.values()
+        }
+        return list(views.values())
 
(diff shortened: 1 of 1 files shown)


vcokltfre (approved):
Nice simple change, efficient and readable enough. LGTM.

spifory (approved):

PR #1234, opened 2024-12-14
build: set minimum python version to 3.12

### LINKED ISSUE #1205: Upgrade the minimum Python version
As we move to `3.0` we have the opportunity to introduce breaking changes, as a part of this I would like to propose a discussion surrounding moving the supported Python versions from it's current state to `Python 3.12` which is the version currently in bugfix ([source](https://devguide.python.org/versions/)).

ooliver1:
## Summary

Resolves #1205. This does not remove deprecated code or code that can be upgraded. This can be done in separate PRs, this is the minimum work to create this breaking change for 3.0 so 3.12 remains as the minimum version for the future.

<!--
## This is a **Code Change**

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8]
+        python-version: [3.12]
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -11,7 +11,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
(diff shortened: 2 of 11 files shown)


vcokltfre:
In `nextcord/utils.py` we have
```py

PY_310 = sys.version_info >= (3, 10)


if PY_310:
    from types import UnionType  # type: ignore

    # UnionType is the annotation origin when doing Python 3.10 unions. Example: "str | None"
else:
    UnionType = None
```
this can probably be replaced simply with
```py
from typing import UnionType
```
now that we're moving to 3.12 as a baseline. Later on we can think about using 'native' annotations instead of importing them from `typing` but I think that sort of large scale refactor is well out of scope for this PR, and likely isn't worth it anyway. Approving since this is minor and insignificant, and shouldn't be a barrier to a release, but a small change is ideal.

vcokltfre (approved):

PR #1235, opened 2024-12-27
refactor: update pyright to 1.1.391

ooliver1:
## Summary

Fix typing issues to update all the way to 1.1.391 from 1.1.317.

The most common change is that Pyright now checks types enclosed in a try except KeyError, so many of my changes refactor these cases to use if statements to avoid the exception. This IMO reads better especially for long try except else blocks where the else block is many lines from the try.

Following from this I will be updating Ruff, and then finally enabling typing rules for the newer formats of ``|``, ``dict[]``, etc.

## This is a **Code Change**

### DIFF
--- a/nextcord/abc.py
+++ b/nextcord/abc.py
@@ -149,9 +149,9 @@ class User(Snowflake, Protocol):
     __slots__ = ()
 
     name: str
-    global_name: str
+    global_name: Optional[str]
     discriminator: str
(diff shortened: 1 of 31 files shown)


Skelmis (approved):
Seems reasonable to me, merge it

teaishealthy (approved):
lgtm 🦅

PR #1255, opened 2025-05-21
feat!: make Member no longer require guild object.

alentoghostflame:
## Summary

Before this PR, `Member` objects required a corresponding `Guild` object to be instantiated. In an older era before interactions, this was fine. Nowadays, it's not only possible but expected (thanks UserApps) for a `Guild` object to not always be available.

This PR makes the `guild` parameter when creating `Member`s optional, and changes the various properties and methods to work with that.

## This is a **Code Change**


alentoghostflame:
Closing this PR in favor of returning a User object instead. Not ideal, but saving a breaking rework of Member for 4.0 would be nice.

PR #1260, opened 2025-06-25
feat: support pins pagination and limit removal

ooliver1:
## Summary

discord/discord-api-docs#7585

The limit has not actually been removed, so this can only be tested by temporarily changing `50` to `10` in `iterators.py` for example. This PR changes all 3 pin methods to use the new routes, making a new `fetch_pins` method over `pins` to avoid breaking changes with the use of an iterator.

## This is a **Code Change**


Soheab:
It’s worth noting that there’s no `after` parameter.

Here are the relevant documentation:

- https://github.com/discord/discord-api-docs/pull/7630
- https://discord.com/developers/docs/resources/message#get-channel-pins

Discord returns the messages in descending order of their pinned_at time, so you are supposed to do `before=last_pinned_at` if you need more messages.

ooliver1:
Interesting, thank you. Not sure why it seemed to paginate well with `after` but oh well.

nextcord/iterators.py
```diff
@@ -44,6 +45,7 @@
 
 
 OLDEST_OBJECT = Object(id=0)
+OLDEST_TIME = datetime.datetime(1970, 1, 1, tzinfo=datetime.timezone.utc)
```
Soheab, line 48:
```suggestion
```
Unused

nextcord/mixins.py
```diff
@@ -65,7 +73,55 @@ async def pins(self) -> List[Message]:
...
         return [state.create_message(channel=channel, data=m) for m in data]
+
+    async def fetch_pins(
+        self, limit: int = 50, before: Optional[SnowflakeTime] = None
+    ) -> AsyncIterator[MessagePin]:
+        """|asynciter|
+
+        Returns an async iterator of all messages that are currently pinned in the channel.
+
+        This requires :attr:`~nextcord.Permissions.read_message_history`, otherwise the result will be empty.
```
Soheab, line 94: According to the docs, view_channel is required too or else it'll return empty.

nextcord/mixins.py
```diff
@@ -65,7 +73,55 @@ async def pins(self) -> List[Message]:
...
+            :attr:`.MessagePin.message.reactions` data.
+
+        Parameters
+        ----------
+        limit: :class:`int`
+            The maximum number of messages to retrieve. Defaults to 50.
+        before: Optional[Union[:class:`.abc.Snowflake`, :class:`datetime.datetime`]]
+            The pin before which to retrieve pinned messages.
+        after: Optional[Union[:class:`.abc.Snowflake`, :class:`datetime.datetime`]]
+            The pin after which to retrieve pinned messages.
```
Soheab, line 111: There is no `after` parameter and "The pin before which to retrieve pinned messages"? That doesn't sound right.

nextcord/mixins.py
ooliver1: I avoided putting that in as to match `history`, but I suppose that could be added too.

PR #1263, opened 2025-07-06
Fix voice connection issues and upgrade to voice v8

Renaud11232:
## Summary

This PR is based on Rapptz/discord.py#10210, adapting it to nextcord

Fixes the 4006 errors caused by how nextcord handles the voice endpoints ports. Also upgrades the voice to v8.

This PR fixes #1262.

## This is a **Code Change**

- [x] I have tested my changes.
  => I've installed my forked version of nextcord and my bot was able to connect again and play audio
- [x] I have updated the documentation to reflect the changes.
  => Not applicable
- [x] I have run `task pyright` and fixed the relevant issues.
  => Remaining issues are not linked to my changes

PR #1265, opened 2025-07-11
feat(voice): Upgrade to voice v8

Renaud11232:
## Summary

This PR is based on Pycord-Development/pycord#2812.

I originally opened a PR (#1263) but thought it would be better to split it in two, since it was modifying two things at once.
This PR doesn't add any feature per se, but is rather about future proofing the voice API to avoid issues when Discord decides to drop support for voice v4.

## This is a **Code Change**

  => After the changes, my bot was still able to join and play audio.
  => No API changes

### DIFF
--- a/nextcord/gateway.py
+++ b/nextcord/gateway.py
@@ -190,6 +190,9 @@ class KeepAliveHandler(threading.Thread):
 
 
 class VoiceKeepAliveHandler(KeepAliveHandler):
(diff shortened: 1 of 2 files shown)


vcokltfre (approved):
Tested & working, thanks

ooliver1 (approved):
All looks good, thanks!

Issue #1258, opened 2025-06-21
forbidden path names for win32


srpcd:
in module "nextcord.__main__" function "to_path" there is a forbidden tuple inside of the if statement to check if the platform is win32. in forbidden there needs to be a CONIN$ and CONOUT$ because these are forbidden as well. thank you

onerandomusername:
> Then make a change for it instead of moaning to us here you could contribute properly

@NotKeira Opening an issue and reporting a bug IS contributing. If this is how you treat bug reports you might as well disable issues on this repo.

vcokltfre:
Thanks for this issue, we'll try to get the open PR for this merged to ensure this is fixed.

Normally I would stop there, but while I'm here I'll address the other comments on this issue.

Firstly, @NotKeira's comment is not an appropriate way to interact with issues; not everyone has the time. knowledge, expertise, or confidence to operate on the live codebase, make changes, submit them, deal with review comments, etc. and nor is there an expectation of that, which is why we have issues. An issue is, as @onerandomusername points out, a contribution, because it raises a - previously not noted - problem with the code to be fixed, which although not a code change is still valuable knowledge to have been given.

Secondly, regarding @onerandomusername's comment, while I agree with your perspective, I would prefer that the issue comments of issues on this repository don't become spaces for rhetorical responses like "you might as well disable issues on this repo" which is a childish response, especially when considering that the person it's aimed at is not a maintainer of the project. People making comments of this type do not represent the opinions of the maintainers, and suggesting they do by retorting in this way is unhelpful.

PR #1267, opened 2025-09-14
fix: add CONIN$ and CONOUT$ to forbidden paths

armup31:
## Summary

Added CONIN$ and CONOUT$ to forbidden path tuple

https://github.com/nextcord/nextcord/issues/1258#issuecomment-3289517726
added a pr for this issue specifically.

<!--
## This is a **Code Change**

- [x] I have tested my changes.
- [x] I have updated the documentation to reflect the changes.
- [x] I have run `task pyright` and fixed the relevant issues.

PR #1270, opened 2025-10-19
fix: forbid CONIN$ and CONOUT$ in to_path

vcokltfre:
## Summary

As suggested in https://github.com/nextcord/nextcord/issues/1258, this adds CONIN$ and CONOUT$ to the forbidden paths list in `to_path` in `nextcord/__main__.py`, as these aren't real paths on Windows and won't be interpreted by the OS as filesystem entries.

## This is a **Code Change**


vcokltfre:
I failed to notice the existing PR because it never linked the original issue, closing this and instead reviewing #1267

PR #1271, opened 2025-10-23
fix(voice): send speaking packet on reconnect

### LINKED ISSUE #882: bug: `VoiceClient.is_playing()` returns `True` even when the bot doesn't reproduce audio
### Summary

When forcily moving the bot into another `VoiceChannel` during an audio reproduction `VoiceClient.is_playing()` returns `True` even if the bot is not reproducing any sound. I haven't still checked if this issue affect other classes too but there could be other bugs. I've also noted a huge retard while the audio reproduction start, other fork of d.py (like disnake) doesn't seem to have this retard (its length is very variable)

### Reproduction Steps

- Connect to a `VoiceChannel`
- Play something with a `FFmpegPCMAudio` audio source
- Move the bot into another `VoiceChannel`

### Minimal Reproducible Code

```python
@bot.command()
async def lmao(ctx):
    channel: nextcord.VoiceChannel = await bot.fetch_channel(874005538285977603)
    vc = await channel.connect()
    vc.play(nextcord.FFmpegPCMAudio("https://github.com/ninjamuffin99/Funkin/blob/a083938ac803a91fbb15f03e432dea620f8a3b4a/assets/songs/bopeebo/Inst.mp3?raw=true", executable="C:\\Users\\davil\\OneDrive\\Desktop\\ffmpeg-master-latest-win64-gpl-shared\\bin\\ffmpeg.exe"))
    for i in range(40):
        print(f"[{i}] playing: {vc.is_playing()} paused: {vc.is_paused()}")
        await asyncio.sleep(1)
```

### Expected Results

`VoiceClient.is_playing()` should return `False`, i think that's a problem with the audio Thread for the value of `VoiceClient.is_playing()` (relying only on the Thread to know if the audio is playing doesn't take in account possible external interruptions, i mean from discord client side, moving the bot in another VoiceChannel doesn't stop the Thread from playing so that's the problem probaly) and maybe there's a problem with package sending, while the bot change VoiceChannel after some few second it should continue to play the Audio (right? or is this a discord api limitation?) 

### Actual Results

`VoiceClient.is_playing()` return `True` no matter what happens on client side

### Intents

Irrelevants

### System Information

nextcord v2.2.0

other information are irrelevants

### Checklist

### Additional Context

- **Issue** DisnakeDev/disnake#845
- **Issue** Rapptz/discord.py#8483

teaishealthy:
## Summary

`AudioPlayer` would previously not update the [`SPEAKING`](https://discord.com/developers/docs/topics/voice-connections#speaking) state to a non-zero value after reconnecting (which happens when moving between channels, for instance). This results in the bot continuing to send audio packets, which aren't processed by Discord as the speaking state would still be `0`.

Cherry-picked from https://github.com/DisnakeDev/disnake/commit/25ba55c53c41908327229b41008d007ef5659fa3

Closes #882

## This is a **Code Change**

### DIFF
--- a/nextcord/player.py
+++ b/nextcord/player.py
@@ -696,12 +696,10 @@ class AudioPlayer(threading.Thread):
             raise TypeError('Expected a callable for the "after" parameter.')
 
     def _do_run(self) -> None:
-        self.loops = 0
-        self._start = time.perf_counter()
+        self._reset_state(speak=True)
 
         # getattr lookup speed ups
         play_audio = self.client.send_audio_packet
-        self._speak(True)
 
         while not self._end.is_set():
             # are we paused?
@@ -715,8 +713,7 @@ class AudioPlayer(threading.Thread):
                 # wait until we are connected
                 self._connected.wait()
                 # reset our internal data
-                self.loops = 0
-                self._start = time.perf_counter()
+                self._reset_state(speak=self._resumed.is_set())
 
             self.loops += 1
             data = self.source.read()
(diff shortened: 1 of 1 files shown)


Skelmis (approved):
lgtm

ooliver1 (approved):

PR #1275, opened 2025-10-27
feat(voice): add SSRC-to-user mapping

tylerstraub:
Summary

Add SSRC → user_id tracking to VoiceClient and wire up handlers in DiscordVoiceWebSocket so bots can reliably attribute incoming voice packets to Discord users. Also fixes the SPEAKING opcode doc to note it’s bidirectional.  ￼

Changes
	•	VoiceClient
	•	New field: ssrc_map: Dict[int, int] to store SSRC→user_id.  ￼
	•	New method: get_user_id_from_ssrc(ssrc: int) -> Optional[int].  ￼
	•	DiscordVoiceWebSocket
	•	Route opcodes to new handlers in received_message: SPEAKING (5), CLIENT_CONNECT (12), CLIENT_DISCONNECT (13).  ￼
	•	handle_speaking: map data["ssrc"] → data["user_id"].  ￼
	•	handle_client_connect: map data["audio_ssrc"] → data["user_id"].  ￼
	•	handle_client_disconnect: remove mapping for the departing user_id.  ￼
	•	Docs: Change SPEAKING from “send only” to “bidirectional” in gateway docs.  ￼

Rationale

Voice receive paths often see SSRCs with no stable user association, leading to “unknown speaker” artifacts. Persisting SSRC→user_id whenever Discord signals speaking/connect/disconnect enables accurate speaker attribution for recording, transcription, and moderation.  ￼

### DIFF
--- a/nextcord/gateway.py
+++ b/nextcord/gateway.py
@@ -754,7 +754,7 @@ class DiscordVoiceWebSocket:
     SESSION_DESCRIPTION
         Receive only. Gives you the secret key required for voice.
     SPEAKING
-        Send only. Notifies the client if you are currently speaking.
+        Bidirectional. Notifies when you or others are speaking.
     HEARTBEAT_ACK
         Receive only. Tells you your heartbeat has been acknowledged.
     RESUME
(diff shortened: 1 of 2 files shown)


tylerstraub:
I've been developing an interactive voice assistant for Discord in my spare time using Nextcord for all of the Discord API interactions. 

It turns out I needed this SSRC to Discord ID mapping pretty badly in order to attribute speakers within a voice group chat correctly. Tried various workarounds until I discovered that it was something which _does_ exists within Discord opcode protocol, but just wasn't being exposed by Nextcord's abstractions. 

This solution worked well for me, and I tried to follow the contributing guidelines as closely as I could (branch name was a blunder). The PR is here now for consideration or at least to serve as an artifact for someone else encountering the same issue in the future. 

All the best-

T

teaishealthy:
of the key features you listed, you only actually implemented two, the others you mentioned:

- voice v8, the crashed voice server handling, "updated heartbeat mechanism" and "voice connection resume capability" was done in #1265 
- iscoroutinefunction was done in #1269

in any case i don't really see these features to be in scope for nextcord at the moment, this is part of voice receive which had open PRs in the past #1113 and #224

but i'd like to hear the thoughts of maintainers and other collaborators on this

tylerstraub:
my bad @teaishealthy... fixed the PR summary. the agent I used to push with must have picked up some other unrelated commit for the summary.

EmmmaTech:
I agree with tea here, I don't think voice recording is really in scope for nextcord, and also this pr has overlap with the currently open voice recording pr. I also checked the discord docs and didn't see a whole lot of info on ssrcs, but to me it seemed implied that they were between the client and the server? I could be wrong, some sample data or such would be nice.

PR #1290, opened 2026-06-21
fix(app-cmds): reject negative numbers when Range min is 0

### LINKED ISSUE #1015: nextcord.Range accepts negative numbers if min_value is 0.
### Summary

if the minimum value of nextcord.Range[x. y] is zero, negative numbers will be also accepted.

### Reproduction Steps

- have a function that uses a field that has a nextcord.Range[0, X] argument.
- type -1 into that field on discord

### Minimal Reproducible Code

```python
seed: nextcord.Range[0, 99999] = nextcord.SlashOption()
```

### Expected Results

Discord should not accept negative numbers if the minimum value is zero.

### Actual Results

i was able to input negative numbers into the field and they were accepted as valid.

### Intents

nextcord.Intents().all()

### System Information

- Python v3.11.2-final
- nextcord v2.4.1-final
- aiohttp v3.8.4
- system info: Windows 10 10.0.19045

### Checklist

### Additional Context

_No response_

ranshaa05:
It was a truthiness issue, where a range starting or ending in 0 would be interpreted as self.min or self.max = False, so negative numbers would be allowed

## Summary

Fix for #1015

<!--
## This is a **Code Change**

- [ *] I have tested my changes.

### DIFF
--- a/nextcord/application_command.py
+++ b/nextcord/application_command.py
@@ -3648,9 +3648,9 @@ class RangeMeta(type):
                 return value
 
             def modify(self, option: SlashCommandOption) -> None:
-                if self.min and option.min_value is None:
+                if self.min is not None and option.min_value is None:
                     option.min_value = self.min
-                if self.max and option.max_value is None:
+                if self.max is not None and option.max_value is None:
                     option.max_value = self.max
 
         if isinstance(value, tuple):


teaishealthy (approved):
tested

PR #1295, opened 2026-08-08
build(deps): update `dave.py` version requirement to `^1.0`

shiftinv:
## Summary

The current `dave.py` version specifier of `^0.1.2` is essentially `>=0.1.2,<0.2.0` (see poetry [spec](https://python-poetry.org/docs/dependency-specification/#caret-requirements)). This PR expands the dependency version to be more permissive and allow any `v1.x.x` version.
`dave.py` uses semver (not sure if I had already added the [versioning section](https://github.com/DisnakeDev/dave.py#versioning) of the readme when you implemented it? perhaps not.), i.e. no breaking changes outside of major versions, so this should not break in the future.

Most notably, the latest builds I pushed recently ([v0.2.0+](https://github.com/DisnakeDev/dave.py/releases/tag/v0.2.0)) include support for Python 3.15. c:

### DIFF
--- a/nextcord/voice_client.py
+++ b/nextcord/voice_client.py
@@ -830,6 +830,9 @@ class E2EEState:
             )
 
             key_package = self._session.get_marshalled_key_package()
+            if not key_package:
+                msg = "Failed to create new MLS key package"
+                raise RuntimeError(msg)
             await self.voice_client.ws.send_dave_mls_key_package(key_package)
 
         # When the epoch is greater than 1, the protocol version of the existing MLS group is changing.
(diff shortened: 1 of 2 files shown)


teaishealthy:
so it's semver without rule 4? 
> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

shiftinv:
I suppose, yeah.

Though now that you mention it, perhaps a v1.0.0 wouldn't be a terrible idea, I've been wanting to drop Python 3.9 for a while now. The 0.x versioning mostly just started out like that when things were still pretty early in development, and I never bothered to change it :>
I'll update things later today~

shiftinv:
Alrighty, updated everything to [v1.0.0](https://github.com/DisnakeDev/dave.py/releases/tag/v1.0.0), which is the latest release now.
Besides dropping support for Python 3.9, `Session.get_marshalled_key_package()` also now returns `None` instead of `b''` in case it fails (realistically this should never happen, though), which this PR now accounts for as well.

teaishealthy (approved):
rocket emoji

ooliver1 (approved):
Thank you!

PR #1300, opened 2026-09-03
fix(http): improve logging, eliminate duplication, and clean up type checking

NotKeira:
## Summary

Fix logging gaps, eliminate code duplication, and clean up type checking issues in the HTTP client.

## This is a **Code Change**

**Changes:**
- Fix missing log parameter in CloudFlare error logging
- Add type assertion for response in context manager  
- Extract payload filtering to helper method (removes 7 duplicated blocks)
- Add type suppression for payload assignment
- Update docstring formatting

**Why:**
- Better debugging info when CloudFlare issues occur
- Cleaner type checking, no IDE warnings
- Centralized payload filtering logic
- Follows codebase conventions

**Testing:**
- Pyright: 0 errors, 0 warnings
- HTTPClient imports successfully
- Logging fix verified
- No breaking changes

### DIFF
--- a/nextcord/http.py
+++ b/nextcord/http.py
@@ -841,6 +841,7 @@ class HTTPClient:
                         proxy_auth=self._proxy_auth,
                         **kwargs,
                     ) as response:
+                        assert response is not None
                         _log.debug(
                             "%s %s with %s has returned %s",
                             route.method,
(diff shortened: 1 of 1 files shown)


teaishealthy (changes requested):
if you type `payload` in `_filter_payload` as `Mapping[str, Any]`, you don't need to change the variance of the input type with the `cast`

PR #1301, opened 2026-09-06
fix(ui): validate custom IDs before storing them

### LINKED ISSUE #1187: Add a Errror for them custom-id
### Summary

Add an error if the custom ID of a UI object is an integer.

The core library

### The Problem

I did not know that the custom-id must be a string and I am almost desperate, the callback is not executed and you do not get an error message.

### The Ideal Solution

Add an error message if the custom-id is not of type string.

### The Current Solution

_No response_

### Additional Context

_No response_

be-student:
## Summary

Passing an integer `custom_id` to a UI constructor currently succeeds, producing an ID that cannot match the string returned in an interaction. This change raises `TypeError` at construction or assignment instead of allowing that invalid state to reach dispatch.

Fixes #1187.

- Share validation between Button, select menus, TextInput, and Modal constructors/setters. A rejected assignment preserves the existing ID.
- Preserve generated IDs through select subclasses that rebuild their underlying component, including omitted-ID sentinels from decorators.
- Preserve link-button behavior. Explicit TextInput and Modal IDs must be strings; Button retains its optional ID.
- Document the public UI contract. Low-level component payload APIs remain unchanged.

## Validation

On `b927878e`, nine temporary, uncommitted unittest groups passed with Python 3.12. They cover all eight public UI classes, invalid construction/assignment, defaults, decorators, reconstruction, and real storage/dispatch paths:

- ViewStore callback dispatch across all six item classes with generated, explicit, empty, Unicode, and 100-character string IDs; rejected assignment preserves the stored ID.
- ModalStore dispatch with generated and explicit IDs, including TextInput value refresh from submitted component data.
- `git diff --check` passed.

No Discord token or live connection was used. The added test harness, unittest lint exemptions, CI test step, and contributor test instructions were removed in response to the scope feedback. The follow-up changes no runtime code; the existing CI checks remain intact. All eight checks passed on the preceding head `0eb3d34`; checks for the updated head are pending.

AI assistance: OpenAI Codex implemented and reviewed the change and ran the automated checks described above.

### DIFF
--- a/nextcord/ui/modal.py
+++ b/nextcord/ui/modal.py
@@ -13,7 +13,7 @@ from typing import TYPE_CHECKING, Any, Callable, Dict, Iterator, List, Optional,
 
 from ..components import Component
 from ..utils import MISSING
-from .item import Item
+from .item import Item, _validate_custom_id
 from .view import _component_to_item, _ViewWeights, _walk_all_components
 
 __all__ = (
@@ -105,6 +105,15 @@ class Modal:
         self.__timeout_task: Optional[asyncio.Task[None]] = None
         self.__background_tasks: Set[asyncio.Task[None]] = set()
         self.__stopped: asyncio.Future[bool] = loop.create_future()
+
+    @property
+    def custom_id(self) -> str:
+        return self._custom_id
+
+    @custom_id.setter
+    def custom_id(self, value: str) -> None:
+        _validate_custom_id(value)
+        self._custom_id = value
 
     async def __timeout_task_impl(self) -> None:
         while True:
--- a/nextcord/ui/modal.py
+++ b/nextcord/ui/modal.py
@@ -93,6 +93,8 @@ class Modal:
         self.title = title
         self.timeout = timeout
         self._provided_custom_id = custom_id is not MISSING
+        # Assignment goes through the property setter, so generated and supplied
+        # IDs use the same validation as later reassignment.
         self.custom_id = os.urandom(16).hex() if custom_id is MISSING else custom_id
         self.auto_defer = auto_defer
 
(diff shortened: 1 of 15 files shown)


nextcord/ui/modal.py
```diff
@@ -106,6 +106,15 @@ def __init__(
...
 
+    @property
+    def custom_id(self) -> str:
+        return self._custom_id
+
+    @custom_id.setter
+    def custom_id(self, value: str) -> None:
+        _validate_custom_id(value)
+        self._custom_id = value
+
```
Skelmis, line 120: This conflicts with line 96 which also defines custom id and uses different logic. I imagine it still works as a result of tests, but figured I'd ask about it

be-student:
The two paths serve different layers. `InteractiveComponent.__init__` is the wire-level component normalization: `None` generates an ID, while `MISSING` remains valid for link and premium buttons. `_validate_custom_id` is used by the public `nextcord.ui` constructors and setters so invalid runtime values fail before reaching the underlying component, while preserving each UI type's existing `None`/generated-ID behavior. The regression tests cover generated IDs, link buttons, required-ID components, assignment, and reconstruction. I don't see a behavioral conflict between them, but I can consolidate validation into the lower layer if that is the preferred API boundary.

teaishealthy:
i don't think this is the time (nor place) for testing

be-student:
Removed the added test harness, unit-test workflow step, and contributor testing section in `6c710863` to keep this PR focused on the UI validation change. The API documentation and runtime fix remain. I reran the checks outside the repository: nine groups passed, including real ViewStore callback dispatch and ModalStore/TextInput value refresh. The PR description now reflects that scope and validation.

